PR #8, opened 2019-06-10
AStatefulSet supports in-place update

FillZpp:
Make Advanced StatefulSet support in-place update.

Currently we can set `spec.updateStrategy.rollingUpdate.podUpdatePolicy` to `InPlaceIfPossible` or `InPlaceOnly` to enable in-place update.

Read the docs for further details.


pkg/apis/apps/v1alpha1/statefulset_types.go
```diff
@@ -23,6 +24,32 @@ import (
...
+// InPlaceUpdateState records latest inplace-update state, including old statuses of containers.
+type InPlaceUpdateState struct {
+	Revision              string                                  `json:"revision"`
+	UpdateTimestamp       metav1.Time                             `json:"updateTimestamp"`
+	LastContainerStatuses map[string]InPlaceUpdateContainerStatus `json:"lastContainerStatuses"`
+}
+
+// InPlaceUpdateContainerStatus records container status in current pod.
+type InPlaceUpdateContainerStatus struct {
+	//ContainerID string `json:"containerID,omitempty"`
```
jian-he: remove these two unneeded fields

pkg/apis/apps/v1alpha1/statefulset_types.go
```diff
@@ -23,6 +24,32 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const (
+	// StatefulSetInPlaceUpdateReady must add into template.spec.readinessGates when pod podUpdatePolicy
+	// is InPlaceIfPossible or InPlaceOnly. The condition in podStatus will be updated to False before in-place
+	// updating and updated to True after finished updating. This ensures pod being not-ready during
+	// in-place updating.
+	StatefulSetInPlaceUpdateReady v1.PodConditionType = "InPlaceUpdateReady"
```
jian-he: Instead of forcing user to set the readiness gate in the statefulset's template,  can this be done by the kruise-controller to update the generated pod spec to include the readiness gate

pkg/webhook/default_server/statefulset/validating/statefulset_create_update_handler.go
```diff
@@ -76,6 +77,13 @@ func (h *StatefulSetCreateUpdateHandler) Handle(ctx context.Context, req types.R
 		if allErrs := append(validationErrorList, updateErrorList...); len(allErrs) > 0 {
 			return admission.ErrorResponse(http.StatusUnprocessableEntity, allErrs.ToAggregate())
 		}
+
+		if obj.Spec.UpdateStrategy.RollingUpdate != nil &&
+			obj.Spec.UpdateStrategy.RollingUpdate.PodUpdatePolicy == appsv1alpha1.InPlaceOnlyPodUpdateStrategyType {
+			if err := validateTemplateInPlaceOnly(&oldObj.Spec.Template, &obj.Spec.Template); err != nil {
+				return admission.ErrorResponse(http.StatusUnprocessableEntity, fmt.Errorf("invalid template modified with InPlaceOnly: %v", err))
+			}
```
jian-he: Make it more clear:
Invalid pod template modified with InPlaceOnly update strategy. Currently, only image update is allowed for InPlaceOnly.

pkg/apis/apps/v1alpha1/statefulset_types.go
FillZpp: I don't think it is a good idea to update readinessGates automatically in controller or webhook.. Modification of statefulset template should be very careful because it could raise pod updating.

For example, someone create a statefulset with 100 replicas, then he updates podUpdatePolicy to `InPlaceIfPossible`. If kruise-manager find this and add `InPlaceUpdateReady` into readinessGates, then pods start to be deleted because of recreate update. This could make user confused and is unsafe for online applications.

jian-he: I meant the controller do this when creating the pods.
It's a good question what if user wants to update the policy to `InPlaceIfPossible` regardless of how readiness is specified originally, in this case the `readinessGate` cannot be updated to the existing pods.

jian-he:
LGTM

PR #9, opened 2019-06-13
update some configurations and client generated

FillZpp:
As the title implied.


Makefile
```diff
@@ -1,6 +1,6 @@
 
 # Image URL to use all building/pushing image targets
-IMG ?= openkruise/kruise-manager:v1alpha1
+IMG ?= openkruise/kruise-manager:test
```
jian-he: should this also be v0.1.0

FillZpp: Default `IMG` should not be release tag, it is just for testing. Otherwise, if someone executes `make build and push`, the release image might be overwrite which is a terrible mistake. 

When we need to build a release image, we must set the environment `IMG=openkruise/kruise-manager:v0.1.0` first.

jian-he:
LGTM

PR #10, opened 2019-06-14
update yamls and add helm chart

FillZpp:
Update yamls:
- use Deployment to deploy kruise-manager instead of StatefulSet
- the default Deployment will create 3 Pods, but only one of them can be Ready through leaderelection, other of them will keep acquiring lease
- once the leader Pod going down, another Pod will get the lease and turn to leader in a short time

Helm chart:
build a helm chart for Kruise, then we need to package and upload it so that users outside can install the chart in their clusters.


jian-he:
LGTM

PR #45, opened 2019-06-24
Fix a bug that StatefulSet applies the revision incorrectly

tcmichael:
I think the root cause of https://github.com/kubernetes/kubernetes/issues/78420 is that StatefulSet, which will be json.Unmarshal to, is not reset.
so that ApplyRevision makes the subfield env of StatefulSet has both field valueFrom and value that is invalid.

fixes https://github.com/kubernetes/kubernetes/issues/78420

unit test

in https://github.com/kubernetes/kubernetes/issues/78420

### DIFF
--- a/pkg/controller/statefulset/stateful_set_utils.go
+++ b/pkg/controller/statefulset/stateful_set_utils.go
@@ -340,11 +340,12 @@ func ApplyRevision(set *appsv1alpha1.StatefulSet, revision *apps.ControllerRevis
 	if err != nil {
 		return nil, err
 	}
-	err = json.Unmarshal(patched, clone)
+	restoredSet := &appsv1alpha1.StatefulSet{}
+	err = json.Unmarshal(patched, restoredSet)
 	if err != nil {
 		return nil, err
 	}
-	return clone, nil
+	return restoredSet, nil
 }
 
 // nextRevision finds the next valid revision number based on revisions. If the length of revisions
--- a/pkg/controller/statefulset/stateful_set_utils_test.go
+++ b/pkg/controller/statefulset/stateful_set_utils_test.go
@@ -20,6 +20,7 @@ package statefulset
 import (
 	"fmt"
 	"math/rand"
+	"reflect"
 	"sort"
 	"strconv"
 	"testing"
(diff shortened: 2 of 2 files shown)


tcmichael:
@resouer

pkg/controller/statefulset/stateful_set_utils.go
```diff
@@ -335,16 +335,17 @@ func newRevision(set *appsv1alpha1.StatefulSet, revision int64, collisionCount *
...
-	clone := set.DeepCopy()
-	patched, err := strategicpatch.StrategicMergePatch([]byte(runtime.EncodeOrDie(patchCodec, clone)), revision.Data.Raw, clone)
+	patched, err := strategicpatch.StrategicMergePatch([]byte(runtime.EncodeOrDie(patchCodec, set)), revision.Data.Raw, set)
 	if err != nil {
 		return nil, err
 	}
-	err = json.Unmarshal(patched, clone)
+
+	restoredSet := &appsv1alpha1.StatefulSet{}
+	err = json.Unmarshal(patched, restoredSet)
```
FillZpp: How about 
```
	var restoredSet *appsv1alpha1.StatefulSet
	err = json.Unmarshal(patched, &restoredSet)
```

FillZpp:
This can be merged after [#78853](https://github.com/kubernetes/kubernetes/pull/78853) merged.

pkg/controller/statefulset/stateful_set_utils.go
tcmichael: it will panic. `var restoredSet *appsv1alpha1.StatefulSet` equals `var restoredSet *appsv1alpha1.StatefulSet = nil`, and can not `json.Unmarshal(nil)`

FillZpp: It will not. Just try it.

xiang90: @FillZpp 

it should be 
```go
    var restoredSet appsv1alpha1.StatefulSet
    err = json.Unmarshal(patched, &restoredSet)
```

or it will panic as @tcmichael mentioned.

tcmichael:
#78853 was merged.
@FillZpp @xiang90 PTAL.

PR #52, opened 2019-06-27
Remove the use of "slave"

jamiebuilds:
Redis has begun renaming "slave" to "replica" which is both a more accurate term to describe what it does and doesn't use the language of oppressors.

There are more instances of the word "slave" used in this repo, if this PR is welcomed I can go through the rest


xiang90:
lgtm

xiang90:
@jamiebuilds thanks!

PR #94, opened 2019-07-23
Fix Kustomize template

Fei-Guo:
Kustomize will add a Kruise- prefix for all the object names in the
template. config/manager/webhookconfiguration.yaml needs to be fixed
to avoid generating objects with wrong names.

Bonus fix: add a prerequsite for `make run` command in README.md.

NONE.

NA

Manually verify that a broadcast job can be created after running `make deploy`. Without this fix,
the job creation will fail with certification error.


jian-he:
LGTM

PR #103, opened 2019-07-25
support to run kruise-manager locally

### LINKED ISSUE #96: [BUG]local debugging is not successful
**What happened**:
 Local debugging with `make run` command has an error.
 When  installing statefulset，it will request kruise-webhook-server-service，but the service not binding an endpoints,So it will connect refuesed.

**What you expected to happen**:Local debugging successfully

1. make run 
2. install all crds
3. install kruise-system namespace
4. install a kruise statefulset demo

For debugging successfully，i try do the following:
1.remove the webhook service selector.
2. create an endpoint with the same name as the webhook service,and the endpoint subset use local ip.The endpoint will  forward traffic to the local.

But there is still a problem，requesting webhook service  has a certificate x509 error.

I am not very familiar with the signature,can you help me ?

**Environment**:
- Kruise version: latest version from github
- Kubernetes version (use `kubectl version`): v1.15.0
- OS (e.g: `cat /etc/os-release`):macOS 10.14.5
- Kernel (e.g. `uname -a`): Darwin Kernel Version 18.6.0
- Install tools: minikube&&kubectl
- Others:

FillZpp:

Support to run kruise-manager process locally.

Fixes #96

### DIFF
--- a/docs/debug/README.md
+++ b/docs/debug/README.md
@@ -1,3 +1,35 @@
+# How to DEBUG
+
+## DEBUG: start with process locally
+
+Start kruise-manager process locally, it can connect to a local or remote k8s-cluster.
+
+**Set up hostname**
+
+First, you should think of a fake hostname for your local machine, such as `power-machine`.
+
+Add this line into `/etc/hosts` file in which machine that runs kube-apiserver or minikube:
+
+```
+IP_OF_YOUR_LOCAL_MACHINE power-machine
+```
+
+*Note that `IP_OF_YOUR_LOCAL_MACHINE` should be ETH_IP instead of 127.0.0.1*
+
+**Install CRDs into cluster and run kruise-manager**
+
+Run these locally:
+
+```bash
+export KUBECONFIG=${PATH_TO_CONFIG}
+export WEBHOOK_HOST=power-machine
+
+make install
+make run
+```
+
+## DEBUG: start with Pod
+
 The followings are the steps to debug Kruise controller manager locally using Pod.
 
 **Install docker**
--- a/pkg/webhook/default_server/server.go
+++ b/pkg/webhook/default_server/server.go
@@ -47,27 +47,32 @@ func Add(mgr manager.Manager) error {
 		secretName = "kruise-webhook-server-secret"
 	}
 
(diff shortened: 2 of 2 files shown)


docs/debug/README.md
```diff
@@ -1,3 +1,32 @@
...
+
+## DEBUG: start with process locally
+
+Start kruise-manager process locally, it can connect to a local or remote k8s-cluster.
+
+**Set up hostname**
+
+First, you should think of a fake hostname for your local machine, such as `power-machine`.
+
+Add this line into `/etc/hosts` file in which machine that runs kube-apiserver:
```
Fei-Guo:
```suggestion
Add this line into `/etc/hosts` file in dev machine that runs kube-apiserver or minikube:
```
I have tested that this changes works for dev box that runs minikube without running apiserver directly.

docs/debug/README.md
```diff
@@ -1,3 +1,32 @@
...
+```
+IP_OF_YOUR_LOCAL_MACHINE power-machine
+```
+
```
Fei-Guo: Please add the following:

Note: "IP_OF_YOUR_LOCAL_MACHINE" cannot be loop device ip 127.0.0.1

docs/debug/README.md
FillZpp: Even if you configure your dev machine using minikube, then this it the machine that runs kube-apiserver.

My point is that we have not to care where the kube-apiserver runs. It may runs in local or dev machine started by minikube, or in pods of a k8s cluster, it doesn't matter.

docs/debug/README.md
FillZpp: Do you mean if we execute `make run` in machine A and the minikube is deployed in machine B, then we should put `IP_OF_A  power-machine` into `/etc/hosts` of B, instead of `127.0.0.1 power-machine` ?

Fei-Guo: I mean put
127.0.0.1 power-machine
in /etc/hosts does not work。we should put 
ETH_IP power-machine

Someone may think 127.0.0.1 is also the IP_OF_YOUR_LOCAL_MACHINE.

FillZpp: done

PR #120, opened 2019-09-01
format codes to pass the code-check

wu8685:


wu8685:
It still gets hits on file `pkg/controller/statefulset/stateful_inplace_utils_test.go`, but I m not able to find them using gometalinter locally.

wu8685:
It still gets hits on file `pkg/controller/statefulset/stateful_inplace_utils_test.go`, but I m not able to find them using gometalinter locally.

Fei-Guo:
/LGTM 

thanks for fixing it.

PR #122, opened 2019-09-06
Change the type of `Parallelism` field in BroadcastJob

xlgao-zju:
Signed-off-by: Xianglin Gao <xianglin.gxl@alibaba-inc.com>

### DIFF
--- a/docs/concepts/broadcastJob/README.md
+++ b/docs/concepts/broadcastJob/README.md
@@ -28,10 +28,11 @@ BroadcastJob.
 ### Parallelism
 
 `Parallelism` specifies the maximal desired number of Pods that should be run at
-any given time. By default, there's no limit.
+any given time. By default, there's no limit.`Parallelism` can be an int or a percent.
 
 For example, if a cluster has ten nodes and `Parallelism` is set to three, there can only be
-three pods running in parallel. A new Pod is created only after one running Pod finishes.
+three pods running in parallel, or if a cluster has ten nodes and `Parallelism` is set to 20%,
+there can only be two pods running in parallel. A new Pod is created only after one running Pod finishes.
 
 ### CompletionPolicy
 
--- a/pkg/apis/apps/v1alpha1/broadcastjob_types.go
+++ b/pkg/apis/apps/v1alpha1/broadcastjob_types.go
@@ -19,6 +19,7 @@ package v1alpha1
 import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 // BroadcastJobSpec defines the desired state of BroadcastJob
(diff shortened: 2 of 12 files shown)


pkg/apis/apps/v1alpha1/broadcastjob_types.go
```diff
@@ -28,7 +28,7 @@ type BroadcastJobSpec struct {
 	// be less than this number when the work left to do is less than max parallelism.
 	// Not setting this value means no limit.
 	// +optional
-	Parallelism *int32 `json:"parallelism,omitempty" protobuf:"varint,1,opt,name=parallelism"`
+	Parallelism *string `json:"parallelism,omitempty" protobuf:"varint,1,opt,name=parallelism"`
```
xiang90: is this a backward compatible change? if not, add this to change log.

pkg/controller/broadcastjob/broadcastjob_controller.go
```diff
@@ -300,15 +304,24 @@ func addLabelToPodTemplate(job *appsv1alpha1.BroadcastJob) {
...
 	// max concurrent running pods
 	var parallelism int32
 	if job.Spec.Parallelism == nil {
 		// not specify,take the max int
 		parallelism = int32(1<<31 - 1)
 	} else {
-		parallelism = *job.Spec.Parallelism
+		parallelismIntStr := intstr.Parse(*job.Spec.Parallelism)
+		if parallelismIntStr.Type == intstr.String {
+			parallelismInt, err := strconv.Atoi(strings.TrimSuffix(parallelismIntStr.StrVal, "%"))
```
xiang90: move this to a util func? func percentageToAbsolute for example.

pkg/controller/broadcastjob/broadcastjob_controller_test.go
```diff
@@ -54,7 +54,7 @@ func TestReconcileJobCreatePod(t *testing.T) {
 	_ = appsv1alpha1.AddToScheme(scheme)
 	_ = v1.AddToScheme(scheme)
 
-	p := int32(2)
+	p := "2"
```
xiang90: we should alao add a test for percentage value.

xiang90:
let us separate this into 3 different PRs for the 3 new features.

docs/concepts/broadcastJob/README.md
```diff
@@ -28,10 +28,10 @@ BroadcastJob.
 ### Parallelism
 
 `Parallelism` specifies the maximal desired number of Pods that should be run at
-any given time. By default, there's no limit.
+any given time. By default, there's no limit.`Parallelism` can be an int or a percent.
 
 For example, if a cluster has ten nodes and `Parallelism` is set to three, there can only be
-three pods running in parallel. A new Pod is created only after one running Pod finishes.
+three pods running in parallel. a cluster has ten nodes and `Parallelism` is set to 20%, there can only be two pods running in parallel. A new Pod is created only after one running Pod finishes.
```
Fei-Guo: How about this 

For example, if a cluster has ten nodes and `Parallelism` is set to three, there can only be
three pods running in parallel, or if a cluster has ten nodes and `Parallelism` is set to 20%, there can only be two pods running in parallel.

pkg/apis/apps/v1alpha1/broadcastjob_types.go
Fei-Guo: Why not Parallelism intstr.IntOrString?  I think it is ok to break backward compatibility for now as long as we document it well in change log.

xlgao-zju:
> let us separate this into 3 different PRs for the 3 new features.

`Add FailurePolicy in BroadcastJob` depends on ` Support Pause in BroadcastJob`, since we have a FailurePolicy called `Pause`. So, I will file two PRs.

xlgao-zju:
@Fei-Guo updated, PTAL

Issue #124, opened 2019-09-08
Install kubebuilder before running test by makefile


wu8685:
In order to use controller-runtime test framework, we should install kubebuilder before running test.

FillZpp:
@wu8685 There is no need to install kubebuilder for running test. Just `make test` is fine.

wu8685:
Without kubebuilder installed, it is not able to run kubebuilder generated test case. 
Additionally, it is a convenient way to build up a test environment, by which we can test our codes better.

Fei-Guo:
Fixed by https://github.com/openkruise/kruise/commit/fe73a2e404ccae28998eceec2874be10eb421568

PR #128, opened 2019-09-13
Install kubebuilder if not available

Fei-Guo:
kubebuilder binary is needed to run default test framework. Change the makefile to automatically install kubebuilder binary if it is not available in the machine

#124 

N/A

This is part of the output of `make` cmd if kubebuilder binary does not exists

```
...
# Download kubebuilder and extract it to tmp
curl -sL https://go.kubebuilder.io/dl/1.0.8/darwin/amd64 | tar -xz -C /tmp/
# You'll need to set the KUBEBUILDER_ASSETS env var if you put it somewhere else
sudo mv /tmp/kubebuilder_1.0.8_darwin_amd64 /usr/local/kubebuilder
Password:
...
```

This change fixes the kubebuilder binary version to 1.0.8 which is the one Kruise used to create controller scaffolding code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ go_check:
 	@scripts/check_go_version "1.11.1"
 
 # Run tests
-test: generate fmt vet manifests
+test: generate fmt vet manifests kubebuilder
 	go test ./pkg/... ./cmd/... -coverprofile cover.out
 
 # Build manager binary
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -50,6 +50,17 @@ endif
 	go generate ./pkg/... ./cmd/...
 	go run ./hack/gen-openapi-spec/main.go > ${CURRENT_DIR}/api/openapi-spec/swagger.json
 
+# kubebuilder binary
+kubebuilder:
+ifeq (, $(shell which kubebuilder))
+	# Download kubebuilder and extract it to tmp
+	curl -sL https://go.kubebuilder.io/dl/2.0.0/$(shell go env GOOS)/$(shell go env GOARCH) | tar -xz -C /tmp/
```
FillZpp: Kruise is generated by kubebuilder 1.x, and has not supported kubebuilder 2.x yet. Maybe we should install 1.0.8 instead of 2.0.0 .

Fei-Guo: Ah, good catch. I meant to install 1.0.8 but makes a mistake in copy/paste. Will change soon.

Fei-Guo: done

Issue #118, opened 2019-08-29
[docs request] kruise install with helm


hellolijj:
**What would you like to be added**:

docs request: kruise install with helm 

**Why is this needed**:

It is convient to install `kruise`  with helm [charts](https://developer.aliyun.com/hub/detail?spm=a2c6h.12873679.0.0.74e51107atHbj2&name=kruise&version=0.1.0#/?_k=up9kfg).

Fei-Guo:
/assign @hellolijj 

As we discussed offline, we need to resolve the CRD yaml version problem in Helm chart before adding the the link in README.

hellolijj:
> /assign @hellolijj
> 
> As we discussed offline, we need to resolve the CRD yaml version problem in Helm chart before adding the the link in README.

@Fei-Guo 

How about we make a folder named helm? it contains installed yaml file. then it use as follows?

```
$ git clone https://github.com/openkruise/kruise.git
$ cd kruise/helm
$ helm install ./
```

hellolijj:
or we can update helm repo as kruise release?

Fei-Guo:
One problem is that you allow user to configure image version in helm chart. Then using per release directory cannot help solving the version problem, unless you disallow this option
"manager.image.tag" 

I am ok with adding a helm directory in kruise repo and keep it updated for every release.

PR #129, opened 2019-09-18
 install kruise  with helm 

hellolijj:
install kruise with helm

kruise can be installed with helm.

#118 

no


Fei-Guo:
I would assume you will create a directory for each released version and put entire kruise directory in each version directory. Otherwise,  we still have a problem such that if user puts a different version tag for kruise-manager, the crd yamls mismatch with kruise-manager image version. Do I miss anything?
 
 # tag for kruise-manager image
    tag: v0.1.0-beta.3

hellolijj:
> different


Maybe we should disable `kruise-manager` image configration.

Fei-Guo:
> > different
> 
> Maybe we should disable `kruise-manager` image configration.

Are you sure the yaml files in this change is the same as those in https://github.com/openkruise/kruise/tree/v0.1.0/config/crds ?

I still think we need to create subdirectroies under ./charts/kruise/ 
say ./charts/kruise/v0.1.0 to copy correct yaml files for each release.

PR #136, opened 2019-09-25
add script to auto generate chart folder

hellolijj:
Signed-off-by: Junjun Li <junjunli666@gmail.com>

add script to auto generate chart folder

#131

### DIFF
--- a/hack/auto_generate_charts.sh
+++ b/hack/auto_generate_charts.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+set -o errexit
+set -o nounset
+
+# receive version version
+echo "---------------------------------"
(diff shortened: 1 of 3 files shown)


hack/auto_generate_charts.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+set -o errexit
+set -o nounset
+
+# receive version version
+echo "----------------------------------"
+echo "Now auto version"
```
Fei-Guo: How about this 
echo "-------------------------------"
echo "Please input Kruise version (available versions are listed in https://github.com/openkruise/kruise/releases)"
echo "If no input, the default version of v0.1.0 will be used"
echo "-------------------------------"

Issue #134, opened 2019-09-23
how to config external api-server address？


jinyuchen724:
hi:
    I changed install config like this:

spec:
      containers:
        - command:
            - /manager
          args:
            - "--metrics-addr=127.0.0.1:8080"
            - "--logtostderr=true"
            - "--v=4"
            - "--master=http://10.39.128.35:80"
          env:
            - name: POD_NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
            - name: SECRET_NAME
              value: kruise-webhook-server-secret
            - name: KUBERNETES_SERVICE_HOST
              value: 10.39.128.35
            - name: KUBERNETES_SERVICE_PORT
              value: "80"

**But i got this error:**

`I0923 12:19:52.712803       1 main.go:54] entrypoint "level"=0 "msg"="setting up client for manager"  
I0923 12:19:52.713181       1 main.go:62] entrypoint "level"=0 "msg"="setting up manager"  
E0923 12:19:52.715201       1 :0] kubebuilder/manager "msg"="Failed to get API Group-Resources" "error"="Get https://10.39.128.35:80/api?timeout=32s: http: server gave HTTP response to HTTPS client"  
E0923 12:19:52.715226       1 main.go:74] entrypoint "msg"="unable to set up overall controller manager" "error"="Get https://10.39.128.35:80/api?timeout=32s: http: server gave HTTP response to HTTPS client"  `

why I use 80 port，but manager still use "https" to access api-server?How can i do it？
thank you!

Fei-Guo:
You have two approaches to connect APIServer from controller manager:
- InCluster Mode (default):
  In this case, the cluster config only supports "https:"
https://github.com/kubernetes/client-go/blob/2bd34d193bacaec5c976bef54bbc1f80939b81ec/rest/config.go#L435
  It is hardcoded to avoid security vulnerability. 
- kubeconfig mode (can be out of cluster )
```
// If --kubeconfig is set, will use the kubeconfig file at that location.  Otherwise will assume running
// in cluster and use the cluster provided kubeconfig.
```
if you start the controller manager by specifying a kubeconfig file, it will use the server url provided by the kubeconfig in cluster map. 

In your case, by adding `--master` option will not work. You can try the kubeconfig mode by
providing a proper kubeconfig file to start the controller manager.

jinyuchen724:
> You have two approaches to connect APIServer from controller manager:
> 
> * InCluster Mode (default):
[...]

PR #140, opened 2019-09-26
Gracefully handle uninitialized objects and notify webhook does not work

Fei-Guo:

Kruise relies on admission webhooks to set default fields for custom resources. If APIServer does not enable MutatingAdmissionWebhook,ValidatingAdmissionWebhook admission plugins, custom resources can be created without any desired initializations. This may lead controller manager to constantly crash in case a pointer is not initialized with a default object. See #134 for example.

Unfortunately, APIServer does not provide any API to query if certain admission plugin is enabled or not. In this change, we workaround this problem by doing the following:

1) Every time controller manager starts, it will do a self webhook enablement check by creating a test pod and checking if an expected signature annotation is added by pod mutant admission webhook. If this check fails, controller manager quits and suggests admin to check if APIServer has enabled required admission plugins.
2) It is still possible that people create kruise custom resources when controller is not working. To avoid controller crashing due to existing objects, controllers now mandatorily set defaults for reconciled objects read from APIServer. Unlike setting defaults in mutant admission webhook where defaults will be updates in APIServer, setting defaults in controller does not guarantee the defaults will be persisted in APIServer. 

We may need other uninitialized object identification mechanisms but i'd defer to do them unless they are necessary.

#134 

Mostly manual tests and verify tutorial works.

In case API Server enable admission plugin, controller log shows the following
I0926 22:15:21.735615       1 main.go:115] entrypoint "level"=0 "msg"="Start self webhook check"
I0926 22:15:21.745001       1 pod_create_handler.go:84] [WebhookTest] Inject annotation for webhook test pod default/webhook-tester-hje3h0rvreaj6w7z
I0926 22:15:22.832078       1 main.go:156] entrypoint "level"=0 "msg"="Webhook is working!"

In case API Server does not enable admission plugin, controller log shows the following

I0926 22:57:23.726331       1 main.go:115] entrypoint "level"=0 "msg"="Start self webhook check"
E0926 22:57:25.303721       1 main.go:150] entrypoint "msg"="fail to validate webhook test pod" "error"="Pod mutating webhook does not work. Please make sure APIServer enables MutatingAdmissionWebhook and ValidatingAdmissionWebhook admission plugins"
I0926 22:57:25.401487       1 main.go:158] entrypoint "level"=0 "msg"="Kruise controller relies on webhooks in order to work properly. Stop the controller since webhook test fails!"

This change also fixes an unstable test.

### DIFF
--- a/pkg/webhook/default_server/broadcastjob/mutating/broadcastjob_create_update_handler_test.go
+++ b/pkg/webhook/default_server/broadcastjob/mutating/broadcastjob_create_update_handler_test.go
@@ -76,6 +76,10 @@ func TestHandle(t *testing.T) {
 	}
 
 	if !reflect.DeepEqual(expectedPatches, resp.Patches) {
-		t.Fatalf("expected patches %+v, got patches %+v", expectedPatches, resp.Patches)
+		// The response order is not deterministic
+		expectedPatches[0], expectedPatches[1] = expectedPatches[1], expectedPatches[0]
+		if !reflect.DeepEqual(expectedPatches, resp.Patches) {
+			t.Fatalf("expected patches %+v, got patches %+v", expectedPatches, resp.Patches)
+		}
 	}
 }
--- a/pkg/webhook/default_server/broadcastjob/mutating/broadcastjob_create_update_handler_test.go
+++ b/pkg/webhook/default_server/broadcastjob/mutating/broadcastjob_create_update_handler_test.go
@@ -3,6 +3,7 @@ package mutating
 import (
 	"context"
 	"reflect"
+	"sort"
 	"testing"
 
 	"github.com/appscode/jsonpatch"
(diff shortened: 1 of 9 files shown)


pkg/webhook/default_server/broadcastjob/mutating/broadcastjob_create_update_handler_test.go
```diff
@@ -76,6 +76,10 @@ func TestHandle(t *testing.T) {
 	}
```
FillZpp: We can sort the response patches.
```
        sort.SliceStable(resp.Patches, func(i, j int) bool {
		return resp.Patches[i].Path < resp.Patches[j].Path
	})
```

Fei-Guo: done

PR #142, opened 2019-09-27
Gracefully handle uninitialized objects

Fei-Guo:

Kruise relies on admission webhooks to set default fields for custom resources. If APIServer does not enable MutatingAdmissionWebhook,ValidatingAdmissionWebhook admission plugins, custom resources can be created without any desired initializations. This may lead controller manager to constantly crash in case a pointer is not initialized with a default object. See #134 for example.

Unfortunately, APIServer does not provide any API to query if certain admission plugin is enabled or not. To avoid controller crashing due to existing uninitialized objects, now we add type defaulting Funcs in API schema for all Kruise CRDs. 

#134 

This change also fixes an unstable test.

### DIFF
--- a/pkg/webhook/default_server/sidecarset/mutating/sidecarset_create_update_handler.go
+++ b/pkg/webhook/default_server/sidecarset/mutating/sidecarset_create_update_handler.go
@@ -22,9 +22,7 @@ import (
 	"net/http"
 
 	"k8s.io/api/admission/v1beta1"
-	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/klog"
-	"k8s.io/kubernetes/pkg/apis/core/v1"
 
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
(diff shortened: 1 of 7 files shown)


pkg/webhook/default_server/sidecarset/mutating/sidecarset_create_update_handler.go
```diff
@@ -160,12 +160,12 @@ func (h *SidecarSetCreateHandler) Handle(ctx context.Context, req types.Request)
...
 	switch req.AdmissionRequest.Operation {
 	case v1beta1.Create, v1beta1.Update:
-		setDefaultSidecarSet(copy)
+		SetDefaultSidecarSet(copy)
 		if err := setHashSidecarSet(copy); err != nil {
 			return admission.ErrorResponse(http.StatusInternalServerError, err)
 		}
 	}
-
+	klog.V(3).Infof("sidecarset after mutating: %v", util.DumpJSON(copy))
```
jian-he: should this log be removed or increase its log level

Fei-Guo: I am OK with increasing the lvl.

jian-he:
LGTM

PR #163, opened 2019-11-13
Remove duplicate roles

damnever:


FillZpp:
Actually, `config/rbac/rbac_role.yaml` is generated by `controller-tools`, when execute `make manifests`. 
So there is no need and also invalid to modify this file directly.

PR #171, opened 2019-11-27
Revise README.md to notify newest updates

Fei-Guo:

Update README.md so that users can be aware of Kruise latest updates.


xiang90:
lgtm

PR #179, opened 2019-12-11
Support gate for custom resource

FillZpp:

Support gate for custom resource. 

The controllers and webhooks for a CRD will not be started, if:

1. This CRD does not exist in kueb-apiserver.
2. `CUSTOM_RESOURCE_ENABLE` env is set and this CRD kind is not contained in it.

### DIFF
--- a/pkg/util/gate/custom_resource_gate.go
+++ b/pkg/util/gate/custom_resource_gate.go
@@ -1,0 +1,95 @@
+/*
+Copyright 2019 The Kruise Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package gate
+
+import (
+	"os"
+	"strings"
+
+	"github.com/openkruise/kruise/pkg/apis"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/discovery"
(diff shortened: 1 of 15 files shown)


pkg/util/gate/custom_resource_gate.go
```diff
@@ -0,0 +1,78 @@
+package gate
```
Fei-Guo: + file header

pkg/util/gate/custom_resource_gate.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		klog.Errorf("unable to set up client config: %v", err)
+		os.Exit(1)
+	}
+	discoveryClient = discovery.NewDiscoveryClientForConfigOrDie(cfg)
+}
+
+// ResourceEnabled help runnable check if the custom resource is valid and enabled
+// 1. If this CRD is not found from kueb-apiserver, it is invalid.
+// 2. If 'CUSTOM_RESOURCE_ENABLE' env is not empty and this CRD kind is not in
```
Fei-Guo: 2.   .... this CRD kind is not in $CUSTOM_RESOURCE_ENABLE.

pkg/util/gate/custom_resource_gate.go
```diff
@@ -0,0 +1,78 @@
...
+	resourceList, err := discoveryClient.ServerResourcesForGroupVersion(gvk.GroupVersion().String())
+	if err != nil {
+		klog.Warningf("custom resource gate not found groupVersionKind %v in discovery: %v", gvk, err)
+		return false
+	}
+
+	for _, r := range resourceList.APIResources {
+		if r.Kind == gvk.Kind {
+			return true
+		}
```
Fei-Guo: Do we need to check group and version as well?

FillZpp: There is no need to check group and version, this `resourceList` is got from discovery by `GroupVersion`.

```
discoveryClient.ServerResourcesForGroupVersion(gvk.GroupVersion().String())
```

pkg/util/gate/custom_resource_gate.go
FillZpp: done

pkg/util/gate/custom_resource_gate.go
FillZpp: done

PR #183, opened 2019-12-12
Stop cloneset syncing when terminating

yuyulei:
Early return if cloneset is terminating.

NONE

### DIFF
--- a/pkg/controller/cloneset/scale/cloneset_scale.go
+++ b/pkg/controller/cloneset/scale/cloneset_scale.go
@@ -9,7 +9,7 @@ import (
 	"k8s.io/klog"
 
 	"k8s.io/apimachinery/pkg/util/sets"
-
+	utilpointer "k8s.io/utils/pointer"
 	appsv1alpha1 "github.com/openkruise/kruise/pkg/apis/apps/v1alpha1"
 	clonesetutils "github.com/openkruise/kruise/pkg/controller/cloneset/utils"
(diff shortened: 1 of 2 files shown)


pkg/controller/cloneset/scale/cloneset_scale.go
```diff
@@ -55,9 +55,13 @@ func (r *realControl) ManageReplicas(
 	pods []*v1.Pod, pvcs []*v1.PersistentVolumeClaim,
 ) (bool, error) {
 	if updateCS.Spec.Replicas == nil {
-		return false, fmt.Errorf("spec.Replicas is nil")
+		// TODO: add webook mutate to set deault value to Replicas
+		updateCS.Spec.Replicas = utilpointer.Int32Ptr(0)
```
Fei-Guo: In Statefulset,  the default is 1. // If unspecified, defaults to 1.

Should we use 1 as well?

FillZpp:
`CloneSet` is unfinished yet. Actually, `spec.replicas` will be set default value in webhook, and `deletionTimestamp` will be checked in next [PR](https://github.com/FillZpp/kruise/blob/cloneset-update/pkg/controller/cloneset/cloneset_controller.go#L249) .

You're most welcome to contribute in Kruise, but it is suggested that you can start with those released resources, e.g. AdvancedStatefulSet and SidecarSet.

yuyulei:
@FillZpp ok, I will close this pr.

PR #207, opened 2020-01-06
feat:  reduce reference path of AddToScheme

LeoLiuYan:
As we know,  v1alpha1.SchemeBuilder.AddToScheme is the same to v1alpha1.AddToScheme,  the reference path of AddToScheme should more concise.


FillZpp:
@LeoLiuYan Sorry, this file is generated by `kubebuilder`. We should not modify it unless it is absolutely necessary.

LeoLiuYan:
@FillZpp  ok.

PR #222, opened 2020-01-22
Avoid to use nil to update status

yuyulei:

Follow #164 with [code](https://github.com/openkruise/kruise/blob/dab47dcb0b863df867de87ce6d82eab93208791b/pkg/controller/statefulset/stateful_set_control.go#L109)
```
        status, getStatusErr := ssc.updateStatefulSet(set, currentRevision, updateRevision, collisionCount, pods, revisions)
	updateStateusErr := ssc.updateStatefulSetStatus(set, status)
```

if status is nil,  at last, when checking if sts is completed in [completeRollingUpdate](https://github.com/openkruise/kruise/blob/dab47dcb0b863df867de87ce6d82eab93208791b/pkg/controller/statefulset/stateful_set_utils.go#L381), it will panic, so try to avoid it.

### DIFF
--- a/pkg/controller/statefulset/stateful_set_control.go
+++ b/pkg/controller/statefulset/stateful_set_control.go
@@ -278,11 +278,11 @@ func (ssc *defaultStatefulSetControl) updateStatefulSet(
 	// get the current and update revisions of the set.
 	currentSet, err := ApplyRevision(set, currentRevision)
 	if err != nil {
-		return nil, err
+		return set.Status.DeepCopy(), err
 	}
 	updateSet, err := ApplyRevision(set, updateRevision)
 	if err != nil {
-		return nil, err
+		return set.Status.DeepCopy(), err
 	}
 
 	// set the generation, and revisions in the returned status


FillZpp:
There is no need to do this, maybe just put an error check between `ssc.updateStatefulSet()` and `ssc.updateStatefulSetStatus`.

Such as:

```
status, err := ssc.updateStatefulSet(set, currentRevision, updateRevision, collisionCount, pods, revisions)
if err != nil {
    return err
}

if err := ssc.updateStatefulSetStatus(set, status); err != nil {
    return err
}
```

yuyulei:
@FillZpp , yours is conflict with #164, the pr tries to update conditions even when some error happens.

FillZpp:
Ok, how about this:

```
currentSet, err := ApplyRevision(set, currentRevision)
if err != nil {
	return nil, err
}
updateSet, err := ApplyRevision(set, updateRevision)
if err != nil {
	return nil, err
}
```

modify to

```
currentSet, err := ApplyRevision(set, currentRevision)
if err != nil {
	return &set.Status, err
}
updateSet, err := ApplyRevision(set, updateRevision)
if err != nil {
	return &set.Status, err
}
```

yuyulei:
If following your suggest,  func `updateStatefulSet` returns:

`situation 1`: a point of (statefulset's) status from cache, like
```
currentSet, err := ApplyRevision(set, currentRevision)
if err != nil {
	return &set.Status, err
}
```

or

`situation 2`: a point of new object, which has nothing with cache, like
```
if set.DeletionTimestamp != nil {
	return status, nil
}
```

But the ways to edit the two points are different. The status from `situation 1` should deep copy before editing content. I think unifying to one form is better. 

What is your opinion ? @FillZpp

yuyulei:
@FillZpp, follow your suggestion and make a small change, ptal.

PR #246, opened 2020-04-14
Add CloneSet core interfaces

FillZpp:
Add CloneSet core interfaces


pkg/util/inplaceupdate/inplace_utils.go
```diff
@@ -327,6 +332,13 @@ func calculateInPlaceUpdateSpec(oldRevision, newRevision *apps.ControllerRevisio
 		Revision:        newRevision.Name,
 		ContainerImages: make(map[string]string),
 	}
+
+	if opts != nil && opts.CustomizeUpdateFunc != nil {
```
Fei-Guo, line 336: Maybe we should leave a comment here. The behavior is not consistent with the calculateInPlaceUpdateSpec method comment above.

PR #274, opened 2020-05-21
add advanced daemonset as a new member in Kruise Project

### LINKED ISSUE #175: [feature request] Add Advanced DaemonSet in kruise project
I want to import an advanced daemonset in this open source project.

**Why is this needed**:
This new added workload will help users have more deploy and update strategies comparing to using native DaemonSet.

zhangxiaoyu-zidif:
I want to create a new controller to enhance original DaemonSet, which is called `Advanced DaemonSet`.

fix #175

This is the first PR for API

@FillZpp  @Fei-Guo 

Different the original daemonset, this one almost rewrite the original logic and add two more rollingupdate types. 
Even for the original one, we also change some APIs to make it be  managed and controlled at the first deployment and update processes. 

I will send a new PR to explain how to use it and what's new in this new workload.


pkg/apis/apps/v1alpha1/daemonset_types.go
```diff
@@ -0,0 +1,270 @@
...
+	CollisionCount *int32 `json:"collisionCount,omitempty" protobuf:"varint,9,opt,name=collisionCount"`
+
+	// Represents the latest available observations of a DaemonSet's current state.
+	// +optional
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	Conditions []DaemonSetCondition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,10,rep,name=conditions"`
+
+	// DaemonSetHash is the controller-revision-hash, which represents the latest version of the DaemonSet.
+	DaemonSetHash string `json:"daemonSetHash" protobuf:"bytes,11,opt,name=daemonSetHash"`
```
Fei-Guo, line 209: How about DaemonSetHash->UpdateRevision? The latter is a commonly used in k8s workload controller.

pkg/apis/apps/v1alpha1/daemonset_types.go
```diff
@@ -0,0 +1,270 @@
...
+	// Example: when this is set to 30%, at most 30% of the total number of nodes
+	// that should be running the daemon pod (i.e. status.desiredNumberScheduled)
+	// can have their pods stopped for an update at any given
+	// time. The update starts by stopping at most 30% of those DaemonSet pods
+	// and then brings up new DaemonSet pods in their place. Once the new pods
+	// are available, it then proceeds onto other DaemonSet pods, thus ensuring
+	// that at least 70% of original number of DaemonSet pods are available at
+	// all times during the update.
+	// +optional
+	MaxUnavailable *intstr.IntOrString `json:"maxUnavailable,omitempty" protobuf:"bytes,2,opt,name=maxUnavailable"`
```
Fei-Guo, line 80: May be we can add a comment say that MaxUnavailable does not take effect if StandardRollingUpdateType is specified.

pkg/apis/apps/v1alpha1/daemonset_types.go
```diff
@@ -0,0 +1,270 @@
...
+	// of DaemonSet pods at the start of the update (ex: 10%). The absolute number is calculated from
+	// the percentage by rounding up. This cannot be 0. The default value is 1. Example: when this is
+	// set to 30%, at most 30% of the total number of nodes that should be running the daemon pod
+	// (i.e. status.desiredNumberScheduled) can have 2 pods running at any given time. The update
+	// starts by starting replacements for at most 30% of those DaemonSet pods. Once the new pods are
+	// available it then stops the existing pods before proceeding onto other DaemonSet pods, thus
+	// ensuring that at most 130% of the desired final number of DaemonSet  pods are running at all
+	// times during the update.
+	// +optional
+	MaxSurge *intstr.IntOrString `json:"maxSurge,omitempty" protobuf:"bytes,7,opt,name=maxSurge"`
```
Fei-Guo, line 110: What is the relationship between  "MaxSurge" and "MaxUnavailable" ? 
Will the smaller value between these two determines the number of updated nodes in parallel? 

Also if the new pod fails but the old pod still runs, will the new pod be counted as an unavailable pod and taken into count when calculating MaxUnavailable?

pkg/apis/apps/v1alpha1/daemonset_types.go
zhangxiaoyu-zidif: At the early design, the advanced DaemonSet focus on keeping the same behaviors as native DaemonSet as possible, so many API types are copied from KCM directly.

pkg/apis/apps/v1alpha1/daemonset_types.go
zhangxiaoyu-zidif: Yes, in fact , `MaxUnavailable` only work in the cases that  RollingUpdateType is "StandardRollingUpdateType" or  "InplaceRollingUpdateType", not work for SurgingRollingUpdateType.

pkg/apis/apps/v1alpha1/daemonset_types.go
zhangxiaoyu-zidif: Yes, here is a place which alway cause confusion，I will send an issue to discuss those APIs. We will avoid both of them work at same time by using webhook validations.

zhangxiaoyu-zidif:
this PR is blocked by go.mod updated. It will continue after conflict being resolved. @FillZpp

zhangxiaoyu-zidif:
PTAL @FillZpp

pkg/controller/daemonset/update.go
```diff
@@ -0,0 +1,708 @@
...
+				klog.V(6).Infof("Surging new pod on node %s", node)
+				numSurge++
+				nodesToSurge = append(nodesToSurge, node)
+			}
+		} else if newPod != nil && podutil.IsPodAvailable(newPod, ds.Spec.MinReadySeconds, metav1.Now()) && oldPod != nil && oldPod.DeletionTimestamp == nil {
+			klog.V(6).Infof("Marking pod %s/%s for deletion", ds.Name, oldPod.Name)
+			oldPodsToDelete = append(oldPodsToDelete, oldPod.Name)
+		} else {
+			if ds.Spec.MinReadySeconds > 0 {
+				// TODO: use return reconcile.Result{RequeueAfter: time.Duration(ds.Spec.MinReadySeconds) * time.Second}, nil
```
Fei-Guo, line 691: I think we should not sleep here.

pkg/controller/daemonset/inplace_update.go
```diff
@@ -0,0 +1,143 @@
...
+				if refreshRes.RefreshErr != nil {
+					klog.Errorf("failed to update pod condition: %v", err)
+				}
+				pod.Labels[apps.DefaultDaemonSetUniqueLabelKey] = hash
+				dsc.client.Update(context.TODO(), pod)
+				return
+			}
+
+			if res.UpdateErr != nil {
+				klog.Warningf("CloneSet %s/%s failed to in-place update Pod %s, so it will back off to ReCreate", ds.Namespace, ds.Name, pod.Name)
```
Fei-Guo, line 128: Nit. Cloneset-> DaemonSet?

zhangxiaoyu-zidif: Good catch

pkg/controller/daemonset/update.go
zhangxiaoyu-zidif: yes, the sleep is so ugly, and i remove it at once.

PR #281, opened 2020-06-08
add api for imagepulljob

hormes:

Add a job interface for image pull, which is used to warm up the typical basic images in the cluster, accelerate the efficiency of Pod startup, and accelerate the efficiency of CloneSet application upgrade.


pkg/apis/apps/v1alpha1/imagepullnode_types.go
```diff
@@ -0,0 +1,193 @@
...
+// +kubebuilder:printcolumn:name="DESIRED",type="integer",JSONPath=".status.desired",description="Number of all images on this node"
+// +kubebuilder:printcolumn:name="PULLING",type="integer",JSONPath=".status.pulling",description="Number of image pull task active"
+// +kubebuilder:printcolumn:name="SUCCEED",type="integer",JSONPath=".status.succeeded",description="Number of image pull task succeeded"
+// +kubebuilder:printcolumn:name="FAILED",type="integer",JSONPath=".status.failed",description="Number of image pull tasks failed"
+// +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp",description="CreationTimestamp is a timestamp representing the server time when this object was created. It is not guaranteed to be set in happens-before order across separate operations. Clients may not set this value. It is represented in RFC3339 form and is in UTC."
+type ImagePullNode struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   ImagePullNodeSpec   `json:"spec,omitempty"`
```
Fei-Guo, line 178: Do you need to specify the node name in the spec? or how will the node daemon find the corresponding crd?

pkg/apis/apps/v1alpha1/imagepullnode_types.go
```diff
@@ -0,0 +1,193 @@
...
+
+// ImagePullNode is the Schema for the imagepullnodes API
+// +k8s:openapi-gen=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="DESIRED",type="integer",JSONPath=".status.desired",description="Number of all images on this node"
+// +kubebuilder:printcolumn:name="PULLING",type="integer",JSONPath=".status.pulling",description="Number of image pull task active"
+// +kubebuilder:printcolumn:name="SUCCEED",type="integer",JSONPath=".status.succeeded",description="Number of image pull task succeeded"
+// +kubebuilder:printcolumn:name="FAILED",type="integer",JSONPath=".status.failed",description="Number of image pull tasks failed"
+// +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp",description="CreationTimestamp is a timestamp representing the server time when this object was created. It is not guaranteed to be set in happens-before order across separate operations. Clients may not set this value. It is represented in RFC3339 form and is in UTC."
+type ImagePullNode struct {
```
Fei-Guo, line 174: How about ImagePullNode -> NodeImages?

pkg/apis/apps/v1alpha1/imagepullnode_types.go
```diff
@@ -0,0 +1,193 @@
...
+// ImagePullSpec defines the pulling spec of an image
+type ImagePullSpec struct {
+	// ImagePullSecrets is an optional list of references to secrets in the same namespace to use for pulling the image.
+	// If specified, these secrets will be passed to individual puller implementations for them to use.  For example,
+	// in the case of docker, only DockerConfig type secrets are honored.
+	// +optional
+	PullSecrets []types.NamespacedName `json:"pullSecrets,omitempty"`
+
+	// Tags is a list of versions of this image
+	Tags []ImageTagPullSpec `json:"tags"`
```
Fei-Guo, line 40: Does the ImagePullJob needs to specify the tags info as well?

pkg/apis/apps/v1alpha1/imagepullnode_types.go
```diff
@@ -0,0 +1,193 @@
...
+	PullPolicy *ImageTagPullPolicy `json:"pullPolicy,omitempty"`
+
+	// List of objects depended by this object. If this image is managed by a controller,
+	// then an entry in this list will point to this controller.
+	// +optional
+	OwnerReferences []v1.ObjectReference `json:"ownerReferences,omitempty"`
+}
+
+// ImageTagPullPolicy defines the policy of the pulling task
+type ImageTagPullPolicy struct {
```
Fei-Guo, line 64: Why is the ImageTagPullPolicy different from the PullPolicy in ImagePullJob? I assume they can be the same and share the same definition.

pkg/apis/apps/v1alpha1/imagepullnode_types.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// PullPolicy is an optional field to set parameters of the pulling task. If not specified,
+	// the system will use the default values.
+	// +optional
+	PullPolicy *ImageTagPullPolicy `json:"pullPolicy,omitempty"`
+
+	// List of objects depended by this object. If this image is managed by a controller,
+	// then an entry in this list will point to this controller.
+	// +optional
+	OwnerReferences []v1.ObjectReference `json:"ownerReferences,omitempty"`
```
Fei-Guo, line 60: I suggest to remove OwnerReferences in the first version if this is used to coordinate with workload controller.

pkg/apis/apps/v1alpha1/imagepulljob_types.go
```diff
@@ -0,0 +1,144 @@
...
+	Parallelism *intstr.IntOrString `json:"parallelism,omitempty"`
+
+	// PullPolicy is an optional field to set parameters of the pulling task. If not specified,
+	// the system will use the default values.
+	// +optional
+	PullPolicy *PullPolicy `json:"pullPolicy,omitempty"`
+
+	// CompletionPolicy indicates the completion policy of the job.
+	// Default is Always CompletionPolicyType.
+	CompletionPolicy CompletionPolicy `json:"completionPolicy"`
```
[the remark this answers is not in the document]
hormes: Reuse the definition in broadcastjob

pkg/apis/apps/v1alpha1/imagepullnode_types.go
hormes: Rename it to ImagePullNode mainly because we have an object called Nodeimage inside, and both of them will exist for a while.

pkg/apis/apps/v1alpha1/imagepullnode_types.go
hormes: The CR will be automatically created by the controller, and use the node name as the CR name.

pkg/apis/apps/v1alpha1/imagepullnode_types.go
hormes: Job and `Task` may still be some different attributes. I'm thinking of moving ImagePullNode to an internal package. This CR doesn't want to be handed over to users, only to use it internally.

pkg/apis/apps/v1alpha1/imagepullnode_types.go
hormes: `OwnerReferences` is used to coordinate with ImagePullJob, and send an event to it when there are some errors happen.

pkg/apis/apps/v1alpha1/nodeimage_types.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// ImageStatus defines the pulling status of an image
+type ImageStatus struct {
+	// Represents statuses of pulling tasks on this node
+	Tags []ImageTagStatus `json:"tags"`
+}
+
+// ImageTagStatus defines the pulling status of an image tag
+type ImageTagStatus struct {
```
FillZpp, line 120: Cloud we add `imageID` in ImageTagStatus? It is really useful for in-place update.

vsxen:
new crd ? the `config/crds/` dir isn't update?

PR #284, opened 2020-06-09
Use go modules instead of dep

FillZpp:

Use go modules instead of dep


.gitignore
```diff
@@ -17,6 +17,9 @@ bin
 
 !vendor/**/zz_generated.*
 
+# go vendor
+vendor/
```
hormes, line 21: Why remove the vendor directory？This may make it very difficult to compile kruise in a new environment.

PR #317, opened 2020-07-21
fix cloneset maxUnavaliable when maxSurge > 0

### LINKED ISSUE #316: [BUG] CloneSet update not expected
**What happened**:
Create a cloneset with replicas 1, and updateStrategy as follows:  
```
updateStrategy:
    maxSurge: 50%
    maxUnavailable: 10%
    partition: 0
    type: ReCreate
```
doing an update, it will delete and create new pod at the same time.

After the newly pod ready，delete the old pod.

**Environment**:
- Kruise version: master
- Kubernetes version (use `kubectl version`): 1.17.4
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

shiyan2016:
fix cloneset maxUnavaliable when maxSurge is not nil

fixes #316

### DIFF
--- a/pkg/controller/cloneset/update/cloneset_update.go
+++ b/pkg/controller/cloneset/update/cloneset_update.go
@@ -141,8 +141,13 @@ func calculateUpdateCount(coreControl clonesetcore.Control, strategy appsv1alpha
 	}
 	waitUpdateIndexes = waitUpdateIndexes[:(len(waitUpdateIndexes) - partition)]
 
+	roundUp := true
+	if strategy.MaxSurge != nil {
+		maxSurge, _ := intstrutil.GetValueFromIntOrPercent(strategy.MaxSurge, totalReplicas, true)
+		roundUp = maxSurge == 0
+	}
 	maxUnavailable, _ := intstrutil.GetValueFromIntOrPercent(
-		intstrutil.ValueOrDefault(strategy.MaxUnavailable, intstrutil.FromString(appsv1alpha1.DefaultCloneSetMaxUnavailable)), totalReplicas, true)
+		intstrutil.ValueOrDefault(strategy.MaxUnavailable, intstrutil.FromString(appsv1alpha1.DefaultCloneSetMaxUnavailable)), totalReplicas, roundUp)
 	usedSurge := len(pods) - totalReplicas
 
 	var notReadyCount, updateCount int
--- a/pkg/apis/apps/v1alpha1/cloneset_types.go
+++ b/pkg/apis/apps/v1alpha1/cloneset_types.go
@@ -89,7 +89,8 @@ type CloneSetUpdateStrategy struct {
 	Partition *int32 `json:"partition,omitempty"`
 	// The maximum number of pods that can be unavailable during the update.
 	// Value can be an absolute number (ex: 5) or a percentage of desired pods (ex: 10%).
-	// Absolute number is calculated from percentage by rounding up.
+	// Absolute number is calculated from percentage by rounding up by default.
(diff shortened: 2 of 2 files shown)


pkg/controller/cloneset/update/cloneset_update.go
```diff
@@ -143,6 +143,10 @@ func calculateUpdateCount(coreControl clonesetcore.Control, strategy appsv1alpha
 
 	maxUnavailable, _ := intstrutil.GetValueFromIntOrPercent(
 		intstrutil.ValueOrDefault(strategy.MaxUnavailable, intstrutil.FromString(appsv1alpha1.DefaultCloneSetMaxUnavailable)), totalReplicas, true)
+	if strategy.MaxSurge != nil {
```
FillZpp, line 146: How about this

```golang
roundUp := strategy.MaxSurge == nil
maxUnavailable, _ := intstrutil.GetValueFromIntOrPercent(
    intstrutil.ValueOrDefault(strategy.MaxUnavailable, intstrutil.FromString(appsv1alpha1.DefaultCloneSetMaxUnavailable)), totalReplicas, roundUp)
```

shiyan2016: done, the ci check was failed by timeout, how to recheck it ?

pkg/controller/cloneset/update/cloneset_update.go
```diff
@@ -141,8 +141,9 @@ func calculateUpdateCount(coreControl clonesetcore.Control, strategy appsv1alpha
 	}
 	waitUpdateIndexes = waitUpdateIndexes[:(len(waitUpdateIndexes) - partition)]
 
+	roundUp := strategy.MaxSurge == nil
```
FillZpp, line 144: My fault. It is meaningless to compare with `nil`, because we have set the `maxSurge` defaults to `0` 

https://github.com/openkruise/kruise/blob/master/pkg/apis/apps/v1alpha1/defaults.go#L223

shiyan2016: so we need to check whether maxSurge is 0, I will update it.

pkg/controller/cloneset/update/cloneset_update.go
```diff
@@ -141,8 +141,9 @@ func calculateUpdateCount(coreControl clonesetcore.Control, strategy appsv1alpha
 	}
 	waitUpdateIndexes = waitUpdateIndexes[:(len(waitUpdateIndexes) - partition)]
 
+	roundUp := !(strategy.MaxSurge != nil && strategy.MaxSurge.IntValue() > 0)
```
FillZpp, line 144: The question is, if `maxSurge` is percentage str, `maxSurge.IntValue()` will also return 0.

How about this:

```golang
roundUp := true
if strategy.MaxSurge != nil {
    roundUp = strategy.MaxSurge.Type == intstr.Int && strategy.MaxSurge.IntValue() == 0
}
```

shiyan2016: But it cannot cover string type case.

shiyan2016: how about this:
` 
roundUp := true
if strategy.MaxSurge != nil {
   maxSurge, _ := intstrutil.GetValueFromIntOrPercent(strategy.MaxSurge, totalReplicas, true)
   roundUp = maxSurge == 0
}

`

shiyan2016: @FillZpp

FillZpp: @shiyan2016 That's ok

Issue #280, opened 2020-06-08
[feature request] Add healthcheck / readiness endpoints


servo1x:
**What would you like to be added**:

The controller should provide a `/healthz` endpoint for healthchecking.

**Why is this needed**:

Currently the healthcheck used is TCP and it clutters the logs with messages such as:

```bash
2020-06-08T08:56:35.288755743Z 2020/06/08 08:56:35 http: TLS handshake error from 10.33.136.236:59752: EOF
2020-06-08T08:56:45.288596633Z 2020/06/08 08:56:45 http: TLS handshake error from 10.33.136.236:59834: EOF
2020-06-08T08:56:55.288749966Z 2020/06/08 08:56:55 http: TLS handshake error from 10.33.136.236:59924: EOF
2020-06-08T08:57:02.669495394Z 2020/06/08 08:57:02 http: TLS handshake error from 10.233.76.190:54688: tls: first record does not look like a TLS handshake
2020-06-08T08:57:05.288829307Z 2020/06/08 08:57:05 http: TLS handshake error from 10.33.136.236:60018: EOF
2020-06-08T08:57:15.288654101Z 2020/06/08 08:57:15 http: TLS handshake error from 10.33.136.236:60100: EOF
2020-06-08T08:57:25.28873608Z 2020/06/08 08:57:25 http: TLS handshake error from 10.33.136.236:60178: EOF
2020-06-08T08:57:32.669429178Z 2020/06/08 08:57:32 http: TLS handshake error from 10.233.76.190:54938: tls: first record does not look like a TLS handshake
```

It can be confusing to a user.

FillZpp:
@servo1x Sounds reasonable. I will improve it, thanks!

FillZpp:
Already added in `v0.6.0`.

Issue #225, opened 2020-02-03
best practice for webhook service


luodw:
Is it a recommended way to use  webhook to inject sidecar container? If yes, webhook service must be very stable.  If webhook has something wrong, It will be a disaster, as it can not create pod any more. So there are some other best practice?

Fei-Guo:
We do, I think we can set up multiple replica for controller manager, aka, the webhook service. @FillZpp may add more about this.

FillZpp:
Fixed by #321

PR #321, opened 2020-07-31
Update controller-runtime and rebuild project structure

FillZpp:

Update controller-runtime and rebuild project structure.

- Update K8s packages to 1.16+ (only the dependency, we still support K8s 1.12+ for deploying and using)
- Update controller-runtime to v0.5.7
- Support webhook multiple replicas for wehook server

Fixes:
- #280 
- #225


Fei-Guo:
This is the infra change that we have to go through. Let me approve first and make tunings later.

Issue #329, opened 2020-08-11
[feature request]Sidecar Resource is calculated according to the proportion of POD Resource.


peacocktrain:
**What would you like to be added**:
SidecarSet，Sidecar Resource is calculated according to the proportion of POD Resource.

**Why is this needed**:
We have some thoughts on the sidecar resource limit.  Business containers consume more resources, so sidecar should also consume more resources than other Pods. so sidecar resource should be set up in two ways，defined and proportional

Fei-Guo:
This sounds to be a reasonable request.  @ryanzhang-oss FYI.

ryanzhang-oss:
@peacocktrain  I think this is a reasonable ask. However, I wonder if you can be a bit more specific on the use case? Like exactly what does the sidecar is trying to achieve?

peacocktrain:
> @peacocktrain I think this is a reasonable ask. However, I wonder if you can be a bit more specific on the use case? Like exactly what does the sidecar is trying to achieve?

Network sidecars，like  envoy.  Logs sidecars. We give sidecar to all pods，but 4C pod flow is significantly higher than 2C

ryanzhang-oss:
What is a 4C pod? Is the sidecar's resource proportional to the total resource of the pod or the "main" container?

ryanzhang-oss:
@peacocktrain Another question is how do you plan to consolidate the differences between this value proportional to the size of the pod and the `Requests` and `Limits` value set in the sidecar container. Will this value replace them all?

peacocktrain:
> What is a 4C pod? Is the sidecar's resource proportional to the total resource of the pod or the "main" container?

“main” container

peacocktrain:
> @peacocktrain Another question is how do you plan to consolidate the differences between this value proportional to the size of the pod and the `Requests` and `Limits` value set in the sidecar container. Will this value replace them all?

In our scenario, limit=request because Sidecar does not change the POD's qos

peacocktrain:
> What is a 4C pod? Is the sidecar's resource proportional to the total resource of the pod or the "main" container?

emmm. 4C is the total requested resource of the pod, This value participates in the schedule

ryanzhang-oss:
> In our scenario, limit=request because Sidecar does not change the POD's qos

now we have 3/4 values,  you can set the Requests and Limits value set in the sidecar container spec with the additional value proportional to the "main" container. Do you want to void the Requests and Limits value set in the sidecar container spec and use the additional value alone?

[...]

PR #334, opened 2020-08-13
sidecar proportional resource

ryanzhang-oss:
The API design to enable proportional resource for sidecarset

https://github.com/openkruise/kruise/issues/329

### III. Special notes for reviews
Let's decide on the interface first


apis/apps/v1alpha1/sidecarset_types.go
```diff
@@ -43,6 +43,10 @@ type SidecarSetSpec struct {
 // SidecarContainer defines the container of Sidecar
 type SidecarContainer struct {
 	corev1.Container `json:",inline"`
+	// the sidecar container resource requirement that is proportional to the pod total resource
+	// this will override the corresponding resources fields in the container with the same resourceName
+	// the key of the map is the percentage
+	ProportionalResource map[corev1.ResourceName]uint8 `json:"proportionalResource,omitempty"`
```
Fei-Guo, line 49: - This field should be optional.
- use int32 instead of uint8

You should explain clearly what the int value should be. E.g., it is a value from 1 - 100. 

One more thought, maybe the algorithm of calculating the size should be `size = max(container.spec.request,  main-containers.request * x /100)`

ryanzhang-oss: > * This field should be optional.

isn't this optional already?

ryanzhang-oss: > * use int32 instead of uint8

try to save some space, is there any advantage of using int32?

ryanzhang-oss: > e size should be `size = max(container.spec.request, main-containers.request * x /100)`

What is "main" here?  I don't fully get the idea and it could be too complicated for general users.

Issue #293, opened 2020-06-22
[feature request] support minReadySeconds in StatefulSet workload


Yetsenn:
**What would you like to be added**:
```
apiVersion: apps.kruise.io/v1alpha1
kind: StatefulSet
spec:
  # ...
  minReadySeconds: 60
```

**Why is this needed**:
In my case, I want a process like:
1. Set minReadySeconds=60 and MaxUnavailable=1.
2. Update statefulSet for an image, and the container in first Pod will be updated, restarted and ready.
3. Then after minReadySeconds, continue to update the second Pod.

This effectively works for container warm-up.

ryanzhang-oss:
Does `minReadySeconds` apply to the scaleup/down too?

PR #336, opened 2020-08-14
add min ready setting

ryanzhang-oss:
Implement https://github.com/openkruise/kruise/issues/293

### Ⅱ. Does this pull request fixes one issue?
https://github.com/openkruise/kruise/issues/293

Added UT test for scale and upgrade statefulset with a minReadySeconds not zero.

Here are the manual tests using the guestbook demo
Here are the common steps
1. make docker-build 
2. modify the v0.6.0 helm charts value in charts/kruise to use the newly built docker image
3. helm install kruise charts/kruise/v0.6.0

- Scaleup Test with 
  - podManagementPolicy = Parallel 
  - podManagementPolicy = OrderedReady

- Update Test with 
  - podManagementPolicy = Parallel, MaxUnavailable = 3,
  - podManagementPolicy = Parallel, MaxUnavailable = 1,
  - podManagementPolicy = OrderedReady

Results are all expected as the new pods will wait for the previous pods to be ready for minReadySeconds.


Fei-Guo:
/lgtm

Nit: Please be more specific about the PR  title.  e.g.,  Add `MinReadySeconds` API definition in AdvancedStatefulSet

FillZpp:
@ryanzhang-oss Could you please add the implement code in statefulset controller in this PR?

ryanzhang-oss:
> @ryanzhang-oss Could you please add the implement code in statefulset controller in this PR?

yes, I am working on it. Just want to make sure that the API design is right first.

pkg/util/inplaceupdate/adapter.go
```diff
@@ -74,7 +74,11 @@ type adapterInformer struct {
 }
 
 func (c *adapterInformer) getPod(namespace, name string) (*v1.Pod, error) {
```
Fei-Guo, line 76: We usually do the DeepCopy in the caller side since not all getPod callers need to revise the object.

pkg/controller/statefulset/stateful_set_utils.go
```diff
@@ -180,6 +182,30 @@ func updateIdentity(set *appsv1alpha1.StatefulSet, pod *v1.Pod) {
 	pod.Labels[apps.StatefulSetPodNameLabel] = pod.Name
 }
 
+// isRunningAndAvailable returns true if pod is in the PodRunning Phase,
+// and it has a condition of PodReady for a minimum of minReadySeconds.
+// return true if it's available
+// return false with zero means it's not ready
+// return false with a positive value means it's not available and should recheck with that time
+func isRunningAndAvailable(pod *v1.Pod, minReadySeconds int32) (bool, time.Duration) {
```
Fei-Guo, line 190: I would suggest the first returned boolean indicates "RunningAndRunning" and the 2nd returned value indicates the wait time. I think this makes the code easier to understand.

pkg/controller/statefulset/stateful_set_utils.go
```diff
@@ -180,6 +182,30 @@ func updateIdentity(set *appsv1alpha1.StatefulSet, pod *v1.Pod) {
...
+	if pod.Status.Phase != v1.PodRunning || !podutil.IsPodReady(pod) {
+		return false, 0
+	}
+	c := podutil.GetPodReadyCondition(pod.Status)
+	minReadySecondsDuration := time.Duration(minReadySeconds) * time.Second
+	if minReadySeconds == 0 {
+		return true, 0
+	}
+	if c.LastTransitionTime.IsZero() {
```
Fei-Guo, line 199: Is this possible to happen if the Pod status is ready but the transition time is 0?

pkg/controller/statefulset/stateful_set_control.go
```diff
@@ -595,13 +613,24 @@ func (ssc *defaultStatefulSetControl) updateStatefulSet(
...
 				completedErr)
 			unavailablePods = append(unavailablePods, replicas[target].Name)
+		} else {
+			// check if the updated pod is running and available given minReadySeconds
+			if isAvailable, waitTime := isRunningAndAvailable(replicas[target], minReadySeconds); !isAvailable {
+				// the pod is not available yet given the minReadySeconds requirement
+				unavailablePods = append(unavailablePods, replicas[target].Name)
+				// make sure that
+				if waitTime != 0 {
+					durationStore.Push(getStatefulSetKey(set), waitTime)
```
Fei-Guo, line 631: Note that this is within a loop so the waitTime should the mininal waitTime across all unavailable Pods.  The durationStore.Push should happen outside of the loop.

pkg/controller/statefulset/stateful_set_utils.go
ryanzhang-oss: > RunningAndRunning

what is "RunningAndRunning"? Currently, it means if it's "RunningAndAvailable"

pkg/controller/statefulset/stateful_set_utils.go
ryanzhang-oss: not sure, this case presents in the `IsPodReady` function in the  "k8s.io/kubernetes/pkg/api/v1/pod"

pkg/controller/statefulset/stateful_set_control.go
ryanzhang-oss: The loop is short-circuited by the return so we can't put the push outside of the loop but I will use the min time.

pkg/controller/statefulset/stateful_set_utils.go
Fei-Guo: Sorry, typo, I mean RunningAndReady.

pkg/util/inplaceupdate/adapter.go
ryanzhang-oss: There is a bug here. The adapterInfomer here is used by the UT and it behaves differently from the real informer which takes a deepCopy. This leads to UT can't repro the behavior in production.

Fei-Guo:
LGTM

FillZpp (approved):
LGTM

PR #339, opened 2020-08-18
Disable in-place update for Advanced DaemonSet

FillZpp:

Disable in-place update for Advanced DaemonSet.


giliao:
@FillZpp may i ask why to disable in place update. can't find related discussions
thanks.

FillZpp:
@giliao The implementation of daemonset in-place update has several race problems, and we thought in-place update may not be required because daemonset pod always recreate on the same node.

PR #356, opened 2020-08-28
Remove cloneset scale race condition on bool

reverson:
Removes a race condition in the CreatePods flow with concurrent access to the created boolean

None

identified with `go test -race .`

`go test -race .` no longer fails.

I don't believe the created boolean was used for anything other than a global flag, but if so, this might need to be updated.

### DIFF
--- a/pkg/controller/cloneset/scale/cloneset_scale.go
+++ b/pkg/controller/cloneset/scale/cloneset_scale.go
@@ -126,7 +126,6 @@ func (r *realControl) createPods(
 		podsCreationChan <- p
 	}
 
-	var created bool
 	successPodNames := sync.Map{}
 	_, err = clonesetutils.DoItSlowly(len(newPods), initialBatchSize, func() error {
 		pod := <-podsCreationChan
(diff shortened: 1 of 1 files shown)


Fei-Guo:
Thanks for digging the race problem! 

There is a slight behavior difference here. Say we want to create 3 Pods, and the first two succeed but the last one fails. 
In previous code, the method returns `true, err`. In your change, the method returns `false, err`. I haven't gone through the
code path to verify but I believe this is the reason to introduce this boolean, to identify the cases where part of the Pods succeed to create.

But this problem should be fixed.

Fei-Guo:
You may consider to change the bool to int and use atomic.AddUint64(&created, 1) to update the value.

reverson:
> You may consider to change the bool to int and use atomic.AddUint64(&created, 1) to update the value.

Updated with the suggestion, thanks!

Fei-Guo:
LGTM. Thanks

FillZpp (approved):
LGTM

Issue #349, opened 2020-08-25
[BUG] Scaling up Clonesets causes kruise-controller to crash


reverson:
**What happened**: Scaled up a cloneset deployment from 10 to 100 pods (and then from 100 to 140 pods) all at once, and the kruise-controller continuously crashed and made only a little progress on each iteration.

**What you expected to happen**: The controller would say up.

**How to reproduce it (as minimally and precisely as possible)**: Scale up a deployment from 10 to 100 (at least in my experience, there might be other circumstances in my environment).

**Anything else we need to know?**: The logs from the kruise-controller would show this.
```
fatal error: concurrent map writes
fatal error: concurrent map writes
```
I also saw something about concurrent map iteration and map writes in a prior crash

**Environment**:
- Kruise version: v0.5.0
- Kubernetes version (use `kubectl version`): 1.16
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Here is a partial thread dump from the concurrent map write crash
```
goroutine 463 [running]:
runtime.throw(0x162178a, 0x15)
        /opt/go/src/runtime/panic.go:617 +0x72 fp=0xc00da9efb8 sp=0xc00da9ef88 pc=0x42e322
runtime.mapassign(0x14503a0, 0xc00d5bf980, 0xc00dc34040, 0xc00d2d6280)
        /opt/go/src/runtime/map.go:682 +0x5c8 fp=0xc00da9f040 sp=0xc00da9efb8 pc=0x40f928
reflect.mapassign(0x14503a0, 0xc00d5bf980, 0xc00dc34040, 0xc00dc34050)
        /opt/go/src/runtime/map.go:1331 +0x3f fp=0xc00da9f070 sp=0xc00da9f040 pc=0x41154f
github.com/openkruise/kruise/vendor/github.com/modern-go/reflect2.(*UnsafeMapType).UnsafeSetIndex(...)
        /go/src/github.com/openkruise/kruise/vendor/github.com/modern-go/reflect2/unsafe_map.go:76
github.com/openkruise/kruise/vendor/github.com/json-iterator/go.(*mapDecoder).Decode(0xc00d284190, 0xc00dada270, 0xc00d2d6280)
        /go/src/github.com/openkruise/kruise/vendor/github.com/json-iterator/go/reflect_map.go:191 +0x24e fp=0xc00da9f0f0 sp=0xc00da9f070 pc=0xaadebe
[...]

PR #357, opened 2020-08-29
fix concurrent bug when create many pods concurrently

hantmac:

The ` github.com/json-iterator/go.(*stringCodec).Decode()` does not support concurrency, but when cloneSet scale, it will create pods concurrently, cause data race.
 
image

May fix https://github.com/openkruise/kruise/issues/349

go test -race .

`go test -race .` return ok.
image


Fei-Guo:
Thanks for working on this! PR#356 addresses the same problem.  Let us use that one for the fix.  Btw, I don't think the createOnePod method needs to be locked.

Issue #366, opened 2020-09-07
[feature request] use sidecarset to replace istio webhook inject


andrewshaoyu:
**What would you like to be added**:
1. use sidecarset inject initContainers

**Why is this needed**:
1. use sidecarset to replace istio webhook totally.
1. initContainers do not need inplace update
1. if mutiple sidecarsets inject mutiple initContainers in a pod , there is a simple solution that initContainers can be sorted by their name and inserted into the suitable position, the initContainer name manager by the users.

PR #371, opened 2020-09-10
Add init containers to the sidecarset

ryanzhang-oss:
Add an initContainer section to the sidecar spec so that users can use it to inject init containers alongside sidecar containers.

Yes, https://github.com/openkruise/kruise/issues/366

Unit tests are added to every code path

build and push an image of the new 
```
make docker-build docker-push IMG=nerdyyatrice/kruise:v0.1
helm install kruise charts/kruise/v0.6.0 --set manager.image.repository=nerdyyatrice/kruise,manager.image.tag=v0.1
```

Sidecar example:  config/samples/apps_v1alpha1_sidecarset.yaml 

1. Apply the sidecar with an init container that runs for 5 mins
2. Apply a statefulset
3. Observe that pods are waiting for the init container to finish
4. Change the initContainer runtime to 1 min
5. Observe that no pod changes
6. Expand the statefulSet
7. Observe that the new pods wait for 1 min instead of 5
8. Subtmit a sidecarset with only initContainers
9. Observe that no pod changes
10. Expand the statefulSet and observe that new pods has the init container
11. Add containers back to the sideCarset
12. Change both the command and images of the initContainer and the container image
13. Observe that the new version of the container image is picked up by the exiting pod

### DIFF
--- a/pkg/webhook/sidecarset/mutating/hash.go
+++ b/pkg/webhook/sidecarset/mutating/hash.go
@@ -37,12 +37,10 @@ func SidecarSetHash(sidecarSet *appsv1alpha1.SidecarSet) (string, error) {
 	return h, nil
 }
 
-// SidecarSetHashWithoutImage calculates hash without sidecars's image
+// SidecarSetHashWithoutImage calculates hash without sidecars's container image
+// This is used to determine if the sidecar reconcile needs to update a pod image
 func SidecarSetHashWithoutImage(sidecarSet *appsv1alpha1.SidecarSet) (string, error) {
 	ss := sidecarSet.DeepCopy()
-	for i := range ss.Spec.InitContainers {
-		ss.Spec.InitContainers[i].Image = ""
-	}
 	for i := range ss.Spec.Containers {
 		ss.Spec.Containers[i].Image = ""
 	}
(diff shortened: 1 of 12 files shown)


Fei-Guo:
API LGTM

pkg/webhook/sidecarset/mutating/hash.go
```diff
@@ -40,6 +40,9 @@ func SidecarSetHash(sidecarSet *appsv1alpha1.SidecarSet) (string, error) {
 // SidecarSetHashWithoutImage calculates hash without sidecars's image
 func SidecarSetHashWithoutImage(sidecarSet *appsv1alpha1.SidecarSet) (string, error) {
 	ss := sidecarSet.DeepCopy()
+	for i := range ss.Spec.InitContainers {
+		ss.Spec.InitContainers[i].Image = ""
+	}
```
furykerry, line 45: init container cannot be updated inplace, so the  image of init container should not be munged

ryanzhang-oss: what is munged? We do need to include the initContainers in the hash to distinguish sidecarset that are only different in their initContainer settings, right?

furykerry: hash without image is used for the sidecarset controller to determine whether inplace update is possible once sidecarset changes. init container cannot be changed inplace, so the value must be included in the SidecarSetHashWithoutImage, and cannot be munged (modified to empty string)

ryanzhang-oss: yes, I get that point and the current PR also has that. However, on second thought, I don't know how can we simply skip the inplace update when both the container and other parts of the sidecarset are both changed.

Fei-Guo (commented):
LGTM

furykerry (commented):
LGTM

FillZpp (approved):
LGTM

PR #377, opened 2020-09-18
fix current revision

### LINKED ISSUE #376: [BUG] CloneSet not work correctly
**What happened**:
Create an clone with replicas 1 maxSurge 50% maxUnavailable 10%. After that we had one pod named: `sample-89hnf`.
 And do some update, for any reason which make the new revision pod (named: `sample-8qbjl`) stuck pending. Then rollback the clone. It will kill the pod `sample-89hnf`, and recreate an new pod. This will cause service unavailable.

Just kill the pod `sample-8qbjl`, and keep the pod  `sample-89hnf`.

**Environment**:
- Kruise version: 
- Kubernetes version (use `kubectl version`): 1.17.2
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

shiyan2016:

fix #376


FillZpp (approved):
LGTM. Thanks for fixing it!

PR #381, opened 2020-09-25
Update CRD and generated file to align with the UnitedDeployment API …

Fei-Guo:
Since UnitedDeployment now supports Cloneset, the CRD and generated deepcopy code need to be updated accordingly.


FillZpp (approved):
LGTM

PR #382, opened 2020-09-25
Fix broken UTs

Fei-Guo:

This change fixes to UTs.

1) The cloneset testscale UT will scale from 1 Pod to 5 Pods. The timeout of 3 seconds can be too small for the scaling to complete occasionally. Therefore, this change increases the timeout to 10seconds.

2) In uniteddeployment for cloneset UT,  the TestTemplateTypeSwtichToCS test does not specify the `replica` in CloneSetSpec, which is a required field for cloneset. This change fixes it.


FillZpp (approved):
LGTM

PR #395, opened 2020-10-10
delete useless incrementalAllocate function, and  return if GetAllocatedReplicas false

kadisi:

because uniteddeployment webhook has restricted the relationship with Replicas and Subset。

so effectiveReplicas() function always return true.

and incrementalAllocate() function is useless.

so if effectiveReplicas() return false, return this reconcile.

### DIFF
--- a/pkg/controller/uniteddeployment/allocator.go
+++ b/pkg/controller/uniteddeployment/allocator.go
@@ -57,7 +57,7 @@ func (n subsetInfos) Swap(i, j int) {
 // GetAllocatedReplicas returns a mapping from subset to next replicas.
 // Next replicas is allocated by replicasAllocator, which will consider the current replicas of each subset and
 // new replicas indicated from UnitedDeployment.Spec.Topology.Subsets.
-func GetAllocatedReplicas(nameToSubset *map[string]*Subset, ud *appsv1alpha1.UnitedDeployment) (*map[string]int32, bool, string) {
+func GetAllocatedReplicas(nameToSubset *map[string]*Subset, ud *appsv1alpha1.UnitedDeployment) (*map[string]int32, error) {
 	subsetInfos := getSubsetInfos(nameToSubset, ud)
 	specifiedReplicas := getSpecifiedSubsetReplicas(ud)
 
@@ -74,9 +74,9 @@ type replicasAllocator struct {
 	subsets *subsetInfos
 }
 
-func (s *replicasAllocator) effectiveReplicas(replicas int32, subsetReplicasLimits *map[string]int32) (bool, string) {
+func (s *replicasAllocator) validateReplicas(replicas int32, subsetReplicasLimits *map[string]int32) error {
 	if subsetReplicasLimits == nil {
-		return true, ""
+		return nil
 	}
 
 	var specifiedReplicas int32
@@ -85,7 +85,8 @@ func (s *replicasAllocator) effectiveReplicas(replicas int32, subsetReplicasLimi
 	}
 
 	if specifiedReplicas > replicas {
-		return false, fmt.Sprintf("Specified subsets' replica (%d) is greater than UnitedDeployment replica (%d)", specifiedReplicas, replicas)
+		return fmt.Errorf("Specified subsets' replica (%d) is greater than UnitedDeployment replica (%d)",
+			specifiedReplicas, replicas)
 	} else if specifiedReplicas < replicas {
 		specifiedCount := 0
 		for _, subset := range *s.subsets {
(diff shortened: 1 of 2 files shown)


pkg/controller/uniteddeployment/allocator.go
```diff
@@ -143,7 +143,7 @@ func getSubsetInfos(nameToSubset *map[string]*Subset, ud *appsv1alpha1.UnitedDep
 // in order to make the scaling smoothly
 func (s *replicasAllocator) AllocateReplicas(replicas int32, specifiedSubsetReplicas *map[string]int32) (*map[string]int32, bool, string) {
```
FillZpp, line 144: Just return error? `(*map[string]int32, error)`

pkg/controller/uniteddeployment/uniteddeployment_controller.go
```diff
@@ -188,14 +188,18 @@ func (r *ReconcileUnitedDeployment) Reconcile(request reconcile.Request) (reconc
...
 		klog.Errorf("Fail to get Subsets of UnitedDeployment %s/%s: %s", instance.Namespace, instance.Name, err)
-		r.recorder.Event(instance.DeepCopy(), corev1.EventTypeWarning, fmt.Sprintf("Failed%s", eventTypeFindSubsets), err.Error())
+		r.recorder.Event(instance.DeepCopy(), corev1.EventTypeWarning, fmt.Sprintf("Failed %s",
+			eventTypeFindSubsets), err.Error())
 		return reconcile.Result{}, nil
 	}
 
 	nextReplicas, effectiveSpecifiedReplicas, ineffectiveReason := GetAllocatedReplicas(nameToSubset, instance)
 	klog.V(4).Infof("Get UnitedDeployment %s/%s next replicas %v", instance.Namespace, instance.Name, nextReplicas)
 	if !effectiveSpecifiedReplicas {
```
FillZpp, line 198:
```golang
nextReplicas, err := GetAllocatedReplicas(nameToSubset, instance)
if err != nil {
    // ...
}
```

pkg/controller/uniteddeployment/uniteddeployment_controller.go
```diff
@@ -188,14 +188,18 @@ func (r *ReconcileUnitedDeployment) Reconcile(request reconcile.Request) (reconc
...
-		r.recorder.Eventf(instance.DeepCopy(), corev1.EventTypeWarning, fmt.Sprintf("Failed%s", eventTypeSpecifySubbsetReplicas), "Specified subset replicas is ineffective: %s", ineffectiveReason)
+		klog.Errorf("Specified subset replicas is ineffective: %s", ineffectiveReason)
```
FillZpp, line 199: Better to add the ud namespace/name in the error log

pkg/controller/uniteddeployment/allocator.go
kadisi: done

pkg/controller/uniteddeployment/uniteddeployment_controller.go
kadisi: done

pkg/controller/uniteddeployment/uniteddeployment_controller.go
kadisi: ok

pkg/controller/uniteddeployment/allocator.go
```diff
@@ -141,12 +141,13 @@ func getSubsetInfos(nameToSubset *map[string]*Subset, ud *appsv1alpha1.UnitedDep
 // If effective, normalAllocate will be called. It will apply these specified replicas, then average the rest replicas to left unspecified subsets.
 // If not, it will incrementally allocate all of the replicas. The current replicas spread situation will be considered,
 // in order to make the scaling smoothly
-func (s *replicasAllocator) AllocateReplicas(replicas int32, specifiedSubsetReplicas *map[string]int32) (*map[string]int32, bool, string) {
+func (s *replicasAllocator) AllocateReplicas(replicas int32, specifiedSubsetReplicas *map[string]int32) (
+	*map[string]int32, error) {
 	if effective, reason := s.effectiveReplicas(replicas, specifiedSubsetReplicas); !effective {
```
FillZpp, line 146: How about rename `effectiveReplicas` to `validateReplicas`? And it could return error instead of bool+string.

kadisi: good idea , and i will also change the function comment

FillZpp (approved):
LGTM

PR #397, opened 2020-10-12
Optimize 'ApplyRevision' logic in revisionControl

dylandee:
it seems not reasonable to cause panic in deployment sync logic and its useless to transform bytes to string and back to bytes


FillZpp (approved):
LGTM

PR #398, opened 2020-10-12
`str` is []byte, it is redundant to convert it to []byte

hantmac:

`str` is []byte, it is redundant to convert it to []byte


FillZpp (approved):
LGTM

PR #405, opened 2020-10-15
clean ups

hantmac:

Some clean ups.

NONE

In [uber-go guide](https://github.com/uber-go/guide/blob/master/style.md#nil-is-a-valid-slice), the zero value (a slice declared with var) is usable immediately without make(), so follow the recommend way.

### DIFF
--- a/pkg/controller/daemonset/daemonset_controller.go
+++ b/pkg/controller/daemonset/daemonset_controller.go
@@ -679,7 +679,9 @@ func (dsc *ReconcileDaemonSet) syncNodes(ds *appsv1alpha1.DaemonSet, podsToDelet
 		deleteDiff = burstReplicas
 	}
 
-	expectations.SetExpectations(dsKey, createDiff, deleteDiff)
+	if err := expectations.SetExpectations(dsKey, createDiff, deleteDiff); err != nil {
+		utilruntime.HandleError(err)
+	}
 	// error channel to communicate back failures.  make the buffer big enough to avoid any blocking
 	errCh := make(chan error, createDiff+deleteDiff)
 
@@ -786,7 +788,7 @@ func (dsc *ReconcileDaemonSet) syncNodes(ds *appsv1alpha1.DaemonSet, podsToDelet
 	deleteWait.Wait()
 
 	// collect errors if any for proper reporting/retry logic in the controller
-	errors := []error{}
+	var errors []error
 	close(errCh)
 	for err := range errCh {
 		errors = append(errors, err)
--- a/pkg/controller/daemonset/inplace_update.go
+++ b/pkg/controller/daemonset/inplace_update.go
@@ -143,7 +143,7 @@ func (dsc *ReconcileDaemonSet) syncNodesWhenInplaceUpdate(ds *appsv1alpha1.Daemo
 	updateWait.Wait()
 
 	// collect errors if any for proper reporting/retry logic in the controller
-	errors := []error{}
+	var errors []error
 	close(errCh)
 	for err := range errCh {
 		errors = append(errors, err)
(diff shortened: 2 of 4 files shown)


pkg/controller/daemonset/daemonset_controller.go
```diff
@@ -679,7 +679,9 @@ func (dsc *ReconcileDaemonSet) syncNodes(ds *appsv1alpha1.DaemonSet, podsToDelet
 		deleteDiff = burstReplicas
 	}
 
-	expectations.SetExpectations(dsKey, createDiff, deleteDiff)
+	if err := expectations.SetExpectations(dsKey, createDiff, deleteDiff); err != nil {
```
Fei-Guo, line 682: You may not need to explicitly check the error here because it is unlikely to hit an error. The check in line 667 already prevents error from happening.

hantmac: > You may not need to explicitly check the error here because it is unlikely to hit an error. The check in line 667 already prevents error from happening.

Got your point, but I'm not check the error of  `SetExpectations(dsKey, createDiff, deleteDiff)`, in line 667 already prevent error from happening.  The line https://github.com/openkruise/kruise/blob/da068736e26f8cf6f1fb99f7599f083f81cb0506/vendor/k8s.io/kubernetes/pkg/controller/controller_utils.go#L156 embed `store` interface, so when call func `Add` we should check the error of it because line https://github.com/openkruise/kruise/blob/da068736e26f8cf6f1fb99f7599f083f81cb0506/vendor/k8s.io/client-go/tools/cache/controller.go#L367 and many other code also check the error of `Add`. And the ` expectations.SetExpectations` call `Add` func.

Fei-Guo: func (c *cache) Add(obj interface{}) error {
	key, err := c.keyFunc(obj)
	if err != nil {
		return KeyError{obj, err}
	}
	c.cacheStorage.Add(key, obj)
	return nil
}

Yes. I checked the Add func and found the only chance when Add would return an error is failing the call to c.keyFunc() and I believe line 667 does the exact same check hence I think the error check you add can be redundant. 

The example you gave does not call c.keyFunc() forehand hence the situation is slightly different.

hantmac: The `keyFunc` in https://github.com/openkruise/kruise/blob/da068736e26f8cf6f1fb99f7599f083f81cb0506/pkg/controller/daemonset/daemonset_controller.go#L667 is different from `keyFunc` in https://github.com/openkruise/kruise/blob/da068736e26f8cf6f1fb99f7599f083f81cb0506/vendor/k8s.io/kubernetes/pkg/controller/controller_utils.go#L131  .

The first `keyFunc` is `KeyFunc = cache.DeletionHandlingMetaNamespaceKeyFunc` and it's function is to parse a object in k8s and the second  `keyFunc` is to parse the struct type `ControllerExpectations` . https://github.com/openkruise/kruise/blob/da068736e26f8cf6f1fb99f7599f083f81cb0506/vendor/k8s.io/kubernetes/pkg/controller/controller_utils.go#L291

```
func (r *ControllerExpectations) SetExpectations(controllerKey string, add, del int) error {
	exp := &ControlleeExpectations{add: int64(add), del: int64(del), key: controllerKey, timestamp: clock.RealClock{}.Now()}
	klog.V(4).Infof("Setting expectations %#v", exp)
	return r.Add(exp)
}
```

SO we can't use first `keyFunc` to check the second `keyFunc` error,  as your said, the situation of two `keyFunc` is different.

Fei-Guo: Got you! Thanks for the explanation.

Fei-Guo (approved):

Issue #406, opened 2020-10-15
[BUG] Dockerfile should pin specific ubuntu version


lowkeyliesmyth:
**What happened**: 

Get released kruise-manager image: `docker pull openkruise/kruise-manager:v0.6.1`
Check underlying OS release:
```
docker run --rm  --entrypoint "/bin/cat"  openkruise/kruise-manager:v0.6.1 /etc/os-release
NAME="Ubuntu"
VERSION="18.04.4 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.4 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic
```

**What you expected to happen**: 
Latest ubuntu LTS release 20.04 would be used since the Dockerfile references `FROM ubuntu:latest`. Instead, version 18.04.4 is used.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kruise version:
- Kubernetes version (use `kubectl version`):
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

FillZpp:
@lowkeyliesmyth emm... this might be caused by the cached image on my vm. I used `--no-cache` option for `docker build` each time, but it seems no effect on the base image.

PR #407, opened 2020-10-15
Use latest ubuntu LTS image

lowkeyliesmyth:
To follow Dockerfile best practices, pin the manager Dockerfile to explicitly use the latest ubuntu 20.04 LTS image. The current `openkruise/kruise-manager:v0.6.1` image is built on ubuntu 18.04 even though the Dockerfile references `ubuntu:latest`.

#406 

Build the docker image
`docker build . -t openkruise/kruise-manager:test`

Verify the image's output contains `VERSION_ID=20.04`:
`docker run --rm  --entrypoint "/bin/cat"  openkruise/kruise-manager:test /etc/os-release`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -mod=vendor -a
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 #FROM gcr.io/distroless/static:nonroot
-FROM ubuntu:latest
+FROM ubuntu:20.04
 WORKDIR /
 COPY --from=builder /workspace/manager .
 ENTRYPOINT ["/manager"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -mod=vendor -a
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 #FROM gcr.io/distroless/static:nonroot
-FROM ubuntu:20.04
+FROM ubuntu:latest
 WORKDIR /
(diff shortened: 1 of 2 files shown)


Fei-Guo (commented):
I doubt we need to hardcode the version in docker file. 

According to the dockerhub description

`The ubuntu:latest tag points to the "latest LTS", since that's the version recommended for general use. The ubuntu:rolling tag points to the latest release (regardless of LTS status).`

We should always follow what presented in docker hub for the image with `latest` tag. 
Of course, you can customize your base image as you want.

lowkeyliesmyth:
The problem with relying on the `latest` tag is that, especially during the build process, if an older image has already been populated in the builder's local cache then that cached image is used instead of the actual newest `latest` image. 

Which I believe is how the `openkruise/kruise-manager:v0.6.1` image is currently based on ubuntu 18.04 instead of 20.04 even though it references the `ubuntu:latest` tag (as referenced in #406 ).

Fei-Guo:
> The problem with relying on the `latest` tag is that, especially during the build process, if an older image has already been populated in the builder's local cache then that cached image is used instead of the actual newest `latest` image.
> 
> Which I believe is how the `openkruise/kruise-manager:v0.6.1` image is currently based on ubuntu 18.04 instead of 20.04 even though it references the `ubuntu:latest` tag (as referenced in #406 ).

I understand your concern. To me, the right fix is to regenerate the 0.6.1 image with the exact latest base os image.  @FillZpp what do you think?

FillZpp:
@Fei-Guo Sure, I'm going to rebuild the image. I will add `--pull` option this time during `docker build` which indicates `Always attempt to pull a newer version of the image`.

The whole command is `docker build --pull --no-cache . -t openkruise/kruise-manager:v0.6.1`.

@lowkeyliesmyth Thanks for reporting. You could help Kruise optimize the `docker build` command in Makefile if you wish.

lowkeyliesmyth:
Updating the makefile probably makes more sense than the change I originally proposed. I added the Makefile docker-build commit to this PR, but if you would rather I add those to a new PR and close this one and create a new PR.

Fei-Guo (approved):

PR #408, opened 2020-10-21
delete useless ObjectMeta field from subset

kadisi:
Signed-off-by: kadisi <iamkadisi@163.com>

### DIFF
--- a/pkg/controller/uniteddeployment/subset.go
+++ b/pkg/controller/uniteddeployment/subset.go
@@ -24,15 +24,15 @@ import (
 
 // Subset stores the details of a subset resource owned by one UnitedDeployment.
 type Subset struct {
-	metav1.ObjectMeta
-
-	Spec   SubsetSpec
-	Status SubsetStatus
+	Name      string
+	RefName   string
+	Namespace string
+	Spec      SubsetSpec
+	Status    SubsetStatus
 }
 
 // SubsetSpec stores the spec details of the Subset
 type SubsetSpec struct {
-	SubsetName     string
 	Replicas       int32
 	UpdateStrategy SubsetUpdateStrategy
 	SubsetRef      ResourceRef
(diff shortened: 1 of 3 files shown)


pkg/controller/uniteddeployment/subset.go
```diff
@@ -24,15 +24,15 @@ import (
 
 // Subset stores the details of a subset resource owned by one UnitedDeployment.
 type Subset struct {
```
hantmac, line 26: In my opinion, using `metav1.ObjectMeta` is better than using specific param of ObjectMeta.

kadisi: `metav1.ObjectMeta` is only  used for `func (m *SubsetControl) objectKey() function `,  and too many fields need to be assigned to it in order to use it，
I think this one adds complexity and Subset is also not a real K8s object

hantmac: It may be that the developer considers scalability more important

Fei-Guo: A compromise is that we can leave most of the fields in subset.ObjectMeta empty/unassigned such as label, annotation, finalizier etc to reduce the cost of initialization. We can add initialization code for those fields later when we need them in new features. Of course, add a comment to explain the intention.   

@wu8685 @kadisi what do you think?

wu8685: Another choice is to use [GetObjectMeta()](https://github.com/openkruise/kruise/blob/master/pkg/controller/uniteddeployment/adapter/advanced_statefulset_adapter.go#L51) in each adapter, if we just want to get rid of the initialization.

Fei-Guo:
It seems that the cleanup does not get a consensus so let us hold it. Please reopen if people changed their minds.

Issue #409, opened 2020-10-22
[BUG] Container Spec with securityContext as privileged is not working


rishi-anand:
**What happened**:
Container Spec with securityContext as privileged is not working. I have tried BroadcastJob and it was failing with below error msg.
```
rishi@Spectro$ k apply -f /Users/rishi/work/src/spectro-node-agent/broadcast-job-manifest.yaml
secret/spectro-common-dev-image-pull-secret unchanged
Error from server: error when creating "/Users/rishi/work/src/spectro-node-agent/broadcast-job-manifest.yaml": admission webhook "vbroadcastjob.kb.io" denied the request: spec.template.spec.containers[0].securityContext.privileged: Forbidden: disallowed by cluster policy
```
However, it works with native Deployment, DaemonSet.

**What you expected to happen**:
I should be able to create BroadcastJob with securityContext as privileged.

**How to reproduce it (as minimally and precisely as possible)**:
Apply below manifest to reproduce it.
```
apiVersion: apps.kruise.io/v1alpha1
kind: BroadcastJob
metadata:
  name: broadcastjob-active-deadline
spec:
  template:
    spec:
      containers:
        - name: sleep
          image: busybox
          command: ["sleep",  "50000"]
          securityContext:
            privileged: true
      restartPolicy: Never
  completionPolicy:
    type: Always
    activeDeadlineSeconds: 10
```

**Anything else we need to know?**:
I was debugging the issue and somehow capabilities doesn't seem to be initialized properly.
https://github.com/openkruise/kruise/blob/cde3bc321b1769c770d52f793a235336ea4f549b/vendor/k8s.io/kubernetes/pkg/apis/core/validation/validation.go#L5477-L5481

https://github.com/openkruise/kruise/blob/master/vendor/k8s.io/kubernetes/pkg/capabilities/capabilities.go#L83-L85

**Environment**:
- Kruise version: 0.6.0
- Kubernetes version (use `kubectl version`): v1.18.4
- OS (e.g: `cat /etc/os-release`): 
- Kernel (e.g. `uname -a`): Darwin Spectro.local 19.6.0 Darwin Kernel Version 19.6.0: Thu Jun 18 20:49:00 PDT 2020; root:xnu-6153.141.1~1/RELEASE_X86_64 x86_64
- Install tools:
- Others:

rishi-anand:
Tried to add below securityContext in deployment for kruise-controller-manager. But that didn't helped.
```
    securityContext:
      privileged: true
```

[...]

PR #414, opened 2020-10-23
Initializing capabilities if allowPrivileged is true

rishi-anand:
It will initialize capabilities if allowed-privileged is true which can be added in helm-values. 
Same way as kubernetes does it by taking `allowed-privileged=true` as api-server parameter
https://github.com/kubernetes/kubernetes/blob/1f0d71cd0d430c69128647b1f7ca50e36474b523/cmd/kube-apiserver/app/server.go#L307-L316

#409

Tested it by running dev image of kruise-controller-manager

While starting kruise manager, we can specify allow-privileged=true and we can run below manifest to verify it.
```
kind: BroadcastJob
metadata:
  name: broadcastjob-active-deadline
spec:
  template:
    spec:
      containers:
        - name: sleep
          image: busybox
          command: ["sleep",  "50000"]
          securityContext:
            privileged: true
      restartPolicy: Never
  completionPolicy:
    type: Always
    activeDeadlineSeconds: 10
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,6 +22,8 @@ import (
 	_ "net/http/pprof"
 	"os"
 
+	"k8s.io/kubernetes/pkg/capabilities"
+
 	extclient "github.com/openkruise/kruise/pkg/client"
 	"github.com/openkruise/kruise/pkg/util/fieldindex"
 	"github.com/openkruise/kruise/pkg/webhook"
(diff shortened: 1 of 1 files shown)


Fei-Guo (approved):

Fei-Guo:
LG. Thanks for working on this!

PR #417, opened 2020-10-26
release ownerRef when pod's labels not macth cloneset's selector

tudou34:
If I delete the lable of cloneset's pod,then the label not match the selector of cloneset, it will not delete the pod's ownerRef and create a new pod.
But for the behavior of rs,it will delete ownerRef and create new pod.
so this pr will fix this.

### DIFF
--- a/pkg/controller/cloneset/cloneset_controller.go
+++ b/pkg/controller/cloneset/cloneset_controller.go
@@ -33,6 +33,7 @@ import (
 	"github.com/openkruise/kruise/pkg/util/fieldindex"
 	"github.com/openkruise/kruise/pkg/util/gate"
 	historyutil "github.com/openkruise/kruise/pkg/util/history"
+	"github.com/openkruise/kruise/pkg/util/refmanager"
 	apps "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -218,6 +219,12 @@ func (r *ReconcileCloneSet) doReconcile(request reconcile.Request) (res reconcil
 
 	// list all active Pods and PVCs belongs to cs
 	filteredPods, filteredPVCs, err := r.getOwnedResource(instance)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	//release Pods ownerRef
+	filteredPods, err = r.claimPods(instance, filteredPods)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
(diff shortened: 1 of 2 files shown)


Fei-Guo (commented):
Thanks for working on this. 

My suggestion is to reuse RefManager package in Kruise. It includes the releasing Pod ownerreference functionality in `ClaimOwnedObjects` method.

Check the following example in UnitedDeployment. 
https://github.com/openkruise/kruise/blob/02b731faf39d8961f00d9c29a6cd187d240eee76/pkg/controller/uniteddeployment/adapter/cloneset_adapter.go#L174

tudou34:
> Thanks for working on this.
> 
> My suggestion is to reuse RefManager package in Kruise. It includes the releasing Pod ownerreference functionality in `ClaimOwnedObjects` method.
> 
> Check the following example in UnitedDeployment.
> https://github.com/openkruise/kruise/blob/02b731faf39d8961f00d9c29a6cd187d240eee76/pkg/controller/uniteddeployment/adapter/cloneset_adapter.go#L174

tudou34:
Already update code to reuse RefManager. @Fei-Guo

pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -498,3 +505,27 @@ func (r *ReconcileCloneSet) truncateHistory(
 	}
 	return nil
 }
+
+func (r *ReconcileCloneSet) claimPods(instance *appsv1alpha1.CloneSet, pods []*v1.Pod) ([]*v1.Pod, error) {
```
FillZpp, line 509: Could you add some tests for `claimPods`?

pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -498,3 +505,27 @@ func (r *ReconcileCloneSet) truncateHistory(
...
+	manager, err := refmanager.New(r, instance.Spec.Selector, instance, r.scheme)
+	if err != nil {
+		return nil, err
+	}
+
+	selected := make([]metav1.Object, len(pods))
+	for i, pod := range pods {
+		selected[i] = pod.DeepCopy()
```
FillZpp, line 517: There is no need to deep copy. The Pods list from controller-runtime have already been deep copied.
https://github.com/openkruise/kruise/blob/master/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/cache_reader.go#L138

pkg/controller/cloneset/cloneset_controller.go
tudou34: done.

pkg/controller/cloneset/cloneset_controller_test.go
```diff
@@ -131,6 +137,109 @@ func TestReconcile(t *testing.T) {
 	testUpdate(g, instance)
 }
 
+func TestClaimPods(t *testing.T) {
```
Fei-Guo, line 140: Can you simply test the `claimPods` method functionality without setup the TestManager? The test can be as simple as just passing a cloneset object and an array of Pods. You don't need a live apiserver to create real objects.

Fei-Guo (commented):
Change to the controller LG.  One comment for the test.

pkg/controller/cloneset/cloneset_controller_test.go
tudou34: use fakeClient instead, done.

Fei-Guo: I  am actually suggesting to remove entire gomega code and client code, even removing the input parameter t. 
Will that work?

tudou34: I think client code can't be moved, it will be used in the interface of k8s update.
can you show me some example?

Fei-Guo: line 191, Will change `claimed, err := reconciler.claimPods(instance, test.pods)` to `claimed, err := claimPods(instance, test.pods)` work?

Fei-Guo: Yes. you are right.  I overlook that aspect. sorry.

Fei-Guo (approved):

PR #421, opened 2020-10-29
Fix remove unused go mod

pigletfly:

remove unused go mod


FillZpp:
@pigletfly `k8s.io/code-generator` is used for generating `pkg/client` in this script https://github.com/openkruise/kruise/blob/master/scripts/generate_client.sh

Issue #215, opened 2020-01-16
[feature request] 类似于 job + cronjob + daemonset 这样的 CRD 控制器


kong62:
**What would you like to be added**:
实现过类似于 job + cronjob + daemonset 这样的 CRD 控制器，功能需求大致是这样的：
1. 添加新节点的时候，自动在新节点上启动一次 job，完成后退出                      【不满足】
2. 当有修改节点的任务 job 分配的时候，启动一次 job，完成后退出                  【满足】
3. 每天定时在所有节点上启动一次 job，完成后退出                                            【不满足】

**Why is this needed**:
代替 ansible 这样的工具做节点上的批量修改，速度比他们快。而且很少占用资源。另外主要就是针对新加入集群的节点自动处理，这一般人为去做比较慢。

Ghostbaby:
类似初始化系统，但是不是所有主机都加入k8s集群

kong62:
其实主要就是2个作用把：
1. 类似于创建云主机时候的初始化脚本，用于一次性初始化新创建的节点
2. 类似于代替 ansible 这样的工具，专门用于日常做一些任务，比如定时删除镜像，删除停止的容器，拷贝文件等等

FillZpp:
> 添加新节点的时候，自动在新节点上启动一次 job，完成后退出 【不满足】

`BroadcastJob` can do this with `spec.completionPolicy.type: "Never"`, it guarantees a completed job on all desired Nodes.
When new nodes are added, the pods will be triggered automatically on those new nodes.

> 当有修改节点的任务 job 分配的时候，启动一次 job，完成后退出 【满足】

`BroadcastJob` can do this.

> 每天定时在所有节点上启动一次 job，完成后退出 【不满足】

A cron `BroadcastJob` is in our plan.

rishi-anand:
@FillZpp, @Fei-Guo, If `BroadcastCronJob` is in the plan and if no one is working on it then I would like to implement it.

Fei-Guo:
> @FillZpp, @Fei-Guo, If `BroadcastCronJob` is in the plan and if no one is working on it then I would like to implement it.

Hi, Rishi, 

You are very welcome to contribute for this feature. From high level perspective, this can be done by either simply enhancing upstream CronJob to support BroadcastJob spec in addition to JobSpec or adding Cron capability directly in BroadcastJob.

[...]

PR #427, opened 2020-11-01
Generic AdvancedCronJob CRD & Controller proposal

rishi-anand:
Implement CRD and controller for `BroadcastCronJob`

#215

This is proposal. Will add test cases in implementation.

This is proposal. Will add more information in implementation PR.

Discussion: https://github.com/openkruise/kruise/issues/215#issuecomment-715506813
Implementation PR: https://github.com/spectrocloud/kruise/pull/1

@Fei-Guo, @FillZpp Please review


Fei-Guo (commented):
Would it make sense to extend upstream CronJob to support BroadcastJob? 

We can call it AdvancedCronJob which supports both upstream Job and BroadcastJob. 
I think when you implement the controller, most of the code will be duplicated with upstream CronJob anyway.

rishi-anand:
@Fei-Guo, CronJob native controller is already there which supports Job. We need a cron controller for just  BroadcastJob. 
Thus, I have created a CRD BroadcastCronJob which will be upstream cron controller for BroadcastJob.

I have implemented and tested the controller. PR: https://github.com/spectrocloud/kruise/pull/1

Fei-Guo:
What I meant is to have a controller to support both native job and broadcastjob so that people can use one controller for all cron type workloads. It is also possible we will have other types of Jobs in the futures so that the new cron controller can be extended to support new job types as well.

rishi-anand:
@Fei-Guo, Few parameters may be different for Job, BroadcatJob, or future CRD. From an implementation perspective, Should we have multiple template object in the spec of AdvancedCronJob or should it have reference to other CRD?

Fei-Guo:
> @Fei-Guo, Few parameters may be different for Job, BroadcatJob, or future CRD. From an implementation perspective, Should we have multiple template object in the spec of AdvancedCronJob or should it have reference to other CRD?

I would suggest to have multiple templates, which is more clearer. Of course, only one of them takes effect is multiple templates are specified and the spec should not be mutated once it is created.

rishi-anand:
Sure @Fei-Guo. I will make respective changes in the proposal docs.

rishi-anand:
@Fei-Guo, Modified proposal to have below points and other related changes.

```
## Proposal

- Adding a new CRD and controller for AdvancedCronJob.
- AdvancedCronJob should be able to reconcile Job/BroadcastJob or any other future CRD if required.
- Once AdvancedCronJob is created, spec cannot be modified.
- Adding webhook for validation of AdvancedCronJob.
```

Fei-Guo:
@rishi-anand Can you rebase your PR please?

rishi-anand:
@Fei-Guo, Somehow this PR got bit messed up. I have created another PR for proposal.
PR: https://github.com/openkruise/kruise/pull/437

PR #430, opened 2020-11-03
support secret cert for webhook with vip

plpan:

support secret cert for webhook controller with vip specified.

### DIFF
--- a/pkg/webhook/util/controller/webhook_controller.go
+++ b/pkg/webhook/util/controller/webhook_controller.go
@@ -231,7 +231,7 @@ func (c *Controller) sync() error {
 	var certWriter writer.CertWriter
 	var err error
 
-	if dnsName = webhookutil.GetHost(); len(dnsName) > 0 {
+	if dnsName = webhookutil.GetHost(); len(dnsName) > 0 && len(webhookutil.GetSecretName()) == 0 {
 		certWriter, err = writer.NewFSCertWriter(writer.FSCertWriterOptions{
 			Path: webhookutil.GetCertDir(),
 		})


pkg/webhook/util/controller/webhook_controller.go
```diff
@@ -231,7 +231,7 @@ func (c *Controller) sync() error {
 	var certWriter writer.CertWriter
 	var err error
 
-	if dnsName = webhookutil.GetHost(); len(dnsName) > 0 {
+	if dnsName = webhookutil.GetHost(); len(dnsName) > 0 && len(webhookutil.GetSecretName()) == 0 {
```
FillZpp, line 234: The `GetSecretName()` has a default value `kruise-webhook-service`, so this modification is not forward compatible.
Also, I don't understand why should this check be added?

plpan: You are right. I'm using it incorrectly. Please reject this pr.

plpan: @FillZpp When I test fsCertWriter in multiple pods with webhook url configured as the vip of pods ip. Every pod is trying to overwrite webhook caBundle with it's own local cert infinitely. 
Have you met this problem?

FillZpp: Oh, got you. How about define a new env in util.go like `WEBHOOK_CERT_WRITER`. If env is set to `fs` or `secret`, this specified writer should be used.

plpan: It sounds good. Maybe we should set dnsName and certWriter individually, so we can achieve the following four combinations:
- webhost with fsCertWriter
- webhost with secretCertWriter
- service with fsCertWriter
- service with secretCertWriter
But it's hard to be forward compatible.

FillZpp: Yes, it is a good idea to separate dnsName and certWriter.

For dnsName, it is easy to set:
```golang
if dnsName = webhookutil.GetHost(); len(dnsName) == 0 {
    dnsName = generator.ServiceToCommonName(webhookutil.GetNamespace(), webhookutil.GetServiceName())
}
```

But for certWriter, it may be more complicated to determine:
```golang
switch webhookutil.GetCertWriter() {
case fs:
    // ...
case secret:
    // ...
default:
    if len(webhookutil.GetHost()) > 0 {
        // fs
    } else {
        // secret
    }
}
```

PR #432, opened 2020-11-03
take ownership of controllerRevisions when listing for CloneSet

dylandee:
…r CloneSet

use ControllerRefManager to reconcile ControllerRef by adopting and orphaning for controllerRevisions

like what it done in DaemonSet and StatefulSet.

### DIFF
--- a/pkg/controller/cloneset/cloneset_controller.go
+++ b/pkg/controller/cloneset/cloneset_controller.go
@@ -20,6 +20,7 @@ package cloneset
 import (
 	"context"
 	"flag"
+	"fmt"
 	"time"
 
 	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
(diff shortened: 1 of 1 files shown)


FillZpp:
@dylandee  Why do you need to take ownership of ControllerRevisions not belong to this CloneSet?

dylandee:
> @dylandee Why do you need to take ownership of ControllerRevisions not belong to this CloneSet?

just like what is done in DaemonSet for controllerRevisions
https://github.com/kubernetes/kubernetes/blob/cd99c63570eb1489dd631c12ea86db708dbdcd59/pkg/controller/daemon/update.go#L242

or in Deployment for ReplicaSets
https://github.com/kubernetes/kubernetes/blob/cd99c63570eb1489dd631c12ea86db708dbdcd59/pkg/controller/deployment/deployment_controller.go#L503

if there's orphan controllerResions which match clonset selector,  adopt ownerRef to them
if there's controllerResions with ownerRef of released ownerRef, relase their ownerRef


some ci problems may remain in this pr, if you agree with this modification, i will fix them..

FillZpp:
@dylandee Yes, I know these workloads have ControllerRevision adoption, and you can also find it in Advanced StatefulSet https://github.com/openkruise/kruise/blob/master/pkg/controller/statefulset/statefulset_controller.go#L254

But this is only in order to be compatible with the original StatefulSet. I don't think it is a good idea to take ownership of orphan ControllerRevisions. This might lead to unpredictable results.

dylandee:
> @dylandee Yes, I know these workloads have ControllerRevision adoption, and you can also find it in Advanced StatefulSet https://github.com/openkruise/kruise/blob/master/pkg/controller/statefulset/statefulset_controller.go#L254
> 
> But this is only in order to be compatible with the original StatefulSet. I don't think it is a good idea to take ownership of orphan ControllerRevisions. This might lead to unpredictable results.

ok, understood.

PR #434, opened 2020-11-04
Optimize startup and cleanup for v1beta1 apis

FillZpp:

Optimize startup and cleanup for v1beta1 apis

### DIFF
--- a/charts/kruise/v0.7.0/templates/pre_delete_job.yaml
+++ b/charts/kruise/v0.7.0/templates/pre_delete_job.yaml
@@ -1,0 +1,31 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "{{ .Release.Name }}"
+  namespace: kruise-system
+  labels:
+    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
+    app.kubernetes.io/version: {{ .Chart.AppVersion }}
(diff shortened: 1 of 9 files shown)


Fei-Guo (approved):
Add some background: the webhook should be always enabled regardless whether user choose to enable a controller or not. Otherwise, if the CRD is installed and a CR is created, the default value will not be set by the webhook. If the use enables the controller later, the controller may panic.

rishi-anand:
@FillZpp, @Fei-Guo, 
I am constantly getting `informer has already started` while adding an indexer in the controller after this PR. If I remove these changes then I am able to add an index in the controller.

```
if err := mgr.GetFieldIndexer().IndexField(&appsv1alpha1.BroadcastJob{}, jobOwnerKey, func(rawObj runtime.Object) []string {
		// grab the job object, extract the owner...
		job := rawObj.(*appsv1alpha1.BroadcastJob)
		owner := metav1.GetControllerOf(job)
		if owner == nil {
			return nil
		}

		// ...make sure it's a CronJob...
		if owner.APIVersion != apiGVStr || owner.Kind != appsv1alpha1.AdvancedCronJobKind {
			return nil
		}

		// ...and if so, return it
		return []string{owner.Name}
	}); err != nil {
		return err
	}

```

My use-case is similar to PR: https://github.com/kubernetes-sigs/kubebuilder/issues/1422 where I am adding same index and while listing child jobs I am using that index.

FillZpp:
@rishi-anand My suggestion is that put your index register in `pkg/util/fieldindex/register.go` instead of in controllers.

PR #435, opened 2020-11-04
optimize the way to calculate parallelism in broadcastjob

dylandee:
optimize the way to calculate parallelism in broadcastjob

consistent with the way other crds  parsing  `intstr.IntOrString` in kruise


Fei-Guo (approved):
LGTM. Thanks for cleaning up this.

PR #439, opened 2020-11-06
add pod label validation in broadcastjob

dylandee:
since "job-name" and "controller-uid"  are selector for bj to find related pods,these two labels should not allow to preset, otherwise something may exceed user expectations.

### DIFF
--- a/pkg/controller/broadcastjob/broadcastjob_controller.go
+++ b/pkg/controller/broadcastjob/broadcastjob_controller.go
@@ -53,6 +53,11 @@ func init() {
 	flag.IntVar(&concurrentReconciles, "broadcastjob-workers", concurrentReconciles, "Max concurrent workers for BroadCastJob controller.")
 }
 
+const (
+	JobNameLabelKey       = "broadcastjob-name"
+	ControllerUIDLabelKey = "broadcastjob-controller-uid"
+)
+
 var (
 	concurrentReconciles = 3
 	controllerKind       = appsv1alpha1.SchemeGroupVersion.WithKind("BroadcastJob")
@@ -546,8 +551,8 @@ func (r *ReconcileBroadcastJob) getNodeToPodMap(pods []*corev1.Pod, job *appsv1a
 
 func labelsAsMap(job *appsv1alpha1.BroadcastJob) map[string]string {
 	return map[string]string{
-		"job-name":       job.Name,
-		"controller-uid": string(job.UID),
+		JobNameLabelKey:       job.Name,
+		ControllerUIDLabelKey: string(job.UID),
 	}
 }
 
--- a/pkg/webhook/broadcastjob/validating/broadcastjob_create_update_handler.go
+++ b/pkg/webhook/broadcastjob/validating/broadcastjob_create_update_handler.go
@@ -23,6 +23,7 @@ import (
 	"regexp"
 
 	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
+	"github.com/openkruise/kruise/pkg/controller/broadcastjob"
 	v1 "k8s.io/api/core/v1"
 	genericvalidation "k8s.io/apimachinery/pkg/api/validation"
 	validationutil "k8s.io/apimachinery/pkg/util/validation"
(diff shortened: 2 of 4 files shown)


dylandee:
Whats more, i am confused why `selecor` is not used in bj struct?

dylandee:
just like logic in job
https://github.com/kubernetes/kubernetes/blob/bd95fb101dd83cfca9d95e55b2f0d29720016e51/pkg/controller/job/job_controller.go#L698

i think its better to kill extra ative pods immediately in order to avoid waste of node resources

dylandee:
add `scripts/tools.go` to force `go mod` to see `k8s.io/code-generator` as dependency 
if not it will be wiped out when typing `go mod tidy`

pkg/controller/broadcastjob/broadcastjob_controller.go
```diff
@@ -369,10 +385,40 @@ func (r *ReconcileBroadcastJob) reconcilePods(job *appsv1alpha1.BroadcastJob,
 	// The rest pods to run
 	rest := int32(len(restNodesToRunPod))
 	var errCh chan error
+	var activeLock sync.Mutex
 	if active > parallelism {
 		// exceed parallelism limit
 		r.recorder.Eventf(job, corev1.EventTypeWarning, "TooManyActivePods", "Number of active pods exceed parallelism limit")
-		//TODO should we remove the extra pods ? it may just finish by its own.
+		diff := active - parallelism
```
Fei-Guo, line 392: Let us not merge two changes into one. This change should only deal with the label validation per the PR title.  For this part of change, please file a separate PR.

pkg/webhook/broadcastjob/validating/broadcastjob_create_update_handler.go
```diff
@@ -94,14 +94,20 @@ func validateBroadcastJobSpec(spec *appsv1alpha1.BroadcastJobSpec, fldPath *fiel
...
-	allErrs = append(allErrs, corevalidation.ValidatePodTemplateSpec(coreTemplate, fldPath.Child("template"))...)
 	if spec.Template.Spec.RestartPolicy != v1.RestartPolicyOnFailure &&
 		spec.Template.Spec.RestartPolicy != v1.RestartPolicyNever {
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("template").Child("spec").Child("restartPolicy"),
 			spec.Template.Spec.RestartPolicy,
 			"pod restartPolicy can only be Never or OnFailure"))
 	}
-	return allErrs
+	if spec.Template.Labels != nil {
+		if spec.Template.Labels["job-name"] != "" || spec.Template.Labels["controller-uid"] != "" {
```
Fei-Guo, line 104: I know it is an existing problem. Can you please replace hardcoded "job-name" and "controller-uid" with constant definition to improve readability?

pkg/controller/broadcastjob/broadcastjob_controller.go
dylandee: sure，i will separate these changes into several PRs

pkg/webhook/broadcastjob/validating/broadcastjob_create_update_handler.go
dylandee: sure, of course

pkg/controller/broadcastjob/broadcastjob_controller.go
```diff
@@ -53,9 +53,17 @@ func init() {
...
+	jobNameLabelKey       = "job-name"
+	controllerUIDLabelKey = "controller-uid"
+)
+
 var (
 	concurrentReconciles = 3
 	controllerKind       = appsv1alpha1.SchemeGroupVersion.WithKind("BroadcastJob")
+
+	// AttachedPodLabelKeys gathers pod label keys for broadcastJob selector
+	AttachedPodLabelKeys = []string{jobNameLabelKey, controllerUIDLabelKey}
```
Fei-Guo, line 66: Since we only have two constants，the array may not needed. Without the array, the code in webhook will not be complicated significantly.

pkg/controller/broadcastjob/broadcastjob_controller.go
```diff
@@ -53,9 +53,17 @@ func init() {
 	flag.IntVar(&concurrentReconciles, "broadcastjob-workers", concurrentReconciles, "Max concurrent workers for BroadCastJob controller.")
 }
 
+const (
+	jobNameLabelKey       = "job-name"
```
Fei-Guo, line 57: How about 
JobNameLabelKey = "broadcastjob-name"
ControllerUIDLabelKey = "broadcastjob-controller-uid"

Fei-Guo (commented):
You should change all the places that use the hardcoded "job-name" and "controller-uid"\
such as 
https://github.com/openkruise/kruise/blob/4ad07453bbde132d64c8871c6442fd7dffb3867c/pkg/controller/broadcastjob/broadcastjob_controller.go#L549

pkg/controller/broadcastjob/broadcastjob_controller.go
dylandee: OK

pkg/controller/broadcastjob/broadcastjob_controller.go
dylandee: OK

Fei-Guo (approved):
LGTM. Thanks for addressing the comments.

PR #442, opened 2020-11-10
pass time duration instead of 'reconcile.Result' through functions in adv ds

dylandee:
`reconcile.Request` is passing through methods in adv ds just for delay time duration. However `reflect.DeepEqual` uses reflect package which is low efficiency.

So I think its better to just pass delay time directly.

### DIFF
--- a/pkg/controller/daemonset/daemonset_controller.go
+++ b/pkg/controller/daemonset/daemonset_controller.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"reflect"
 	"sort"
 	"sync"
 	"time"
@@ -367,9 +366,12 @@ func (dsc *ReconcileDaemonSet) syncDaemonSet(request reconcile.Request) (reconci
 		return dsc.updateDaemonSetStatus(ds, nodeList, hash, true)
 	}
 
-	result, err := dsc.manage(ds, hash)
-	if err != nil || !reflect.DeepEqual(result, reconcile.Result{}) {
-		return result, err
+	delay, err := dsc.manage(ds, hash)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if delay != 0 {
+		return reconcile.Result{RequeueAfter: delay}, nil
 	}
 
 	// Process rolling updates if we're ready.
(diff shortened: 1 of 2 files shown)


pkg/controller/daemonset/daemonset_controller.go
```diff
@@ -367,26 +367,27 @@ func (dsc *ReconcileDaemonSet) syncDaemonSet(request reconcile.Request) (reconci
...
 	}
 
-	result, err := dsc.manage(ds, hash)
-	if err != nil || !reflect.DeepEqual(result, reconcile.Result{}) {
-		return result, err
+	delay, err := dsc.manage(ds, hash)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if int(delay) != 0 {
```
Fei-Guo, line 374: Do you need to convert to int type?

pkg/controller/daemonset/daemonset_controller.go
```diff
@@ -367,26 +367,27 @@ func (dsc *ReconcileDaemonSet) syncDaemonSet(request reconcile.Request) (reconci
...
-				if !reflect.DeepEqual(result, reconcile.Result{}) {
-					klog.V(6).Infof("delay to requeue.: %v", result)
-					return result, err
-				}
+			delay, err = dsc.rollingUpdate(ds, hash)
+			if err != nil {
+				klog.Errorf("rollingUpdate failed: %v", err)
+				return reconcile.Result{}, err
+			}
+			if int(delay) != 0 {
```
Fei-Guo, line 388: Do you need int conversion?

pkg/controller/daemonset/daemonset_controller.go
```diff
@@ -601,13 +601,12 @@ func (dsc *ReconcileDaemonSet) manage(ds *appsv1alpha1.DaemonSet, hash string) (
...
 		}
-		re, nodesNeedingDaemonPodsOnNode, podsToDeleteOnNode, err := dsc.podsShouldBeOnNode(node, nodeToDaemonPods, ds, hash)
+		delay, nodesNeedingDaemonPodsOnNode, podsToDeleteOnNode, err := dsc.podsShouldBeOnNode(node, nodeToDaemonPods, ds, hash)
 		if err != nil {
 			continue
 		}
-
-		if !reflect.DeepEqual(re, reconcile.Result{}) {
-			return re, nil
+		if int(delay) != 0 {
```
Fei-Guo, line 608: same here.

pkg/controller/daemonset/daemonset_controller.go
```diff
@@ -805,7 +799,9 @@ func (dsc *ReconcileDaemonSet) podsShouldBeOnNode(
 	nodeToDaemonPods map[string][]*corev1.Pod,
 	ds *appsv1alpha1.DaemonSet,
 	hash string,
-) (result reconcile.Result, nodesNeedingDaemonPods, podsToDelete []string, err error) {
+) (duration time.Duration, nodesNeedingDaemonPods, podsToDelete []string, err error) {
+
+	var requeueDuration = requeueduration.Duration{}
```
Fei-Guo, line 804: Why do you use requeueduration.Duration here? I don't see concurrent access issue in this method.

Fei-Guo (commented):
Thanks for working on this. I think this is a reasonable cleanup. Comments below.

pkg/controller/daemonset/daemonset_controller.go
dylandee: > Why do you use requeueduration.Duration here? I don't see concurrent access issue in this method.

I just find it convenient to update time duarion with requeueduration.Duration,  i will fix it.

pkg/controller/daemonset/daemonset_controller.go
dylandee: yeah, its redundant

Fei-Guo (approved):
LGTM.

Thanks for the clean up.

PR #444, opened 2020-11-11
fix pending pods created by daemonset controller

### LINKED ISSUE #443: [BUG] DaemonsetController didn't delete pod after drained node from cluster
**What happened**:

We meet a scenario that when we are draining some nodes from cluster and in the meantime daemonset controller creates some pods to these nodes, all created pods becomes pending since its nodeAffinity cannot be satisfied. 

image

Also, daemonset controller does not remove these pending pods in later reconcile loop and just prints some logs like these: 

```
I1111 15:06:31.797452     113 daemonset_controller.go:933] get node: i-xxx failed
I1111 15:06:31.821762     113 daemonset_controller.go:933] get node: i-xxx failed
I1111 15:06:31.822620     113 daemonset_controller.go:933] get node: i-xxx failed
I1111 15:06:31.842987     113 daemonset_controller.go:933] get node: i-xxx failed
I1111 15:06:31.860574     113 daemonset_controller.go:933] get node: i-xxx failed
I1111 15:06:31.878892     113 daemonset_controller.go:933] get node: i-xxx failed
```
 
Daemonset controller remove these pending pods in its reconcile loop like what original Kubernetes daemonset-controller does.

It's a little difficult to reproduce, try to create an advanced daemonset and at the same time use `kubectl drain node` to remove one node that this adv daemonset will create pod to.

**Environment**:
- Kruise version: 0.6.1
- Kubernetes version (use `kubectl version`): 1.18
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

jetmuffin:
Signed-off-by: jetmuffin <jeffchen328@gmail.com>

Do not filter out "nodes cannot be deployed" when we collect `nodeToDaemonPods` info,  all pending pods created on drained node will be considered as "should not continue running but exists" and delete in `manage` procedure.

fixes #443 

Reproduce situation describe in #443 and all pending pods with no node to run will be deleted automatically.

### DIFF
--- a/pkg/controller/daemonset/daemonset_controller.go
+++ b/pkg/controller/daemonset/daemonset_controller.go
@@ -923,10 +923,18 @@ func (dsc *ReconcileDaemonSet) getNodesToDaemonPods(ds *appsv1alpha1.DaemonSet) 
 		}
 		node := &corev1.Node{}
 		err = dsc.client.Get(context.TODO(), types.NamespacedName{Name: nodeName}, node)
-		if err == nil && CanNodeBeDeployed(node, ds) {
+		if err != nil {
+			klog.V(4).Infof("get node %s failed: %v", nodeName, err)
+			if errors.IsNotFound(err) {
+				// Unable to find the target node in the cluster, which means it has been drained from cluster,
+				// we still add this node to nodeToDaemonPods in order that the daemon pods that failed to be
+				// scheduled on it will be removed in later reconcile loop.
+				nodeToDaemonPods[nodeName] = append(nodeToDaemonPods[nodeName], pod)
+			}
+			continue
+		}
+		if CanNodeBeDeployed(node, ds) {
 			nodeToDaemonPods[nodeName] = append(nodeToDaemonPods[nodeName], pod)
-		} else {
-			klog.V(4).Infof("get node: %s failed", nodeName)
 		}
 	}
 


pkg/controller/daemonset/daemonset_controller.go
```diff
@@ -925,13 +925,7 @@ func (dsc *ReconcileDaemonSet) getNodesToDaemonPods(ds *appsv1alpha1.DaemonSet)
 				pod.Namespace, pod.Name, ds.Namespace, ds.Name)
 			continue
 		}
-		node := &corev1.Node{}
-		err = dsc.client.Get(context.TODO(), types.NamespacedName{Name: nodeName}, node)
-		if err == nil && CanNodeBeDeployed(node, ds) {
-			nodeToDaemonPods[nodeName] = append(nodeToDaemonPods[nodeName], pod)
-		} else {
-			klog.V(4).Infof("get node: %s failed", nodeName)
```
evertrain, line 933: should just work for "node not found" error, not all kind of errors

jetmuffin: update and add some comments on this.

pkg/controller/daemonset/daemonset_controller.go
```diff
@@ -923,10 +923,19 @@ func (dsc *ReconcileDaemonSet) getNodesToDaemonPods(ds *appsv1alpha1.DaemonSet)
...
+				// Unable to find the target node in the cluster, which means it has been drained from cluster,
+				// we still add this node to nodeToDaemonPods in order that the daemon pods that failed to be
+				// scheduled on it will be removed in later reconcile loop.
+				nodeToDaemonPods[nodeName] = append(nodeToDaemonPods[nodeName], pod)
+			} else {
+				klog.V(4).Infof("get node %s failed: %v", nodeName, err)
+				continue
+			}
+		}
+		if CanNodeBeDeployed(node, ds) {
```
Fei-Guo, line 937: If node is not found, the node object is empty. Will CanNodeBeDeployed return true in this case? If so, the Pods in the node are appended twice.

pkg/controller/daemonset/daemonset_controller.go
```diff
@@ -923,10 +923,19 @@ func (dsc *ReconcileDaemonSet) getNodesToDaemonPods(ds *appsv1alpha1.DaemonSet)
 		}
 		node := &corev1.Node{}
 		err = dsc.client.Get(context.TODO(), types.NamespacedName{Name: nodeName}, node)
-		if err == nil && CanNodeBeDeployed(node, ds) {
+		if err != nil {
+			if errors.IsNotFound(err) {
+				// Unable to find the target node in the cluster, which means it has been drained from cluster,
```
Fei-Guo, line 928: Correct me if I am wrong. If the node object is deleted from the apiserver, I think the default gc controller will remove the Pods that are associated to the deleted node. Is it true?

My understanding of the reported problem is that for the drained node, the "CanNodeBeDeployed" method returns false hence the Pods are not added to the map.

jetmuffin: > If the node object is deleted from the apiserver, I think the default gc controller will remove the Pods that are associated to the deleted node.

Right, If a pod has been scheduled to the node, it will be removed. However, if a pod has been created and added nodeAffinity by daemonset-controller but has not been scheduled to the node, it will stuck in pending state and will not be added to the map, also.

pkg/controller/daemonset/daemonset_controller.go
jetmuffin: update, PTAL

pkg/controller/daemonset/daemonset_controller.go
Fei-Guo: Now I understand the problem when looking at the util.GetTargetNodeName().

Fei-Guo (approved):
LGTM

PR #446, opened 2020-11-12
execute deletion if quantity of active pods is greater than parallelism in bj

dylandee:
just like logic in native job
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/job/job_controller.go#L698

In order to avoid the waste of node resource, i think redundant active pods ought to be deleted.

### DIFF
--- a/pkg/controller/broadcastjob/broadcastjob_controller.go
+++ b/pkg/controller/broadcastjob/broadcastjob_controller.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"sort"
 	"sync"
 	"time"
 
@@ -28,6 +29,7 @@ import (
 	"github.com/openkruise/kruise/pkg/util/gate"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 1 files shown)


Fei-Guo:
I vaguely remember that we had discussed this behavior when BroadcastJob was first released. The concern is that the job in BroadcastJob can be different from regular computation job. It most likely does maintenance operations like node daemon update/clean up, pre-loading image etc. If we kill the job during those operations, the node may enter weird intermediate state which we need to clean up with extra effort.  Based on the assumptions that 1) BroadcastJob mostly does maintenance work and 2) the job should not be long-running job, we decide not to kill the extra active Pods.

Let me know your thoughts about this. 
Thanks

dylandee:
> I vaguely remember that we had discussed this behavior when BroadcastJob was first released. The concern is that the job in BroadcastJob can be different from regular computation job. It most likely does maintenance operations like node daemon update/clean up, pre-loading image etc. If we kill the job during those operations, the node may enter weird intermediate state which we need to clean up with extra effort. Based on the assumptions that 1) BroadcastJob mostly does maintenance work and 2) the job should not be long-running job, we decide not to kill the extra active Pods.
> 
> Let me know your thoughts about this.
> Thanks

Yeah, you are right, the node may enter some kind of  intermediate state if we clean extra pods directly. But  I have some questions about this workload. I think BroadcastJob should be more like job on each node, and 'spec.Parallelism' means how many same jobs may be running in the same node. Actually, BroadcastJob is more like daemon and just do some clean up. In this case,  'spec.Parallelism' seems weird and dispensable.

I am going to refactory BroadcastJob into the first situation I described and I can contribute the changes if needed.

Fei-Guo:
The spec.Parallelism is still meaningful in the case where we need to control the global parallelism level. For example, in the case of pre-loading image, using spec.Parallelism can avoid a burst imaging pulling requests to the image server when the number of node is large.

dylandee:
> The spec.Parallelism is still meaningful in the case where we need to control the global parallelism level. For example, in the case of pre-loading image, using spec.Parallelism can avoid a burst imaging pulling requests to the image server when the number of node is large.

Cool, understood.

PR #448, opened 2020-11-16
add rate limiter config

shiyan2016:
add rate limiter config

### DIFF
--- a/pkg/controller/cloneset/cloneset_controller.go
+++ b/pkg/controller/cloneset/cloneset_controller.go
@@ -33,6 +33,7 @@ import (
 	"github.com/openkruise/kruise/pkg/util/fieldindex"
 	"github.com/openkruise/kruise/pkg/util/gate"
 	historyutil "github.com/openkruise/kruise/pkg/util/history"
+	"github.com/openkruise/kruise/pkg/util/ratelimiter"
 	"github.com/openkruise/kruise/pkg/util/refmanager"
 	apps "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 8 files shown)


pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -99,7 +101,9 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
 func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// Create a new controller
-	c, err := controller.New("cloneset-controller", mgr, controller.Options{Reconciler: r, MaxConcurrentReconciles: concurrentReconciles})
+	c, err := controller.New("cloneset-controller", mgr, controller.Options{
+		Reconciler: r, MaxConcurrentReconciles: concurrentReconciles,
+		RateLimiter: workqueue.NewItemExponentialFailureRateLimiter(util.BaseDelay, util.MaxDelay)})
```
FillZpp, line 106: The default rate limiter is https://github.com/openkruise/kruise/blob/master/vendor/k8s.io/client-go/util/workqueue/default_rate_limiters.go#L39 , including `ItemExponentialFailureRateLimiter` and `BucketRateLimiter`.
We should not just use `ItemExponentialFailureRateLimiter` here.

pkg/util/rate_limiter.go
```diff
@@ -0,0 +1,14 @@
+package util
+
+import (
+	"flag"
+	"time"
+)
+
+func init() {
+	flag.DurationVar(&BaseDelay, "rate-limiter-base-delay", time.Millisecond*5, "The base delay for rate limiter. Defaults 5ms")
+	flag.DurationVar(&MaxDelay, "rate-limiter-max-delay", time.Minute, "The max delay for rate limiter. Defaults 1min")
```
FillZpp, line 10: `time.Minute` might break the default `1000s`

FillZpp (changes requested):
The default rate limiter is https://github.com/openkruise/kruise/blob/master/vendor/k8s.io/client-go/util/workqueue/default_rate_limiters.go#L39 , including `ItemExponentialFailureRateLimiter` and `BucketRateLimiter`.
We should not just use `ItemExponentialFailureRateLimiter` here.

pkg/controller/cloneset/cloneset_controller.go
shiyan2016: done

pkg/util/rate_limiter.go
shiyan2016: done

FillZpp (approved):
LGTM

PR #449, opened 2020-11-16
Resolve situations where a broadcastJob runs a multiple pods on a Node

devincd:
PodList client is gets list from a cache
```	
// GetClient returns a client configured with the Config. This client may
// not be a fully "direct" client -- it may read from a cache, for
// instance.  See Options.NewClient for more information on how the default
// implementation works.
GetClient() client.Client
```
So there's a good chance that when executing the Reconcile() function, you can't Reconcile the list of pods created last time, resulting in two pods running on one Node

### DIFF
--- a/pkg/controller/broadcastjob/broadcastjob_controller.go
+++ b/pkg/controller/broadcastjob/broadcastjob_controller.go
@@ -408,7 +408,7 @@ func (r *ReconcileBroadcastJob) reconcilePods(job *appsv1alpha1.BroadcastJob,
 					// parallelize pod creation
 					klog.Infof("creating pod on node %s", nodeName)
 					err := r.createPodsOnNode(nodeName, job.Namespace, &job.Spec.Template, job, asOwner(job))
-					if err != nil && errors.IsTimeout(err) {
+					if err != nil && errors.IsTimeout(err) || errors.IsAlreadyExists(err) {
 						// Pod is created but its initialization has timed out.
 						// If the initialization is successful eventually, the
 						// controller will observe the creation via the informer.
@@ -653,6 +653,13 @@ func (r *ReconcileBroadcastJob) createPods(nodeName, namespace string, template 
 	if err != nil {
 		return err
 	}
+	accessor, err := meta.Accessor(object)
+	if err != nil {
+		klog.Errorf("parentObject does not have ObjectMeta, %v", err)
+		return nil
+	}
+	// Specify the pod.Name to prevent the creation of two Pods in a Node
+	pod.Name = fmt.Sprintf("%s-%s", accessor.GetName(), nodeName)
 	pod.Namespace = namespace
 	if len(nodeName) != 0 {
 		pod.Spec.NodeName = nodeName
@@ -670,11 +677,6 @@ func (r *ReconcileBroadcastJob) createPods(nodeName, namespace string, template 
 		return err
 	}
 
-	accessor, err := meta.Accessor(object)
-	if err != nil {
-		klog.Errorf("parentObject does not have ObjectMeta, %v", err)
-		return nil
-	}
 	klog.Infof("Controller %v created pod %v", accessor.GetName(), pod.Name)
 	r.recorder.Eventf(object, corev1.EventTypeNormal, kubecontroller.SuccessfulCreatePodReason, "Created pod: %v", pod.Name)
 


pkg/controller/broadcastjob/broadcastjob_controller.go
```diff
@@ -653,6 +653,13 @@ func (r *ReconcileBroadcastJob) createPods(nodeName, namespace string, template
 	if err != nil {
 		return err
 	}
+	accessor, err := meta.Accessor(object)
+	if err != nil {
+		klog.Errorf("parentObject does not have ObjectMeta, %v", err)
+		return nil
+	}
+	// Specify the pod.Name to prevent the creation of two Pods in a Node
+	pod.Name = fmt.Sprintf("%s-%s", accessor.GetName(), nodeName)
```
Fei-Guo, line 662: To avoid naming conflict, we should use the full name of the broadcastjob CR instead of using name only.

Fei-Guo (commented):
Overall, I think fixing the Pod name is not a clean solution. We should use the standard Kubernetes expectation mechanism to resolve the problem that the informer cache is not updated on time.

pkg/controller/broadcastjob/broadcastjob_controller.go
devincd: I don't think it's necessary. Each object in your cluster has a Name that is unique for that type of resource. BroadcastJob itself wants to run only once on each Node, so a pod's name consist of the name of broadcastJob Name and nodeName is enough to avoid naming conflict.

devincd:
> Overall, I think fixing the Pod name is not a clean solution. We should use the standard Kubernetes expectation mechanism to resolve the problem that the informer cache is not updated on time.



> Overall, I think fixing the Pod name is not a clean solution. We should use the standard Kubernetes expectation mechanism to resolve the problem that the informer cache is not up

devincd:
> Overall, I think fixing the Pod name is not a clean solution. We should use the standard Kubernetes expectation mechanism to resolve the problem that the informer cache is not updated on time.

You are right. But broadcast ieself is intended to run one-time tasks on each Node, so it makes sense to change the podName suffix form a random string to nodeName, and to see at glance which node the task if running on from the podName. And I think it's the quickest way to fix the issue where a broadcastJob runs a multiple pods on a Node

pkg/controller/broadcastjob/broadcastjob_controller.go
Fei-Guo: Broadcastjob is namespace scoped. So you can have two broadcastjob CRs with the same name but in different namespace. They will conflict when counting how many Pods has been running in each node.

Fei-Guo: BroadcastJob is namespace scoped. So if you create two BroadcastJob CRs with the same name but in different namespace, they will conflict by the Pod name when creating Pod in each node.

Fei-Guo:
"and to see at glance which node the task if running on from the podName"  I agree we can improve for this regard, for example, appending node name as a suffix.

But adding expectation is the right thing to do even if you change the Pod name, it avoids unnecessary reconciles and reduce the pressure to apiserver in the worst case. 

Fixing Pod name has another side effect. If the Pods from previous run are not deleted (stuck in deletion) for some reason, the controller stops working when user applies the same bcj CR again.

pkg/controller/broadcastjob/broadcastjob_controller.go
devincd: Pod is also namespace scoped. So it's ok if podName is same in different namespace.

Fei-Guo: You are right. I take it back

devincd:
> Fixing Pod name has another side effect. If the Pods from previous run are not deleted (stuck in deletion) for some reason, the controller stops working when user applies the same bcj CR again.

@Fei-Guo  I dont's see any problem. After the new same name BroadcastJob is created, if the old pod has not deleted for some reason, wait for the old pod to be deleted and re-create the pod the next Reconcile

devincd:
> > Fixing Pod name has another side effect. If the Pods from previous run are not deleted (stuck in deletion) for some reason, the controller stops working when user applies the same bcj CR again.
> 
> @Fei-Guo I dont's see any problem. After the new same name BroadcastJob is created, if the old pod has not deleted for some reason, wait for the old pod to be deleted and re-create the pod the next Reconcile

And I tested it many times without any problems

Fei-Guo:
Thanks for filing the PR, the problem is resolved in #455 hopefully. Please raise an issue if the problem still exists.

PR #452, opened 2020-11-18
lower access to webhook

shiyan2016:
lower access to webhook


Fei-Guo:
Can you describe more about why you need to make this change?

shiyan2016:
> Can you describe more about why you need to make this 

> Can you describe more about why you need to make this change?

First, kruise does not required create/delete webhook access. Second, this can reduce risk when some bugs that cause create/delete webhook. I think the access should be as low as possible.

FillZpp (approved):
LGTM

PR #455, opened 2020-11-18
Add ScaleExpectation for BroadcastJob

Fei-Guo:
### Describe what this PR does

This PR revisits the fix proposed in #449. Thanks @devincd for finding the problem.

Currently BroadcastJob does not handle a rare informer cache delay problem hence it is possible
that the controller will create more than one Pod in one Node. The informer cache delay problem 
is a known problem in Kubernetes community. Usually we use the mechanism of setting
expectations to record the events of object CUD and removing the expectations when the corresponding
events are received in the informer cache. The controller will not proceed if previous expectations have
not completely met. This change adds ScaleExpectation support for BroadcastJob controller.

We use the Kruise built-in ScaleExpectation implementation. One tricky part is that since
the Job Pod uses the generated name, the nodename is added to the expectation instead of
the Poid node, which means we expect there is one Pod created in the desired node. This
expectation has equal effectiveness compared to expecting a particular Pod is created
considering the characteristics of BroadcastJob - create ONE Pod is each qualified node.   

This change also refactors the test code. The tests do not use real informer hence
we give each test a unique name to avoid conflicting in ScaleExpectation map. 

### Describe how to verify it

I have done manual tests and verified that bcj controller works with this change. 
It is hard to conduct an informer cache delay test though.

### Special notes for reviews

I choose not to change the Pod name in this change in order to make 
bcj's Pod name consistent with the Pod name pattern of native Kubernetes Job.

### DIFF
--- a/pkg/controller/broadcastjob/broadcastjob_controller.go
+++ b/pkg/controller/broadcastjob/broadcastjob_controller.go
@@ -33,6 +33,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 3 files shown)


FillZpp (approved):
LGTM

Issue #454, opened 2020-11-18
[feature request] CloneSet Partition support `intstr.IntOrString` format


hellolijj:
**What would you like to be added**:

change cloneset spec partition format from `*int32` to `*intstr.IntOrString`

**Why is this needed**:
sometimes, we want to set partition percentage not number.

PR #456, opened 2020-11-19
cloneset partition support intorstr format

hellolijj:

cloneset partition support intorstr format

https://github.com/openkruise/kruise/issues/454

### DIFF
--- a/apis/apps/v1alpha1/cloneset_types.go
+++ b/apis/apps/v1alpha1/cloneset_types.go
@@ -89,10 +89,12 @@ type CloneSetUpdateStrategy struct {
 	// Type indicates the type of the CloneSetUpdateStrategy.
 	// Default is ReCreate.
 	Type CloneSetUpdateStrategyType `json:"type,omitempty"`
-	// Partition is the desired number of pods in old revisions. It means when partition
-	// is set during pods updating, (replicas - partition) number of pods will be updated.
+	// Partition is the desired number of pods in old revisions.
+	// Value can be an absolute number (ex: 5) or a percentage of desired pods (ex: 10%).
+	// Absolute number is calculated from percentage by rounding up by default.
+	// It means when partition is set during pods updating, (replicas - partition value) number of pods will be updated.
 	// Default value is 0.
-	Partition *int32 `json:"partition,omitempty"`
+	Partition *intstr.IntOrString `json:"partition,omitempty"`
 	// The maximum number of pods that can be unavailable during the update.
 	// Value can be an absolute number (ex: 5) or a percentage of desired pods (ex: 10%).
 	// Absolute number is calculated from percentage by rounding up by default.
--- a/pkg/controller/cloneset/cloneset_controller_test.go
+++ b/pkg/controller/cloneset/cloneset_controller_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/onsi/gomega"
 	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
 	clonesetutils "github.com/openkruise/kruise/pkg/controller/cloneset/utils"
+	"github.com/openkruise/kruise/pkg/util"
 	"github.com/openkruise/kruise/pkg/util/fieldindex"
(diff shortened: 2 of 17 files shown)


apis/apps/v1alpha1/cloneset_types.go
```diff
@@ -92,7 +92,7 @@ type CloneSetUpdateStrategy struct {
 	// Partition is the desired number of pods in old revisions. It means when partition
 	// is set during pods updating, (replicas - partition) number of pods will be updated.
 	// Default value is 0.
-	Partition *int32 `json:"partition,omitempty"`
+	Partition *intstr.IntOrString `json:"partition,omitempty"`
```
Fei-Guo, line 95: You need to revise the comment accordingly since it is not a `int` type anymore.

pkg/controller/cloneset/scale/cloneset_scale_util_test.go
```diff
@@ -384,3 +384,7 @@ func TestCalculateDiffs(t *testing.T) {
 		})
 	}
 }
+
+func getIntOrStr(i intstr.IntOrString) *intstr.IntOrString {
```
Fei-Guo, line 388: Why do you need this method?

pkg/controller/cloneset/cloneset_controller_test.go
```diff
@@ -396,6 +396,10 @@ func getInt32(i int32) *int32 {
 	return &i
 }
 
+func getIntOrStr(i intstr.IntOrString) *intstr.IntOrString {
```
Fei-Guo, line 399: Why is this needed?

pkg/controller/cloneset/update/cloneset_update_test.go
```diff
@@ -71,6 +71,10 @@ func getInt32Pointer(i int32) *int32 {
 	return &i
 }
 
+func getIntOrStr(i intstrutil.IntOrString) *intstrutil.IntOrString {
```
Fei-Guo, line 74: same here?

apis/apps/v1alpha1/cloneset_types.go
```diff
@@ -89,10 +89,12 @@ type CloneSetUpdateStrategy struct {
 	// Type indicates the type of the CloneSetUpdateStrategy.
 	// Default is ReCreate.
 	Type CloneSetUpdateStrategyType `json:"type,omitempty"`
-	// Partition is the desired number of pods in old revisions. It means when partition
-	// is set during pods updating, (replicas - partition) number of pods will be updated.
+	// Partition is the desired number of pods in old revisions.
+	// Value can be an absolute number (ex: 5) or a percentage of desired pods (ex: 10%).
+	// Absolute number is calculated from percentage by rounding up by default.
+	//It means when partition is set during pods updating, (replicas - partition value) number of pods will be updated.
```
Fei-Guo, line 95: Nit:  Add a space between '//' and 'It' in the comment.

Fei-Guo (commented):
I have a high level question:

If a user sets 50% as partition and the replica size is 100, this means 50 Pods are in older version. If the user changes the replica size to 200, the partition changes to 100 from controller perspective. I believe in current implementation, controller will not revert 50 Pods to old version. Will this cause confusion? Such confusion does not exist if an absolute value is used in partition.

Fei-Guo:
@hellolijj This is an API change. Can you address my comment above?

hellolijj:
> @hellolijj This is an API change. Can you address my comment above?

Theoretically, this is what we expect, but I need to do some experiments to verify it

hellolijj:
> @hellolijj This is an API change. Can you address my comment above?

here are my [experiments result](https://docs.google.com/document/d/1P4wg2HfSMj8_oFgIxrrRTs0T2MCAjDAMryXL_-0EQew/edit?usp=sharing) about partition change. It's exactly what we want.

Fei-Guo:
@hellolijj Please fix the CI build failure.

Fei-Guo (approved):

PR #470, opened 2020-12-01
fix client-go rest config not work

shiyan2016:
When the request is large, client-go will throttle the request even if  we adjust the rest config. This PR fixes the problem that client-go rest config does not work.

### DIFF
--- a/pkg/util/gate/custom_resource_gate.go
+++ b/pkg/util/gate/custom_resource_gate.go
@@ -27,10 +27,10 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/discovery"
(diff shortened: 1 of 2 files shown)


Fei-Guo (dismissed):
Thanks for fixing this bug.

FillZpp:
Looks like my mistake 😄

pkg/util/gate/custom_resource_gate.go
```diff
@@ -44,10 +46,11 @@ var (
 	isNotNotFound = func(err error) bool { return !errors.IsNotFound(err) }
 )
 
-func init() {
+func Init() {
 	_ = apis.AddToScheme(internalScheme)
 	cfg, err := config.GetConfig()
```
FillZpp, line 51: This cfg could reuse the rest config in main.go

FillZpp (approved):

PR #471, opened 2020-12-01
Support reserveOrdinal for Advanced StatefulSet

### LINKED ISSUE #307: [feature request] support advanced stateuful pod offline
## What would you like to be added

Support marking advanced stateful pod ordinal offline, a offline pod ordinal serves as a dummy placeholder, advancedstatefulset controller will skip checking the corresponding pod status and the pod will be treated as a condemed pod.

## Why is this needed 

Consider the following scenarios:

1. A stateful pod is running on and strongly bound to a node, once the node unavailable for some reason(e.g. network partition, one of csi, cni or cri not ready), advancedstatefulset controller will stuck on waiting the pod on the failed node become health again.
E.g. stateful-pod-1 has mounted a local pv volume on node-1, node-1 has network issue at some time(or permanently) that cause stateful-pod-1 status cannot be reported. Advancedstatefulset controller has to wait on stateful-pod-1 become running & ready, all other stateful pods with the same parent as stateful-pod-1 will stuck on updating, force delete stateful-pod-1 doesn't help because it will be rescheduled on node-1 for its pv volume is on node-1 and will enter 'Pending' state. 
Real-world examples are 
https://github.com/kubernetes/kubernetes/issues/64869 
https://github.com/kubernetes/kubernetes/issues/74947#issuecomment-473603927
https://github.com/kudobuilder/kudo/issues/1482

2. A stateful pod is running on and weakly bound(hasn't bound to local pv volumes) to a node, the node with the stateful pod goes down accidentally and permanently, force delete and reschedule the pod to another node may fix this situation(e.g., https://github.com/kubernetes/kubernetes/issues/74689 ), but if we have no additional nodes available, the we eventually will face the same problem in scenarios 1.
E.g. all other nodes unavailable happens when the rescheduled pod violates antiaffinity rules with existing pods, insufficient resources or port conflicts on the current nodes.

3. I want delete one or more stateful pods in an unordered way. Suppose I have tasks running in a pod, I have a horizontal auto scaler knows how to scale the number of the pod in proportion to the total number of the tasks(sum up of each pod tasks). It works as expected when scaling up, things become a bit tricky when scaling down. 
E.g. at somet time, the auto scaler suggests that the replias should scale from 4 down to 2,  while `pod-0, pod-1` have no task running, `pod-2, pod-3` have tasks running. Simply scale the statefulset's replicas down to 2 will kill `pod-2,pod-3`, even the running tasks can be decomissioned and migrated to `pod-0,pod-1`, the action can be expensive because we have to rerun those tasks(especially for long running jobs) and break the overall pipeline.
A real-world example is when I deploy spark executors as advancedstateful pods and I want to follow autoscaling rules like 
https://github.com/apache/spark/blob/master/docs/job-scheduling.md#resource-allocation-policy or  
https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/autoscaling#create_an_autoscaling_policy or
https://docs.databricks.com/clusters/configure.html#how-autoscaling-behaves

For the above scenarios, selectively offline unwanted pods can mitigate the problem. For scenarios 1 and 2, offline stukcing pods so statefulset won't stucking on updating/upgrading, once things getting better, e.g. either node become available or user successfully migrated the pod to another place, we can online(let advancedstatefulset controller reconcile them again) those pods. For scenarios 3, scale down action can be converted to operate idle pods offline, optionally scale up can be converted to operate pod online if there exisiting previously offlined pod.

## Proposal

Create a way to define advanced stateful pod offline, this will be an additional field to the advanced statefulset annotations. Suppose the annotation key is "offline-pod.kruise.io/ordinal-list", when inserting pod ordianls into metadata.annotations["offline-pod.kruise.io/ordinal-list"], advancedstatefulset controller will delete corespondeing pods(if exists) and the pods  ordinals serves as dummy  placeholders, the controller will delete those pods(perserve pvc if exists) and will on longer watch the coresponding pods status on updating/scaling.

## Implementation Details/Notes/Constraints

Inserting pod ordinals into metadata.annotations["offline-pod.kruise.io/ordinal-list"] triggering statefulset controller offline pod when ordinals in range [0, statefulsets.apps.kruise.io.spec.replicas),
removing pod ordinals from metadata.annotations["offline-pod.kruise.io/ordinal-list"] triggering statefulset controller online pod when ordinals in range [0, statefulsets.apps.kruise.io.spec.replicas),

### API Changes:
A field key "offlinedReplicas" will be add into statefulset.status indicates the number of pods offlined by the controller.
```
    // OfflinedReplicas is the number of Pods offlined by the StatefulSet controller.
    OfflinedReplicas int32 `json:"offlinedReplicas"`
```

A filed key "OFFLINED" will be add into CRD config as a additional column
```
  - JSONPath: .status.offlinedReplicas
    description: The number of pods offlined.
    name: OFFLINED
    type: integer
```

An annotation key "offline-pod.kruise.io/ordinal-list" will be reserved for controller to store the offlined ordinals, its value expects to an ordinal list. 
```
    // offlineOrdinalAnnotationKey indicates offline pod ordinals, the key's value expects to be a integer list, e.g. "[1,2]".
    // statefulset treats a offline ordianl as a dummy placeholders, it no longer monitors the corresponding pod,
    // operations such as delete/update on the offline pod has no affect on statefulset, a offlined pod is a condemed pod and
    // will be deleted
    offlineOrdinalAnnotationKey = "offline-pod.kruise.io/ordinal-list"
```

this is an opt-in feature and no breaking API changes, without specifying offline ordinals, the controller should fully compitable with current behaviors, when enableed this feature things different in

### MaxUnavailable Rolling Update Strategy
in parallel mode, the number of offlined pods will not be counted.

### In-Place Pod Update Strategy
inplace update ignore the offlined pods.

### Priority Unordered Rolling Update Strategy
offlined pods ordinals will be ignored and will not be sorted. 

## PoC and Demo
There is a [commit](https://github.com/yuchengwu/kruise/commit/d4b32a55f072703fe050890f79aef1a432ec4817) here with a working Proof of concept for this proposal, it's just a draft but should help illustrate what these changes would look like.
Please view this video(bilibili: https://www.bilibili.com/video/BV1xt4y197wb/, youtube: https://youtu.be/acyEVT487tE ) if you want to see what the PoC looks like in action.

## Risks and Mitigations
TBD

### Test Plan
- Units test in statefulset package test a variety of scenarios.
- New E2E Tests to validate that advanced statefulset controller offline pods behave as expected e.g:
- Pods whose ordinals in [0, statefulsets.apps.kruise.io.spec.replicas) should be deleted
- Scaling up/down should skip offline ordinal, both in orderReady(podManagementPolicy is OrderedReady) and parallel mode(podManagementPolicy is Parallel)
- OnDelete update should skip offline ordinal, both in orderReady and parallel mode
- Rolling update with inplace/recreate strategy should skip offline ordinal, both in orderReady and parallel mode
- Rolling update with MaxUnavailable strategy should skip offline ordinal in parallel mode
- Rolling update with MaxUnavailable strategy in combination with inplace/recreate strategy should skip offline ordinal in parallel mode
- Rolling update with Priority Unordered Rolling Update Strategy should skip offline ordinal in parallel mode
- Partition or Pause value in Rolling update should have no effect on offline ordinal, that means whether Pause is true or false, the pod should be offlined, whether what the Partition value is, the pod should be offlined.

## Upgrade / Downgrade Strategy
When upgrading no changes should be needed to maintain existing behaviour as all of this behaviour is optional and disabled by default. To activate the feature they will need to enable the feature and mark their pods as offline pods in the statefulset annotations.

When downgrading, no changes are required, but users should be cautioned that ordinals in annotations["offline-pod.kruise.io/ordinal-list"] field will be onlined.

## Alternatives
For the above scenarios 1 and 2, situation can be mitigated by setting proper update strategy:
- As @furykerry mentioned that advancestateset support setting maxUnavailable to value larger than one, so that rolling upgrade will not stuck on waiting pod. Thats means in parallel mode, we can tolerates at most (maxUnavailable-1) pods failure. 
- Setting Priority Unordered Rolling Update Strategy can also helps to upgrading in an unordered way(give failure pods  low priority and updating them at the end) so the controller won't stuck

FillZpp:

Support reserveOrdinal for Advanced StatefulSet

fixes #307 

### Ⅲ. Special notes for reviews
Note that this feature is more like the [PingCAP impl](https://docs.pingcap.com/tidb-in-kubernetes/stable/advanced-statefulset#specify-the-pod-to-be-scaled-in) rather than the description in #307 .

 Issue #307 wants `replicas` contains the reserve ordinal. If so, for `replicas=5` and `reserveOrdinal=[1,3]`, the existing Pods will be `[0, 2, 4]`.

But we consider `replicas` to be the expectation number of running Pods. For `replicas=5` and `reserveOrdinal=[1,3]`, the existing Pods will be `[0, 2, 4, 5, 6]`.


pkg/controller/statefulset/stateful_set_control.go
```diff
@@ -309,7 +310,11 @@ func (ssc *defaultStatefulSetControl) updateStatefulSet(
 	status.CollisionCount = utilpointer.Int32Ptr(collisionCount)
 	status.LabelSelector = selector.String()
 
-	replicaCount := int(*set.Spec.Replicas)
+	reserveOrdinalIds := sets.NewInt()
+	if len(set.Spec.ReserveOrdinal) > 0 {
+		reserveOrdinalIds.Insert(set.Spec.ReserveOrdinal...)
+	}
+	replicaCount := int(*set.Spec.Replicas) + reserveOrdinalIds.Len()
 	// slice that will contain all Pods such that 0 <= getOrdinal(pod) < set.Spec.Replicas
```
Fei-Guo, line 318: change this comment as well.

pkg/controller/statefulset/stateful_set_control.go
```diff
@@ -342,12 +347,12 @@ func (ssc *defaultStatefulSetControl) updateStatefulSet(
 			}
 		}
 
-		if ord := getOrdinal(pods[i]); 0 <= ord && ord < replicaCount {
+		if ord := getOrdinal(pods[i]); 0 <= ord && ord < replicaCount && !reserveOrdinalIds.Has(ord) {
```
Fei-Guo, line 350: update the comment accordingly.

pkg/controller/statefulset/stateful_set_control.go
```diff
@@ -342,12 +347,12 @@ func (ssc *defaultStatefulSetControl) updateStatefulSet(
...
 			// if the ordinal of the pod is within the range of the current number of replicas,
 			// insert it at the indirection of its ordinal
 			replicas[ord] = pods[i]
 
-		} else if ord >= replicaCount {
+		} else if ord >= replicaCount || reserveOrdinalIds.Has(ord) {
 			// if the ordinal is greater than the number of replicas add it to the condemned list
```
Fei-Guo, line 356: Update the comment accordingly.

Fei-Guo (commented):
We need to document the controller behavior of removing `reserveOridnals` as well.

For example, if reserveOridnals = [0, 1, 2, 3], replica =4, Pod [4,5,6,7] will be created. If we clear the reserveOridnals, Pod[4,5,6,7] will be removed and Pod [1,2,3,4] will be created.

It seems that in most use cases, reserveOrdinals and replica need to be changed together to avoid unexpected Pod deletion/creation.

Fei-Guo (approved):

PR #481, opened 2020-12-16
fix isExpected function bugs

kadisi:
Signed-off-by: kadisi <iamkadisi@163.com>

for statefulset:

```
func (a *StatefulSetAdapter) IsExpected(obj metav1.Object, revision string) bool {
	return obj.GetLabels()[appv1.ControllerRevisionHashLabelKey] != revision
}
```

The workload never has this label `appv1.ControllerRevisionHashLabelKey`, so `IsExpected` always `true`

so when kruise controller reconciler, it  always updates the workload。

so need change label from `appv1.ControllerRevisionHashLabelKey` to `alpha1.ControllerRevisionHashLabelKey`


FillZpp (approved):
Thanks for fixing this.

PR #486, opened 2020-12-17
adopt matching pods

27149chen:
This PR fixed an issue that once a pod without an owner matches the label selector of a cloneSet, it is not adopted by the cloneSet. Which does not match the behavior of a Deployment/ReplicaSet

NONE

Improved the test case `TestClaimPods`

1. remove a label (a label which has impact on selector) of a pod owned by a cloneSet
2. check that the pod is released (has no owner)
3. add the label back
4. check that the pod is adopted by the origin cloneSet (has the same owner again)

### DIFF
--- a/pkg/controller/cloneset/cloneset_controller.go
+++ b/pkg/controller/cloneset/cloneset_controller.go
@@ -229,6 +229,9 @@ func (r *ReconcileCloneSet) doReconcile(request reconcile.Request) (res reconcil
 	// list all pods to include the pods that don't match the cs`s selector
 	// anymore but has the stale controller ref.
 	filteredPods, err := r.getAllActivePodsInNamespace(instance)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
 
 	filteredPods, err = r.claimPods(instance, filteredPods)
 	if err != nil {
--- a/pkg/controller/cloneset/cloneset_controller.go
+++ b/pkg/controller/cloneset/cloneset_controller.go
@@ -220,13 +220,16 @@ func (r *ReconcileCloneSet) doReconcile(request reconcile.Request) (res reconcil
 		return reconcile.Result{RequeueAfter: expectations.ExpectationTimeout - unsatisfiedDuration}, nil
 	}
 
-	// list all active Pods and PVCs belongs to cs
-	filteredPods, filteredPVCs, err := r.getOwnedResource(instance)
+	// list all active PVCs belongs to cs
+	filteredPVCs, err := r.getOwnedResource(instance)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
 
-	//release Pods ownerRef
+	// list all pods to include the pods that don't match the cs`s selector
+	// anymore but has the stale controller ref.
+	filteredPods, err := r.getAllActivePodsInNamespace(instance)
+
 	filteredPods, err = r.claimPods(instance, filteredPods)
 	if err != nil {
 		return reconcile.Result{}, err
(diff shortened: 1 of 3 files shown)


pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -509,6 +507,11 @@ func (r *ReconcileCloneSet) truncateHistory(
 	return nil
 }
 
+// claimPods tries to take ownership of a list of Pods.
+//
+// It will reconcile the following:
+//   * Adopt orphans if the selector matches.
+//   * Release owned objects if the selector no longer matches.
```
Fei-Guo, line 514: Is this new logic added by this change? I cannot find it here and there.

27149chen: The adopt/release logic are already implemented in function `manager.ClaimOwnedObjects`. What I did in this pr is changing the scope of the incoming pods (from pods owned by the cloneSet to all pods in the same namespace) which makes it possible to consume pods with no owner but matching the cloneSet selector.

Fei-Guo: Sorry, I don't quite understand what "Release owned objects " mean here, do you mean the controller will remove the ownerreference from the Pod if the Pod's selector does not match with CS selector anymore?

27149chen: yes, you are right, this operation is called "release" in https://github.com/openkruise/kruise/blob/65a2e644190649731cd96fcc5f69332828f6115e/pkg/util/refmanager/ref_manager.go#L228

I think it is inspired by https://github.com/kubernetes/kubernetes/blob/979c644df2ad051015a8ccf6475e3476eda0cea0/pkg/controller/controller_ref_manager.go#L87

27149chen: btw, "Release owned objects" is introduced in #417 , this pr tries to support "Adopt orphans"

FillZpp:
@27149chen  I disagree with this adoption. CloneSet creates Pods and PVCs with owner reference, and both Pods and PVCs should have `apps.kruise.io/cloneset-instance-id` in labels.
These are rules of CloneSet and Pods must observe them. You can not optionally adopt a Pod just because its label matched the selector.

FillZpp (changes requested):
I disagree with this adoption. CloneSet creates Pods and PVCs with owner reference, and both Pods and PVCs should have `apps.kruise.io/cloneset-instance-id` in labels.
These are rules of CloneSet and Pods must observe them. You can not optionally adopt a Pod just because its label matched the selector.

27149chen:
> I disagree with this adoption. CloneSet creates Pods and PVCs with owner reference, and both Pods and PVCs should have `apps.kruise.io/cloneset-instance-id` in labels.
> These are rules of CloneSet and Pods must observe them. You can not optionally adopt a Pod just because its label matched the selector.

@FillZpp , I'm I appreciate your point of view, but I am not the initiator of the introduction of this "adoption", I just enable it for cloneSet, like #417 which enables "release". Both "adopt" and "release" exist in Kruise and they are gneral mechanisms which are used in many workloads (advancedStatefulSet, unitedDeployment, ...). Also they are gneral mechanisms in k8s which are used in all native workloads (replicaSet, statefulSet, daemonSet, ...). If we do not support it, the behavior in cloneSet will be strange, since it is different from native workloads and even other Kruise workloads.
Talking about the `apps.kruise.io/cloneset-instance-id` label, if we rely on it, we (especially the cloneSet controller) should make sure it exists in all owned pods, even when it is removed by others for some reason or not found for a new adoption.

27149chen:
@FillZpp  @Fei-Guo  any update?

FillZpp:
@27149chen If you want to adopt a Pod, you should also adopt the PVC related to it, which means they have the same value of `apps.kruise.io/cloneset-instance-id` label. Not like StatefulSet, CloneSet always puts owner reference into both Pods and PVCs. 

And many things you should also consider, such as:
1. What if the Pod has one or more PVCs related but not match the `volumeClaimTemplates` defined in current CloneSet?
2. What if the Pod has no PVC related but there is `volumeClaimTemplates` defined in current CloneSet?
3. What if the Pod has no `controller-revision-hash` in label, how does current CloneSet manage its version?
4. What if the ControllerRevision of Pod's `controller-revision-hash` has been deleted or belongs to other CloneSet, how does current CloneSet manage its version?

In my opinion, this auto-adoption brings big risk and I don't think it is necessary for most users.

27149chen:
> @27149chen If you want to adopt a Pod, you should also adopt the PVC related to it, which means they have the same value of `apps.kruise.io/cloneset-instance-id` label. Not like StatefulSet, CloneSet always puts owner reference into both Pods and PVCs.
> 
> And many things you should also consider, such as:
> 
> 1. What if the Pod has one or more PVCs related but not match the `volumeClaimTemplates` defined in current CloneSet?
> 2. What if the Pod has no PVC related but there is `volumeClaimTemplates` defined in current CloneSet?
> 3. What if the Pod has no `controller-revision-hash` in label, how does current CloneSet manage its version?
> 4. What if the ControllerRevision of Pod's `controller-revision-hash` has been deleted or belongs to other CloneSet, how does current CloneSet manage its version?
> 
> In my opinion, this auto-adoption brings big risk and I don't think it is necessary for most users.

@FillZpp , good point! I will think again of the things you mentioned, thank you.
But before that, I want to ask why CloneSet is so different from other kruise workloads which support gneral k8s mechanisms such as "release" and "adoption". Can we rethink whether the current logic is reasonable, is it too complex?

27149chen:
@FillZpp  can you answer my question?

PR #504, opened 2021-01-06
Fix revision label when collision

LeoLiuYan:
Signed-off-by: LeoLiuYan <929908264@qq.com>

Use `**collisionCount++` to calculate the hash value of `controllerRevision` if there is a collision happen,  and we should change the hash label of that `controllerRevision` object.


FillZpp (approved):
LGTM

PR #512, opened 2021-01-12
the default sidecarset updateStrategy is set to RollingUpdate

zmberg:
1. the default sidecarset updateStrategy is set to RollingUpdate
2. NotUpdate strategy also can update SidecarSet.Status
3. the default strategy MaxUnavailable is set to 1


FillZpp (approved):
LGTM

Issue #516, opened 2021-01-18
[feature request] support lifecycle hook for AdvancedStatefulSet 


shiyan2016:
**What would you like to be added**:
support lifecycle hook for AdvancedStatefulSet 

**Why is this needed**:

PR #525, opened 2021-01-29
support lifecycle hook for AdvancedStatefulSet

shiyan2016:
support lifecycle hook for AdvancedStatefulSet

https://github.com/openkruise/kruise/issues/516


pkg/controller/statefulset/stateful_set_control.go
```diff
@@ -602,9 +610,28 @@ func (ssc *defaultStatefulSetControl) updateStatefulSet(
 		}
 	}
 
+	// refresh states for all pods
+	requeueDuration := requeueduration.Duration{}
```
FillZpp, line 614: There is no need to new `requeueDuration` here, for the `duration` returned below can directly push into durationStore.

FillZpp (commented):
Please add some unit tests for advanced statefulset lifecycle hook.

FillZpp (approved):
LGTM

PR #529, opened 2021-02-02
fix condition for pod state from Updated to Normal

### LINKED ISSUE #528: [BUG] inplace update lifecycle does not work correct
**What happened**:
create a cloneset as follows:
```
kind: CloneSet
metadata:
  name: web
spec:
  lifecycle:
    inPlaceUpdate:
      labelsHandler:
        foo-unready-block: "true"
        test-unready-block: "true"
  updateStrategy:
    maxUnavailable: 1
    partition: 0
    type: InPlaceIfPossible
  replicas: 2
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
        foo-unready-block: "true"
        test-unready-block: "true"
    spec:
      containers:
        - name: nginx
          image: nginx
          imagePullPolicy: IfNotPresent
          ports:
            - containerPort: 80
```
And then just update the cloneset image to make it inplace update, it will cause one pod state change to PreparingUpdate, then update the pod's label  foo-unready-block to false, it will keep PreparingUpdate state. Then update the pod's label  test-unready-block to false, the pod state will changed to Updating and start inplace update. When inplace update done, the pod state will changed to Updated. Then when I just change the pod label  foo-unready-block to true. The pod will changed to Normal state. And another pod will changed to PreparingUpdate state.

Pod from Updated to Normal state, it should  satified all the hook.
 
**Environment**:
- Kruise version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

shiyan2016:
fix condition for pod state from Updated to Normal

fixes #528

### DIFF
--- a/pkg/controller/cloneset/update/cloneset_update.go
+++ b/pkg/controller/cloneset/update/cloneset_update.go
@@ -160,7 +160,7 @@ func (c *realControl) refreshPodState(cs *appsv1alpha1.CloneSet, coreControl clo
 	case appspub.LifecycleStateUpdated:
 		if cs.Spec.Lifecycle == nil ||
 			cs.Spec.Lifecycle.InPlaceUpdate == nil ||
-			lifecycle.IsPodHooked(cs.Spec.Lifecycle.InPlaceUpdate, pod) {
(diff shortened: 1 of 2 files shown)


FillZpp (approved):
LGTM

Issue #531, opened 2021-02-04
[BUG] CloneSet name length can't be greater than 63


fatedier:
**What happened**:
Create a CloneSet with name `sample-long-long-long-long-long-long-long-long-long-long-long-long-name`.

CloneSet controller will create a pod with label `controller-revision-hash` and value is `{CloneSet name}-{hash}`.

Label value length must be less than 64 bytes, so pod create request will be rejected by kube-apiserver.

**What you expected to happen**:
Allowed to create a CloneSet with name length <= 253

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

Same issue for StatefulSet: https://github.com/kubernetes/kubernetes/issues/79337

I prefer to change `controller-revision-hash` label format and remove `CloneSet name` from it, so it will not be affected by CloneSet name length.

From `controller-revision-hash: sample-long-long-long-long-long-long-long-long-long-long-long-long-name-54d7c84944` to `controller-revision-hash: 54d7c84944`

To compatible with exist cloneset and pod, we can add a new function such as:

```go
func IsPodRevisionEqual(pod *v1.Pod, revision *apps.ControllerRevision) bool {
    podRevision := GetPodRevision(pod)
    if podRevision == revision.Name || podRevision == GetRevisionHash(revision) {
        return true
    }
    return false
}
```

**Environment**:
- Kruise version: 0.7.0
- Kubernetes version (use `kubectl version`): 1.14
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

FillZpp:
@fatedier Good point, we can do this. And also a flag should be added to let users choose to use the feature, because I'm afraid some users may already depend on the `controller-revision-hash` label to be the full name of ControllerRevision.

So... are you interested in implementing it? 😄

fatedier:
Sure.

PR #536, opened 2021-02-22
add CloneSetHashOnlyRevisionName feature-gate

fatedier:
Add ClonetSetSimpleRevisionLabel featrue-gate, default value is false. If enabled, CloneSet will create pod with simple revision hash label not including it's name.

So we can create CloneSet with name length greater than 63.

Fixes https://github.com/openkruise/kruise/issues/531

### DIFF
--- a/pkg/features/kruise_features.go
+++ b/pkg/features/kruise_features.go
@@ -32,11 +32,15 @@ const (
 
 	// PodWebhook enables webhook for Pods creations. This is also related to SidecarSet.
 	PodWebhook featuregate.Feature = "PodWebhook"
+
+	// CloneSetHashOnlyRevisionName enables CloneSet controller only set revision hash name to pod.
+	CloneSetHashOnlyRevisionName featuregate.Feature = "CloneSetHashOnlyRevisionName"
 )
 
 var defaultFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
-	PodWebhook:   {Default: true, PreRelease: featuregate.Beta},
-	KruiseDaemon: {Default: true, PreRelease: featuregate.Beta},
+	PodWebhook:                   {Default: true, PreRelease: featuregate.Beta},
+	KruiseDaemon:                 {Default: true, PreRelease: featuregate.Beta},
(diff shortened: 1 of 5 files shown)


pkg/features/kruise_features.go
```diff
@@ -32,11 +32,15 @@ const (
 
 	// PodWebhook enables webhook for Pods creations. This is also related to SidecarSet.
 	PodWebhook featuregate.Feature = "PodWebhook"
+
+	// ClonetSetSimpleRevisionLabel enables simple revision hash label for CloneSet.
+	ClonetSetSimpleRevisionLabel featuregate.Feature = "ClonetSetSimpleRevisionLabel"
```
Fei-Guo, line 37: How about change the feature name to "ClonetSetHashOnlyRevisionName" ?

Fei-Guo (commented):
This looks reasonable to me. A minor comment below.

Can you exclude all the changes to the vendor directory and go.sum? The PR should only include the necessary changes for the feature.

fatedier:
> Can you exclude all the changes to the vendor directory and go.sum? The PR should only include the necessary changes for the feature.

image


@Fei-Guo I remove the changes of vendor files and it failed on unit-tests. Vendor files are auto generated after i execute `make generate manifests`.

Fei-Guo:
> > Can you exclude all the changes to the vendor directory and go.sum? The PR should only include the necessary changes for the feature.


Can you please double check if this import "k8s.io/kubernetes/pkg/controller/history" introduces so many vendor code changes? If so,  we may just redefine the `history.ControllerRevisionHashLabel` in cloneset repo to avoid Kubernetes dependency.

fatedier:
> > > Can you exclude all the changes to the vendor directory and go.sum? The PR should only include the necessary changes for the feature.
> 
> Can you please double check if this import "k8s.io/kubernetes/pkg/controller/history" introduces so many vendor code changes? If so, we may just redefine the `history.ControllerRevisionHashLabel` in cloneset repo to avoid Kubernetes dependency.

@Fei-Guo It's strange that `go.mod and go.sum` are auto modified when i run `go mod tidy` in my local environment. I remove the changes manually and it works now.

PR #544, opened 2021-03-04
Allows the customization of the container manager socket location

cerebrate:

This PR enables the Helm chart to specify a location other than `/var/run` for the container manager socket (i.e., docker.sock, container.sock, etc.). This allows the kruise-daemon element of kruise 0.8.0 to function on systems in which this socket is not in the standard location; for example, on k3s-based Kubernetes installations, containerd.sock is located in `/var/run/k3s/containerd` rather than `/var/run/containerd`, resulting in kruise-daemon failing on startup.

None; I opened no issue because it was an easy fix. :smiley:

No test case applies on standard system; this would require one of the different-location systems available to test on specifically.


FillZpp (approved):
Good job! It is so important for those systems with non-standard socket location.
Thanks a lot.

Issue #547, opened 2021-03-06
[BUG] ImagePullJob reports error if image already exists


cerebrate:
**What happened**:

I am using ImagePullJob to ensure that certain images needed by all nodes are always downloaded in advance. However, when this is configured and executes, I receive a large number of PullImageFailed warning events of the form:

```
Image drnic/kubectl:latest pulling image drnic/kubectl:latest error get error in pull response: image "docker.io/drnic/kubectl@sha256:248c0c48844f1169c99b8c48b48cc752cc5f48124ac743cbccbe223437c87c0d": already exists
```

for those nodes on which the image has already been downloaded. The pull is recorded as failed in the NodeImage and ImagePullJob objects.

**What you expected to happen**:

If the image already exists on the node, the ImagePullJob should detect this and succeed silently without even trying to pull the image, recording a success in the NodeImage and ImagePullJob objects.

(While a relatively harmless failure, insofar as the image has to exist on the node in order for the failure to happen, it does make it hard to track successful image pulls and determine which failures are "real" and which are not.)

**How to reproduce it (as minimally and precisely as possible)**:

Set up an ImagePullJob for an image that already exists on one but not all nodes. The one that failed as above was:

```
apiVersion: apps.kruise.io/v1alpha1
kind: ImagePullJob
metadata:
  name: kubectl-everywhere
  namespace: infrastructure
spec:
  image: drnic/kubectl:latest
  parallelism: 2
  completionPolicy:
    type: Never
```

**Environment**:
- Kruise version: 0.8.0
- Kubernetes version (use `kubectl version`):

Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:50:19Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.4+k3s1", GitCommit:"838a906ab5eba62ff529d6a3a746384eba810758", GitTreeState:"clean", BuildDate:"2021-02-22T19:49:27Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}

- OS (e.g: `cat /etc/os-release`): Mixed (see below)
- Kernel (e.g. `uname -a`): Mixed (see below)

[...]

PR #548, opened 2021-03-09
imagepulljob: use containerd/errdefs instead of moby

fuweid:
Signed-off-by: Wei Fu <fuweid89@gmail.com>

If the image has been downloaded into containerd, the repull will show error. 
The root cause is using wrong package to detect the error.

fixes: #547 

will add it in the follow-up

verify the result by log


FillZpp (approved):
LGTM

PR #551, opened 2021-03-10
support reserved ordinal large than replica

plpan:
Support reserveOrdinals which is large than `replicas+len(reservedOrdinals)`.

Imagine, sts has five replicas with the reservedOrdinals set to be {0,5}, then all the created pod are {1,2,3,4,6}. 

Later, when I want to scale sts to 4 replicas, nothing else but the spec.replicas should be changed.

What's more, when I want to scale sts back to 5 replicas, the reservedOrdinals previously set can still take effects.

On the other hand, if `0 <= order < replicas+len(reservedOrdinals)` is a must, mutating webhook might be a better choice.

### DIFF
--- a/pkg/controller/statefulset/stateful_set_control.go
+++ b/pkg/controller/statefulset/stateful_set_control.go
@@ -314,11 +314,14 @@ func (ssc *defaultStatefulSetControl) updateStatefulSet(
 	status.CollisionCount = utilpointer.Int32Ptr(collisionCount)
 	status.LabelSelector = selector.String()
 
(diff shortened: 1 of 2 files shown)


plpan:
Is `0 <= order < len(reservedOrdinals) + replicas` a must?

plpan:
@furykerry In current, if I scale down the sts with reservedOrdinals set to 0 replicas, reservedOrdinals must be emptied at the same time. I wonder if this is the desired behavior?

FillZpp (approved):

PR #558, opened 2021-03-12
change cloneset scale expectation structure

hellolijj:
Signed-off-by: JunjunLi <junjunli666@gmail.com>

change cloneset scale expectation structure 

cloneset scale expectation map structure depends on cloneset's name and namespaces.

but in some cases, pod's `ownReference` will be set empty by other's task. in that case,  pod's owner, that is cloneset,  scale expectation will not be observed, that result it dirty pod in workload.

https://github.com/hellolijj/kruise/blob/fixup/cloneset-expectation-key-structure/pkg/controller/cloneset/cloneset_event_handler_test.go#L737-L788

### DIFF
--- a/pkg/controller/cloneset/cloneset_controller.go
+++ b/pkg/controller/cloneset/cloneset_controller.go
@@ -192,7 +192,6 @@ func (r *ReconcileCloneSet) doReconcile(request reconcile.Request) (res reconcil
 			// Object not found, return.  Created objects are automatically garbage collected.
 			// For additional cleanup logic use finalizers.
 			klog.V(3).Infof("CloneSet %s has been deleted.", request)
-			clonesetutils.ScaleExpectations.DeleteExpectations(request.String())
 			clonesetutils.UpdateExpectations.DeleteExpectations(request.String())
 			return reconcile.Result{}, nil
 		}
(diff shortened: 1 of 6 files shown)


Fei-Guo:
Isn't this breaking the expectation functionality? Is it valid that a cloneset's Pod does not have owner reference being set?

pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -213,7 +212,7 @@ func (r *ReconcileCloneSet) doReconcile(request reconcile.Request) (res reconcil
 	}
 
 	// If scaling expectations have not satisfied yet, just skip this reconcile.
-	if scaleSatisfied, unsatisfiedDuration, scaleDirtyPods := clonesetutils.ScaleExpectations.SatisfiedExpectations(request.String()); !scaleSatisfied {
+	if scaleSatisfied, unsatisfiedDuration, scaleDirtyPods := clonesetutils.ScaleExpectations.SatisfiedExpectations(request.Namespace); !scaleSatisfied {
```
FillZpp, line 215: It will cause all CloneSets in a namespace paused, if one Pod in the namespace has not satisfied.
Any other good ideas?

FillZpp (changes requested):

PR #576, opened 2021-03-30
Support pre-download images for CloneSet in-place update

FillZpp:

Support pre-download images for CloneSet in-place update

Reference to proposal [image pre-download during in-place update](https://github.com/openkruise/kruise/blob/master/docs/proposals/20210325-imagepredownload-during-inplace-update.md).

### DIFF
--- a/pkg/controller/cloneset/cloneset_controller.go
+++ b/pkg/controller/cloneset/cloneset_controller.go
@@ -20,7 +20,10 @@ package cloneset
 import (
 	"context"
 	"flag"
+	"fmt"
 	"time"
+
+	intstrutil "k8s.io/apimachinery/pkg/util/intstr"
 
 	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
 	kruiseclient "github.com/openkruise/kruise/pkg/client"
@@ -28,11 +31,15 @@ import (
 	revisioncontrol "github.com/openkruise/kruise/pkg/controller/cloneset/revision"
 	synccontrol "github.com/openkruise/kruise/pkg/controller/cloneset/sync"
 	clonesetutils "github.com/openkruise/kruise/pkg/controller/cloneset/utils"
+	"github.com/openkruise/kruise/pkg/features"
 	"github.com/openkruise/kruise/pkg/util"
 	utildiscovery "github.com/openkruise/kruise/pkg/util/discovery"
 	"github.com/openkruise/kruise/pkg/util/expectations"
+	utilfeature "github.com/openkruise/kruise/pkg/util/feature"
 	"github.com/openkruise/kruise/pkg/util/fieldindex"
 	historyutil "github.com/openkruise/kruise/pkg/util/history"
+	utilimagejob "github.com/openkruise/kruise/pkg/util/imagejob"
+	"github.com/openkruise/kruise/pkg/util/inplaceupdate"
 	"github.com/openkruise/kruise/pkg/util/ratelimiter"
 	"github.com/openkruise/kruise/pkg/util/refmanager"
 	apps "k8s.io/api/apps/v1"
@@ -41,6 +48,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/record"
@@ -62,6 +70,9 @@ func init() {
 
 var (
 	concurrentReconciles = 3
+
+	isPreDownloadDisabled             bool
+	minimumReplicasToPreDownloadImage int32 = 3
 )
 
 // Add creates a new CloneSet Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
@@ -69,6 +80,11 @@ var (
 func Add(mgr manager.Manager) error {
 	if !utildiscovery.DiscoverGVK(clonesetutils.ControllerKind) {
 		return nil
+	}
+	if !utildiscovery.DiscoverGVK(appsv1alpha1.SchemeGroupVersion.WithKind("ImagePullJob")) ||
+		!utilfeature.DefaultFeatureGate.Enabled(features.KruiseDaemon) ||
+		!utilfeature.DefaultFeatureGate.Enabled(features.PreDownloadImageForInPlaceUpdate) {
+		isPreDownloadDisabled = true
 	}
 	return add(mgr, newReconciler(mgr))
 }
@@ -257,8 +273,19 @@ func (r *ReconcileCloneSet) doReconcile(request reconcile.Request) (res reconcil
 		klog.V(4).Infof("Not satisfied update for %v, updateDirtyPods=%v", request.String(), updateDirtyPods)
 		return reconcile.Result{RequeueAfter: expectations.ExpectationTimeout - unsatisfiedDuration}, nil
 	}
+
 	// If resourceVersion expectations have not satisfied yet, just skip this reconcile
+	clonesetutils.ResourceVersionExpectations.Observe(updateRevision)
+	if isSatisfied, unsatisfiedDuration := clonesetutils.ResourceVersionExpectations.IsSatisfied(updateRevision); !isSatisfied {
+		if unsatisfiedDuration < expectations.ExpectationTimeout {
+			klog.V(4).Infof("Not satisfied resourceVersion for %v, wait for updateRevision %v updating", request.String(), updateRevision.Name)
+			return reconcile.Result{RequeueAfter: expectations.ExpectationTimeout - unsatisfiedDuration}, nil
+		}
+		klog.Warningf("Expectation unsatisfied overtime for %v, wait for updateRevision %v updating, timeout=%v", request.String(), updateRevision.Name, unsatisfiedDuration)
+		clonesetutils.ResourceVersionExpectations.Delete(updateRevision)
+	}
 	for _, pod := range filteredPods {
+		clonesetutils.ResourceVersionExpectations.Observe(pod)
 		if isSatisfied, unsatisfiedDuration := clonesetutils.ResourceVersionExpectations.IsSatisfied(pod); !isSatisfied {
 			if unsatisfiedDuration >= expectations.ExpectationTimeout {
 				klog.Warningf("Expectation unsatisfied overtime for %v, wait for pod %v updating, timeout=%v", request.String(), pod.Name, unsatisfiedDuration)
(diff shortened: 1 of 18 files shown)


pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -536,3 +576,132 @@ func (r *ReconcileCloneSet) claimPods(instance *appsv1alpha1.CloneSet, pods []*v
...
+
+	// ignore if update type is ReCreate
+	if cs.Spec.UpdateStrategy.Type == appsv1alpha1.RecreateCloneSetUpdateStrategyType {
+		klog.V(4).Infof("CloneSet %s/%s skipped to create ImagePullJob for update type is %s",
+			cs.Namespace, cs.Name, cs.Spec.UpdateStrategy.Type)
+		return r.patchControllerRevisionLabels(updateRevision, appsv1alpha1.ImagePreDownloadIgnoredKey, "true")
+	}
+
+	// ignore if replicas <= 3
+	if *cs.Spec.Replicas <= 3 {
```
Fei-Guo, line 595: Use constant variable instead of magic number?

Fei-Guo (commented):
Overall LG. Small Nit. 

One question: should we consider updating  imagePullJob results in the revision? It is still possible a pull job was created but failed to completion. Next time, if the revision is picked again, no pulljob will be created anymore.

pkg/controller/cloneset/cloneset_controller.go
FillZpp: done

FillZpp:
> One question: should we consider updating imagePullJob results in the revision? It is still possible a pull job was created but failed to completion. Next time, if the revision is picked again, no pulljob will be created anymore.

I think there is no need to create ImagePullJobs for one revision multiple times:

1. If the previous job succeeded, the new images should already be on those Nodes.
2. If the previous job totally failed, it means the new image has something wrong, such as it has a non-existing tag.
3. If the previous job partially failed, it is caused by several Nodes with load/network/... problems. It is not necessary to recreate the whole job for these abnormal Nodes, even if a few of them might have returned normal.

Fei-Guo (commented):
LGTM

pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -536,3 +577,132 @@ func (r *ReconcileCloneSet) claimPods(instance *appsv1alpha1.CloneSet, pods []*v
...
+		}
+	}
+
+	labelMap := make(map[string]string)
+	for k, v := range cs.Spec.Template.Labels {
+		labelMap[k] = v
+	}
+	labelMap[history.ControllerRevisionHashLabel] = updateRevision.Labels[history.ControllerRevisionHashLabel]
+
+	for name, image := range spec.ContainerImages {
```
furykerry, line 646: can init containers be supported too ?

pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -278,6 +305,20 @@ func (r *ReconcileCloneSet) doReconcile(request reconcile.Request) (res reconcil
...
+	if !isPreDownloadDisabled {
+		if currentRevision.Name != updateRevision.Name {
+			// pre-download images for new revision
+			if err := r.createImagePullJobsForInPlaceUpdate(instance, currentRevision, updateRevision); err != nil {
+				klog.Errorf("Failed to create ImagePullJobs for %s: %v", request, err)
+			}
+		} else {
+			// delete ImagePullJobs if revisions have been consistent
+			if err := r.deleteImagePullJobs(instance); err != nil {
+				klog.Errorf("Failed to delete imagepulljobs for %s: %v", request, err)
```
furykerry, line 317: what if the cloneset rollout partially, and rollback, in such case, new imagejob will not be recycled

pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -536,3 +577,132 @@ func (r *ReconcileCloneSet) claimPods(instance *appsv1alpha1.CloneSet, pods []*v
...
+
+	// ignore if update type is ReCreate
+	if cs.Spec.UpdateStrategy.Type == appsv1alpha1.RecreateCloneSetUpdateStrategyType {
+		klog.V(4).Infof("CloneSet %s/%s skipped to create ImagePullJob for update type is %s",
+			cs.Namespace, cs.Name, cs.Spec.UpdateStrategy.Type)
+		return r.patchControllerRevisionLabels(updateRevision, appsv1alpha1.ImagePreDownloadIgnoredKey, "true")
+	}
+
+	// ignore if replicas <= minimumReplicasToPreDownloadImage
+	if *cs.Spec.Replicas <= minimumReplicasToPreDownloadImage {
```
furykerry, line 596: is it necessary enforce minimumReplicas? can we just compare spec.replicas, partition and maxUnavailable, e.g. abort pre-download only if  paritition is zero and replicas==maxUnavailable

pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -536,3 +577,132 @@ func (r *ReconcileCloneSet) claimPods(instance *appsv1alpha1.CloneSet, pods []*v
...
+		}
+
+		err := r.Create(context.TODO(), job)
+		if err != nil {
+			if !errors.IsAlreadyExists(err) {
+				klog.Errorf("CloneSet %s/%s failed to create ImagePullJob %s: %v", cs.Namespace, cs.Name, job.Name, err)
+			}
+			continue
+		}
+		klog.V(3).Infof("CloneSet %s/%s created ImagePullJob %s for image: %s", cs.Namespace, cs.Name, job.Name, image)
```
furykerry, line 674: shall we emit an event as cloneset about the imagepulljobs here ?

pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -536,3 +577,132 @@ func (r *ReconcileCloneSet) claimPods(instance *appsv1alpha1.CloneSet, pods []*v
...
+	opts := &inplaceupdate.UpdateOptions{}
+	inplaceupdate.SetOptionsDefaults(opts)
+	spec := opts.CalculateSpec(currentRevision, updateRevision, opts)
+	if spec == nil {
+		klog.V(4).Infof("CloneSet %s/%s skipped to create ImagePullJob for %s -> %s can not update in-place",
+			cs.Namespace, cs.Name, currentRevision.Name, updateRevision.Name)
+		return r.patchControllerRevisionLabels(updateRevision, appsv1alpha1.ImagePreDownloadIgnoredKey, "true")
+	}
+
+	// start to create jobs
```
furykerry, line 612: how about extract the create jobs logic into a separate utility function, so that the pre-download for other workload such as adv statefulset can reuse the logic

furykerry (changes requested):

pkg/controller/cloneset/cloneset_controller.go
FillZpp: we have not supported to in-place update init containers

pkg/controller/cloneset/cloneset_controller.go
FillZpp: Even if the cloneset rollback, the imagepulljob will still download image on all nodes. We have not to create it again.

pkg/controller/cloneset/cloneset_controller.go
FillZpp: i think for most cloneset with one or two replicas, predownload image could hardly help them accelerate updating

pkg/controller/cloneset/cloneset_controller.go
FillZpp: done

pkg/controller/cloneset/cloneset_controller.go
FillZpp: i have done my best to extract the logic into `CreateJobForWorkload`

pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -536,3 +577,95 @@ func (r *ReconcileCloneSet) claimPods(instance *appsv1alpha1.CloneSet, pods []*v
...
+	}
+
+	// ignore if all Pods update in one batch
+	var partition, maxUnavailable int
+	if cs.Spec.UpdateStrategy.Partition != nil {
+		partition, _ = intstrutil.GetValueFromIntOrPercent(cs.Spec.UpdateStrategy.Partition, int(*cs.Spec.Replicas), true)
+	}
+	maxUnavailable, _ = intstrutil.GetValueFromIntOrPercent(
+		intstrutil.ValueOrDefault(cs.Spec.UpdateStrategy.MaxUnavailable, intstrutil.FromString(appsv1alpha1.DefaultCloneSetMaxUnavailable)), int(*cs.Spec.Replicas), false)
+	if partition == 0 && maxUnavailable >= int(*cs.Spec.Replicas) {
```
furykerry, line 609: we should also skip image downloading if 
1. cs.spec.replicas == 0

pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -536,3 +577,95 @@ func (r *ReconcileCloneSet) claimPods(instance *appsv1alpha1.CloneSet, pods []*v
...
+
+	// ignore if update type is ReCreate
+	if cs.Spec.UpdateStrategy.Type == appsv1alpha1.RecreateCloneSetUpdateStrategyType {
+		klog.V(4).Infof("CloneSet %s/%s skipped to create ImagePullJob for update type is %s",
+			cs.Namespace, cs.Name, cs.Spec.UpdateStrategy.Type)
+		return r.patchControllerRevisionLabels(updateRevision, appsv1alpha1.ImagePreDownloadIgnoredKey, "true")
+	}
+
+	// ignore if replicas <= minimumReplicasToPreDownloadImage
+	if *cs.Spec.Replicas <= minimumReplicasToPreDownloadImage {
```
furykerry, line 596: can we just drop the minimumReplicas check ?

furykerry (changes requested):

pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -536,3 +577,95 @@ func (r *ReconcileCloneSet) claimPods(instance *appsv1alpha1.CloneSet, pods []*v
...
+	selector := cs.Spec.Selector.DeepCopy()
+	selector.MatchExpressions = append(selector.MatchExpressions, metav1.LabelSelectorRequirement{
+		Key:      apps.ControllerRevisionHashLabelKey,
+		Operator: metav1.LabelSelectorOpNotIn,
+		Values:   []string{updateRevision.Name, updateRevision.Labels[history.ControllerRevisionHashLabel]},
+	})
+
+	labelMap := make(map[string]string)
+	for k, v := range cs.Spec.Template.Labels {
+		labelMap[k] = v
```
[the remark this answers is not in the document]
FillZpp: As cloneset is the  job's owner, we have the convention that all resources owned by cloneset have to match the selector of cloneset, such as pod, pvc and controllerrevision. So we had better put the labels into jobs.

pkg/controller/cloneset/cloneset_controller.go
FillZpp: It contains in the minimumReplicas check.

pkg/controller/cloneset/cloneset_controller.go
FillZpp: I love this check, it may well avoid many nonessential jobs.

furykerry (approved):

furykerry (approved):

PR #592, opened 2021-04-13
remove redundant code

hantmac:

Remove redundant code, because the pods those can't through [predicate funcs](https://github.com/openkruise/kruise/blob/786beb4c0fd5fa6f84cf78a4255d514e0c34ed59/pkg/controller/podreadiness/pod_readiness_controller.go#L67), can not be reconciled, so in `reconcile` we have no need to check it twice.


FillZpp (changes requested):
it is needed in conflict retry, for other controller could also update this readiness condition

hantmac:
> it is needed in conflict retry, for other controller could also update this readiness condition

Got it, If there is a time difference between entering the queue and the actual processing,  it is reasonable to judge again.

Issue #597, opened 2021-04-14
[BUG] create sidecarSet with volumes.downwardAPI failed


zmberg:
create sidecarSet with volumes.downwardAPI failed.
sidecarSet.yaml
```
# sidecarset.yaml
apiVersion: apps.kruise.io/v1alpha1
kind: SidecarSet
metadata:
  name: test-sidecarset
spec:
  selector:
    matchLabels:
      app: nginx
  updateStrategy:
    type: RollingUpdate
    maxUnavailable: 1
  containers:
  - name: sidecar1
    image: centos:6.7
    command: ["sleep", "999d"] # do nothing at all
    volumeMounts:
    - name: podinfo
      mountPath: /etc/podinfo
  volumes:
  - name: podinfo
    downwardAPI:
      items:
      - path: "labels"
        fieldRef:
          fieldPath: metadata.labels
      - path: "annotations"
        fieldRef:
          fieldPath: metadata.annotations
```
kubectl create -f sidecarset.yaml
Error from server: error when creating "sidecarset.yaml": admission webhook "vsidecarset.kb.io" denied the request: [spec.volumes[0].downwardAPI.fieldRef.apiVersion: Required value, spec.containers[0].volumeMounts[0].name: Not found: "podinfo"]

zmberg:
This issue is due to the lack of default setting for sidecarset.spec.volumes, resulting in a null of spec.volumes[0].downwardAPI.fieldRef.apiVersion

Fei-Guo:
This should be fixed. Please reopen if the problem still occurs.

PR #598, opened 2021-04-14
fix sidecarSet.spec.volumes are not assigned by default #597

zmberg:
fix sidecarSet.spec.volumes are not assigned by default


Fei-Guo (approved):
LGTM

PR #603, opened 2021-04-16
fix pod.annotations[sidecarset-hash] compatibility #602

zmberg:

### DIFF
--- a/pkg/webhook/pod/mutating/sidecarset.go
+++ b/pkg/webhook/pod/mutating/sidecarset.go
@@ -176,14 +176,34 @@ func buildSidecars(isUpdated bool, pod *corev1.Pod, oldPod *corev1.Pod, matchedS
 	// parse sidecar hash in pod annotations
 	if oldHashStr := pod.Annotations[sidecarcontrol.SidecarSetHashAnnotation]; len(oldHashStr) > 0 {
 		if err := json.Unmarshal([]byte(oldHashStr), &sidecarSetHash); err != nil {
-			return nil, nil, nil, nil,
-				fmt.Errorf("invalid %s value %v, unmarshal failed: %v", sidecarcontrol.SidecarSetHashAnnotation, oldHashStr, err)
+			// to be compatible with older sidecarSet hash struct, map[string]string
+			olderSidecarSetHash := make(map[string]string)
+			if err = json.Unmarshal([]byte(oldHashStr), &olderSidecarSetHash); err != nil {
+				return nil, nil, nil, nil,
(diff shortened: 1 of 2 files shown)


pkg/webhook/pod/mutating/sidecarset.go
```diff
@@ -176,14 +176,34 @@ func buildSidecars(isUpdated bool, pod *corev1.Pod, oldPod *corev1.Pod, matchedS
 	// parse sidecar hash in pod annotations
 	if oldHashStr := pod.Annotations[sidecarcontrol.SidecarSetHashAnnotation]; len(oldHashStr) > 0 {
 		if err := json.Unmarshal([]byte(oldHashStr), &sidecarSetHash); err != nil {
-			return nil, nil, nil, nil,
-				fmt.Errorf("invalid %s value %v, unmarshal failed: %v", sidecarcontrol.SidecarSetHashAnnotation, oldHashStr, err)
+			// to be compatible with older sidecarSet hash struct, map[string]string
+			olderSidecarSetHash := make(map[string]string)
+			if err = json.Unmarshal([]byte(pod.Annotations[sidecarcontrol.SidecarSetHashAnnotation]), &olderSidecarSetHash); err != nil {
```
furykerry, line 181: we can use []byte(oldHashStr) directly here

hantmac (commented):
I think the sidecarSet hash value should be update with it's spec, use old hash value for new sidecarSet is not reasonable.

zmberg:
> I think the sidecarSet hash value should be update with it's spec, use old hash value for new sidecarSet is not reasonable.

This is a compatibility issue and no changes have been made to sidecarSet.Spec. The format of the sidecarSet Hash stored in pod.annotations in 0.7.0 is: map[sidecarSet.Name] = sidecarSet Hash. After version 0.8.0 the structure is adjusted to: map[sidecarSet.Name] = sidecarControl.SidecarSetUpgradeSpec. So it needs to make the data structure compatible in this place.

hantmac:
> > I think the sidecarSet hash value should be update with it's spec, use old hash value for new sidecarSet is not reasonable.
> 
> This is a compatibility issue and no changes have been made to sidecarSet.Spec. The format of the sidecarSet Hash stored in pod.annotations in 0.7.0 is: map[sidecarSet.Name] = sidecarSet Hash. After version 0.8.0 the structure is adjusted to: map[sidecarSet.Name] = sidecarControl.SidecarSetUpgradeSpec. So it needs to make the data structure compatible in this place.

Thanks for your reply, get it.

furykerry (approved):

PR #610, opened 2021-04-21
Fix inaccurate resourceVersion expectation for ImagePullJob

FillZpp:

Fix inaccurate resourceVersion expectation for ImagePullJob.

The previous `resourceVersionExpectations` observes nodeimage in event handler, which might cause the satisfied check to be inaccurate in reconcile, and the active pulling count may be bigger than `parallelism`.

### DIFF
--- a/pkg/controller/imagepulljob/imagepulljob_controller.go
+++ b/pkg/controller/imagepulljob/imagepulljob_controller.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"reflect"
 	"sort"
 	"time"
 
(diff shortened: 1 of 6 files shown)


pkg/controller/imagepulljob/imagepulljob_controller.go
```diff
@@ -151,6 +150,17 @@ func (r *ReconcileImagePullJob) Reconcile(request reconcile.Request) (res reconc
 		return reconcile.Result{}, err
 	}
 
+	// If resourceVersion expectations have not satisfied yet, just skip this reconcile
+	resourceVersionExpectations.Observe(job)
```
hantmac, line 154: May I ask a question?
Why some workloads such as cloneSet, ImagePullJob need check whether their resourceVersion expectations  have been satisfied and some workloads have no need to check?

FillZpp: Controllers like CloneSet, ImagePullJob have reentrant risk.

For example, a imagepulljob with `parallelism=1` had added the pulling task into one nodeimage in previous reconcile, if there is no `resourceVersionExpectations`, it might add the task into another nodeimage in next immediate reconcile, because the modification of the previous nodeimage has not been watched and synced into informer.

furykerry (approved):

PR #613, opened 2021-04-27
Add proposal of UnitedTopology, a bypass scheme for elasticity and multi-domain management

FillZpp:

Add proposal of UnitedTopology, a bypass scheme for elasticity and multi-domain management.

UnitedTopology supports configuration and management for elastic and multi-domain applications,
even more powerful and flexible than UnitedDeployment in some ways.

Not like the workload type, UnitedTopology can be used with those stateless workloads, such as CloneSet, Deployment and ReplicaSet.

So users needn't learn the usage of a new workload and how to deploy and update their application by it.
They only have to create UnitedTopology CRs related to their existing workloads.

### DIFF
--- a/docs/proposals/20210427-unitedtopology.md
+++ b/docs/proposals/20210427-unitedtopology.md
@@ -1,0 +1,256 @@
+---
+title: UnitedTopology
+authors:
+- "@FillZpp"
+reviewers:
+- "@Fei-Guo"
+- "@furykerry"
+creation-date: 2021-04-27
+last-updated: 2021-04-27
+status: implementable
+---
+
+# UnitedTopology
+
+## Table of Contents
+
+A table of contents is helpful for quickly jumping to sections of a proposal and for highlighting
+any additional information provided beyond the standard proposal template.
+[Tools for generating](https://github.com/ekalinin/github-markdown-toc) a table of contents from markdown are available.
+
+
+## Motivation
+
+Define a CRD named UnitedTopology, which can support configuration and management for elastic and multi-domain applications,
+even more powerful and flexible than [UnitedDeployment](https://openkruise.io/en-us/docs/uniteddeployment.html) in some ways.
+
+Not like the workload type, UnitedTopology can be used with those stateless workloads, such as CloneSet, Deployment and ReplicaSet.
+
+So users needn't learn the usage of a new workload and how to deploy and update their application by it.
+They only have to create UnitedTopology CRs related to their existing workloads. Everything is just great!
+
+## Proposal
+
+UnitedTopology definition looks like this:
+
+```yaml
+apiVersion: apps.kruise.io/v1alpha1
+kind: UnitedTopology
+metadata:
+  namespace: ...
+  name: ...
+  # ...
+spec:
+  targetRef:
+    apiVersion: apps/v1 | apps.kruise.io/v1alpha1
+    kind: Deployment | CloneSet
+    name: workload-xxx
+  subsets:
+  - name: subset-a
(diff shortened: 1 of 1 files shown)


Fei-Guo (commented):
I understand the intention and I think it is a good idea. 

UnitedTopology does not sound clear for what the controller will do. It is essentially a workload Pod distribution manager that influences the scheduling results. 

So I would suggest to change CRD name to a more explicit one, such as WorkloadDistribution/PodDistribution/TopologyTrack etc.

docs/proposals/20210427-unitedtopology.md
```diff
@@ -0,0 +1,256 @@
...
+  - `name`: name of this subset
+  - `nodeSelectorTerm`: the additional node selector of this subset, it will be injected into each term of `pod.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms`
+  - `tolerations`: the additional tolerations of this subset, it will be injected into `pod.spec.tolerations`
+  - `patch`: the specified strategic patch body of this subset, it will be patched to Pods that belong to this subset
+  - `maxReplicas`: the maximum number of Pods in this subset, it can be an absolute number or a percentage string.
+- `scheduleStrategy`: strategies for schedule
+  - `type`: can be `Fixed` or `Adaptive`
+    - `Fixed` means always choose the former subset for Pods unless the subset's maxReplicas has been satisfied
+    - `Adaptive` means choose the right subset by sequence of subsets and also enough resources
+  - `adaptive`:
```
Fei-Guo, line 98: The `adaptive` strategy and the `adaptive` type look duplicated concepts.

I would suggest to remove the `adaptive` strategy in the first version. It seems that 
the  `rescheduleCriticalSeconds` can be hard coded in the beginning. 
The `disableSimulationSchedule` needs explanation which seems to be optional in most cases.

Fei-Guo:
One more thing - it will great to compare this proposal with K8s PodTopologySpread API so that people can understand why we want to enhance the upstream scheduling API.

docs/proposals/20210427-unitedtopology.md
```diff
@@ -0,0 +1,256 @@
...
+      pod-xxx: "2021-04-27T08:34:11Z"
+    deleting:
+      pod-yyy: "2021-04-27T08:34:11Z"
+  - name: subset-b
+    missingReschedule: 1
+```
+
+**spec**:
+- `targetRef`: reference to a specific workload, it supports `Deployment` in `apps` group and `CloneSet` in `apps.kruise.io` group
+- `subsets`: list of subsets which defines different topologies
```
furykerry, line 88: comments about the order of topology which may influence creation and deletion order

docs/proposals/20210427-unitedtopology.md
```diff
@@ -0,0 +1,256 @@
...
+5. Pods that have deployed on the same Node
+6. Pods with higher restart count
+7. Pods that have newer creationTimestamp
+8. ...
+
+CloneSet will also support this annotation to implement deletion priority.
+
+And then, UnitedTopology will define three levels for the Pods belongs to subsets:
+
+1. `Guaranteed` for Pods that in the `maxReplicas` of its subset, deletion-cost will be set bigger than `0`
```
furykerry, line 161: consider the case that deployments with idc and cloud instance, we may prefer to delete cloud instance even without maxReplicas

furykerry (changes requested):

FillZpp:
There will be a new PR for this proposal.

PR #617, opened 2021-05-12
Optimize manage phase of daemonset controller

jetmuffin:

Optimize manage function when synchronize daemonsets.  In manage phase, daemonset controller figures out which nodes should run daemonset pods but not yet running and  which should not run but current running one. And then upgrades limited number of replicas according to partition number configured in RollingUpdate field.

However, current implementation does not work when is-first-deploy label is added to the daemonset. We found that when the label `is-first-deploy` is set to true, daemonset controller creates all new replicas in a batch no matter what the partition is; And if this label is set to false, daemonset controller deletes all old replicas once we change the node affinity or toleration in pod template spec.

Some conformance tests for daemonset controller has been added in this PR.

### DIFF
--- a/pkg/controller/daemonset/daemonset_controller.go
+++ b/pkg/controller/daemonset/daemonset_controller.go
@@ -93,8 +93,6 @@ const (
 	// StatusUpdateRetries limits the number of retries if sending a status update to API server fails.
 	StatusUpdateRetries = 1
 
-	IsFirstDeployedFlag = "daemonset.kruise.io/is-first-deployed"
-
 	IsIgnoreNotReady     = "daemonset.kruise.io/ignore-not-ready"
 	IsIgnoreUnscheduable = "daemonset.kruise.io/ignore-unscheduable"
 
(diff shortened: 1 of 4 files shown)


test/e2e/framework/daemonset_util.go
```diff
@@ -0,0 +1,197 @@
...
+			}
+			if podutil.IsPodAvailable(&pod, ds.Spec.MinReadySeconds, metav1.Now()) {
+				nodesToPodCount[pod.Spec.NodeName] += 1
+			}
+		}
+		Logf("Number of nodes with available pods: %d", len(nodesToPodCount))
+
+		// Ensure that exactly 1 pod is running on all nodes in nodeNames.
+		for _, nodeName := range nodeNames {
+			if nodesToPodCount[nodeName] != 1 {
```
hantmac, line 155: Should be `if nodesToPodCount[nodeName] > 1 ` ?

pkg/controller/daemonset/daemonset_controller.go
```diff
@@ -596,24 +596,29 @@ func (dsc *ReconcileDaemonSet) manage(ds *appsv1alpha1.DaemonSet, hash string) (
 		return delay, fmt.Errorf("couldn't get nodeList %v", err)
 	}
 
-	var nodesNeedingDaemonPods, podsToDelete []string
+	var nodesNeedingDaemonPods, nodesDesireScheduled, podsToDelete []string
```
FillZpp, line 599: can this `nodesDesireScheduled` be an integer?

FillZpp (commented):
can we remove the `daemonset.kruise.io/is-first-deployed` label definition?

test/e2e/framework/daemonset_util.go
jetmuffin: Okay, I'll fix this.

pkg/controller/daemonset/daemonset_controller.go
jetmuffin: The last two comments have been fixed in the most recent commit :)

FillZpp (approved):
LGTM

PR #620, opened 2021-05-14
add  .Values.crds.namespace for helm chart

hantmac:
Add  .Values.crds.namespace for helm chart and fix https://github.com/openkruise/kruise/issues/619

Fixes https://github.com/openkruise/kruise/issues/619

Modify `.Values.crds.namespace` use customized value to install kruise successfuly.

### DIFF
--- a/charts/kruise/v0.8.1/values.yaml
+++ b/charts/kruise/v0.8.1/values.yaml
@@ -3,6 +3,10 @@
 crds:
   managed: true
 
+# namespace for kruise installation
+installation:
+  namespace: kruise-system
+
 featureGates: ""
 
 manager:
--- a/charts/kruise/v0.8.1/templates/apps.kruise.io_statefulsets.yaml
+++ b/charts/kruise/v0.8.1/templates/apps.kruise.io_statefulsets.yaml
@@ -49,7 +49,7 @@ spec:
       caBundle: Cg==
       service:
         name: kruise-webhook-service
-        namespace: kruise-system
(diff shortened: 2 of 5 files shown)


charts/kruise/v0.8.1/values.yaml
```diff
@@ -2,6 +2,7 @@
 
 crds:
   managed: true
+  namespace: kruise-system
```
FillZpp, line 5: `namespace` is inappropriate to be defined in `crds`, because CRDs have no namespace. It is for the other resources like pods, service, configmap.

Maybe a new `installation`:

```yaml
installation:
  namespace: kruise-system
```

FillZpp (commented):
The namespaces in `rback_role.yaml` also need to update

hantmac:
@FillZpp Do we need to separate the chart of CRD from other resources?

hantmac:
> The namespaces in `rback_role.yaml` also need to update

The namespaces in `/charts/kruise/v0.8.1/templates/rbac_role.yaml` has been updated

yxxhero:
I think  you shoud rebase your code, The commit info is very disorderly. My english is not good. @hantmac

FillZpp:
> @FillZpp Do we need to separate the chart of CRD from other resources?

You mean move the CRD yamls into `crds/` directory? I used to try this in `v0.7.0` chart, but I found the CRDs will be deleted during `helm upgrade` from an old version. So I had to move them back.

FillZpp (approved):
LGTM

Issue #635, opened 2021-06-10
[BUG] inplace update process will be hang up when the image id is not changed


wdwangcumt:
**What happened**:

First, create a cloneset `sample` by using the following yaml content:
```
apiVersion: apps.kruise.io/v1alpha1
kind: CloneSet
metadata:
  labels:
    app: sample
  name: sample
spec:
  replicas: 3
  selector:
    matchLabels:
      app: sample
  template:
    metadata:
      labels:
        app: sample
    spec:
      containers:
      - image: nginx:1.18
        name: nginx
  updateStrategy:
    partition: 3  --- !!! equal to the replicas !!!
    type: InPlaceOnly
```
apply it, watch the pods by command `kubectl get pod  -L controller-revision-hash -L lifecycle.apps.kruise.io/state -w`, it will work well.

Then, we edit the cloneset `sample`, modify the following fields:
```
spec:
  replicas: 3
  ...
  template:
    metadata:
      labels:
        app: sample
    spec:
      containers:
      - image: nginx-copy:1.18   --- !!! change it to a new image tag, but with the same image id with the old !!!
        ...
  updateStrategy:
    ...
    partition: 1 --- !!! change it from 3 to 1 !!!
    ...
```
Watch pods, we will see the problem: 
1. **Only one pod's revision will changed, while two pods' revision changed is expected by the meaning of the `updateStrategy.partition`, and the  update process will hung up.**
```
# Outputs below are the results after 5 minutes

% kubectl get pod  -L controller-revision-hash -L lifecycle.apps.kruise.io/state -w
NAME           READY   STATUS    RESTARTS   AGE   CONTROLLER-REVISION-HASH   STATE
sample-7jrpp   1/1     Running   1          5m   sample-648847746c          Updating   --- !!! continues to be Updating !!!
sample-cgsr6   1/1     Running   0          5m   sample-7f4d7465f8          Normal
sample-vzbpx   1/1     Running   0          5m   sample-7f4d7465f8          Normal
```

2. The label `lifecycle.apps.kruise.io/state`'s value continues to be `Updating` for the pod which revision changed.
3. The status of cloneset `sample` is:
```
status:
  availableReplicas: 2
  collisionCount: 0
  currentRevision: sample-7f4d7465f8
  labelSelector: app=sample
  observedGeneration: 2
  readyReplicas: 2
  replicas: 3
  updateRevision: sample-648847746c
  updatedReadyReplicas: 0  --- !!! continues to be 0 !!!
[...]

PR #636, opened 2021-06-10
BugFix: inplace update process will be hang up when the image id is not changed

wdwangcumt:

This PR fixed the bug described at https://github.com/openkruise/kruise/issues/635

fixes https://github.com/openkruise/kruise/issues/635

Verify as described in the issue https://github.com/openkruise/kruise/issues/635


FillZpp:
@wdwangcumt You can not judge the update is finished if the imageID not changed. Actually, it usually means kubelet has not recreated the container with new image yet.

PR #648, opened 2021-06-15
Advanced StatefulSet supports scaling up with rate limit

### LINKED ISSUE #645: [feature request] StatefulSet should support scaling with rate limit
StatefulSet should support scaling with rate limit.

**Why is this needed**:
Currently, a StatefulSet with `podManagementPolicy=OrderedReady` will create Pods one by one, or with `podManagementPolicy=Parallel` will create all Pods at once.

It should support a rate limiter for scaling with `podManagementPolicy=Parallel`, as the `maxUnavailable` for updating.

FillZpp:

Advanced StatefulSet supports scaling up with rate limit

fixes #645


Fei-Guo:
The change looks ok. I still don't quite understand what problem this change will fix, even in the reported issue #645. What is the problem of creating new Pods too quickly?

FillZpp:
> What is the problem of creating new Pods too quickly?

1. Kube-apiserver will get thousands of Pods creation at the same time, , as well as scheduler will suffer from a long queue.
2. If the asts has an incorrect affinity, all Pods will stay in Pending and scheduler has to reschedule them periodically.
3. If the asts has an incorrect image or configuration, all Pods scheduled can not start up, which will cause lots of resources wasting.

CloneSet may also need this, we are considering to support it later.

pkg/controller/statefulset/stateful_set_utils.go
```diff
@@ -506,3 +506,12 @@ func filterOutCondition(conditions []apps.StatefulSetCondition, condType apps.St
 func getStatefulSetKey(o metav1.Object) string {
 	return o.GetNamespace() + "/" + o.GetName()
 }
+
+func decreaseAndThrottleForMaxUnavailable(maxUnavailable *int) bool {
```
furykerry, line 510: i cannot understand what the function returns through the function interface. Consider renaming to decreaseAndCheckMaxUnavailable

furykerry (approved):

PR #650, opened 2021-06-22
podUnavailableBudget control 

zmberg:

podUnavailableBudget public method called by webhook and controller

### DIFF
--- a/pkg/control/pubcontrol/api.go
+++ b/pkg/control/pubcontrol/api.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2021 The Kruise Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package pubcontrol
+
+import (
+	policyv1alpha1 "github.com/openkruise/kruise/apis/policy/v1alpha1"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+type PubControl interface {
+	// Common
+	// get PodUnavailableBudget
+	GetPodUnavailableBudget() *policyv1alpha1.PodUnavailableBudget
+	// IsPodReady indicates whether pod is fully ready
+	// 1. pod.Status.Phase == v1.PodRunning
+	// 2. pod.condition PodReady == true
+	IsPodReady(pod *corev1.Pod) bool
+	// IsPodStateConsistent indicates whether pod.spec and pod.status are consistent after updating containers
+	IsPodStateConsistent(pod *corev1.Pod) bool
+
+	// webhook
+	// determine if this change to pod might cause unavailability
+	IsPodUnavailableChanged(oldPod, newPod *corev1.Pod) bool
+}
+
+func NewPubControl(pub *policyv1alpha1.PodUnavailableBudget) PubControl {
+	return &commonControl{PodUnavailableBudget: pub}
+}
--- a/pkg/control/pubcontrol/util.go
+++ b/pkg/control/pubcontrol/util.go
@@ -1,0 +1,106 @@
+/*
+Copyright 2021 The Kruise Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package pubcontrol
+
+import (
+	"context"
+
+	policyv1alpha1 "github.com/openkruise/kruise/apis/policy/v1alpha1"
+	"github.com/openkruise/kruise/pkg/util"
+	"github.com/openkruise/kruise/pkg/util/controllerfinder"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func GetPodUnavailableBudgetForPod(kClient client.Client, finders *controllerfinder.ControllerFinder, pod *corev1.Pod) (*policyv1alpha1.PodUnavailableBudget, error) {
+	var err error
+	if len(pod.Labels) == 0 {
+		return nil, nil
+	}
+
+	pubList := &policyv1alpha1.PodUnavailableBudgetList{}
+	if err = kClient.List(context.TODO(), pubList, &client.ListOptions{Namespace: pod.Namespace}); err != nil {
+		return nil, err
+	}
+
+	var matchedPubs []policyv1alpha1.PodUnavailableBudget
+	for _, pub := range pubList.Items {
+		// if targetReference isn't nil, priority to take effect
+		if pub.Spec.TargetReference != nil {
+			targetRef := pub.Spec.TargetReference
+			// check whether APIVersion, Kind, Name is equal
+			ref := metav1.GetControllerOf(pod)
+			if ref == nil {
+				continue
+			}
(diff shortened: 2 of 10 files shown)


pkg/control/pubcontrol/util.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/openkruise/kruise/pkg/util/controllerfinder"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func GetPodPodUnavailableBudget(kClient client.Client, finders *controllerfinder.ControllerFinder, pod *corev1.Pod) (*policyv1alpha1.PodUnavailableBudget, error) {
```
Fei-Guo, line 32: GetPodPodUnavailableBudget -> GetPodUnavailableBudgetForPod

pkg/control/pubcontrol/validate_pub.go
```diff
@@ -0,0 +1,160 @@
...
+		klog.Errorf("pod(%s.%s) operation(%s) failed: %s", pod.Namespace, pod.Name, operation, err.Error())
+		return false, err.Error(), nil
+	}
+
+	klog.V(3).Infof("admit pod(%s.%s) operation(%s) for pub(%s.%s)", pod.Namespace, pod.Name, operation, pub.Namespace, pub.Name)
+	return true, "", nil
+}
+
+func checkAndDecrement(podName string, pub *policyv1alpha1.PodUnavailableBudget, operation Operation) error {
+	if pub.Status.UnavailableAllowed < 0 {
```
Fei-Guo, line 120: Question with regard to UnavailableAllowed:

1) How is the value updated? or initialized in the first place? 
2) Why do we need to persist the value in the status instead of calculating based on MaxUnavailable and MinAvailable in the Spec online?

zmberg: The amount of code submitted this time is relatively large, so the core code is divided into the following sections for submission.
1. Pub control: webhook and controller some public methods, as well as interfaces used to decouple the group and community implementation
2. webhook: used to intercept all pod requests and determine whether to allow the request according to UnavailableAllowed
3. controller: Calculate UnavailableAllowed based on the real situation of online pods (whether they are ready), etc. UnavailableAllowed is a complex calculation process and is not suitable to be calculated in webhook.
PodUnavailableBudgetValidatePod method should be a logical part of the webhook, but because it is intended to be called directly by in-place upgrade logic such as cloneset to achieve in-place upgrade protection, so it is placed in pub control

pkg/control/pubcontrol/api.go
```diff
@@ -0,0 +1,43 @@
...
+type PubControl interface {
+	// Common
+	// get PodUnavailableBudget
+	GetPodUnavailableBudget() *policyv1alpha1.PodUnavailableBudget
+	// IsPodReady indicates whether pod is fully ready
+	// 1. pod.Status.Phase == v1.PodRunning
+	// 2. pod.condition PodReady == true
+	IsPodReady(pod *corev1.Pod) bool
+	// IsPodUpdatedConsistent indicates whether pod.spec and pod.status are consistent after updating containers
+	IsPodUpdatedConsistent(pod *corev1.Pod) bool
```
Fei-Guo, line 34: Sorry, I did not suggest a good name in my previous review. Since this is the API definition, please forgive my paranoid about naming. 

IsPodUpdatedConsistent can be IsPodStateConsistent or IsPodInDesiredState.

Fei-Guo (commented):
LGTM. One nit.

Fei-Guo:
LGTM.

Please fix lint error.

pkg/control/pubcontrol/util.go
```diff
@@ -0,0 +1,105 @@
...
+				return nil, err
+			}
+
+			// belongs the same workload
+			if isReferenceEqual(podRef, pubRef) {
+				matchedPubs = append(matchedPubs, pub)
+			}
+		} else {
+			// This error is irreversible, so continue
+			labelSelector, err := metav1.LabelSelectorAsSelector(pub.Spec.Selector)
```
furykerry, line 70: if we validate the pub selector in webhook beforehand, we can call GetFastLabelSelector to speedup label selection

pkg/control/sidecarcontrol/sidecarset_control.go
```diff
@@ -199,7 +199,7 @@ func (c *commonControl) IsPodAvailabilityChanged(pod, oldPod *v1.Pod) bool {
 
 // isContainerInplaceUpdateCompleted checks whether imageID in container status has been changed since in-place update.
 // If the imageID in containerStatuses has not been changed, we assume that kubelet has not updated containers in Pod.
-func isSidecarContainerUpdateCompleted(pod *v1.Pod, sidecarSetName string, containers sets.String) bool {
+func IsSidecarContainerUpdateCompleted(pod *v1.Pod, sidecarSets, containers sets.String) bool {
```
furykerry, line 202: plz add ut for this function, it seems testable.

pkg/util/inplaceupdate/options.go
```diff
@@ -145,10 +145,10 @@ func defaultCalculateInPlaceUpdateSpec(oldRevision, newRevision *apps.Controller
 	return updateSpec
 }
 
-// defaultCheckInPlaceUpdateCompleted checks whether imageID in pod status has been changed since in-place update.
+// DefaultCheckInPlaceUpdateCompleted checks whether imageID in pod status has been changed since in-place update.
 // If the imageID in containerStatuses has not been changed, we assume that kubelet has not updated
 // containers in Pod.
-func defaultCheckInPlaceUpdateCompleted(pod *v1.Pod) error {
+func DefaultCheckInPlaceUpdateCompleted(pod *v1.Pod) error {
```
furykerry, line 151: plz add ut

pkg/control/pubcontrol/validate_pub.go
```diff
@@ -0,0 +1,160 @@
...
+		return errors.NewForbidden(policyv1alpha1.Resource("podunavailablebudget"), pub.Name, fmt.Errorf("pub unavailable allowed is negative"))
+	}
+	if len(pub.Status.DisruptedPods)+len(pub.Status.UnavailablePods) > MaxUnavailablePodSize {
+		return errors.NewForbidden(policyv1alpha1.Resource("podunavailablebudget"), pub.Name, fmt.Errorf("DisruptedPods and UnavailablePods map too big - too many unavailable not confirmed by PUB controller"))
+	}
+	if pub.Status.UnavailableAllowed == 0 {
+		err := errors.NewTooManyRequests("Cannot operate pod as it would violate the pod's unavailable budget.", 0)
```
furykerry, line 127: consider rename the error to NewForbidden

furykerry (changes requested):

pkg/control/pubcontrol/util.go
```diff
@@ -0,0 +1,105 @@
...
+			matchedPubs = append(matchedPubs, pub)
+		}
+	}
+
+	if len(matchedPubs) == 0 {
+		klog.V(6).Infof("could not find PodUnavailableBudget for pod %s in namespace %s with labels: %v", pod.Name, pod.Namespace, pod.Labels)
+		return nil, nil
+	}
+	if len(matchedPubs) > 1 {
+		klog.Warningf("Pod %q/%q matches multiple PodUnavailableBudgets. Choose %q arbitrarily.", pod.Namespace, pod.Name, matchedPubs[0].Name)
```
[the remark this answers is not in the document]
zmberg: The current design doc above is that a group of Pods can only belong to one PUB, and PDB is similarly implemented. And the validate code also does a check on the selector overlap.
However, similar to the scenario of pub1.selector=labels[key1]=val1, pub2.selector=labels[key2]=val2, it is possible that there are multiple pubs, and only one is selected for this scenario. The correct way of thinking would be to try to avoid this situation, but we don't have a particularly good solution at kruise yet.

pkg/util/inplaceupdate/options.go
zmberg: defaultCheckInPlaceUpdateCompleted already has some ut and they meet the requirements for pub

pkg/util/controllerfinder/controller_finder.go
```diff
@@ -0,0 +1,322 @@
...
+
+	// kruise
+	controllerKruiseKindCS = appsv1alpha1.SchemeGroupVersion.WithKind("CloneSet")
+	controllerKruiseKindSS = appsv1beta1.SchemeGroupVersion.WithKind("StatefulSet")
+)
+
+// getPodReplicaSet finds a replicaset which has no matching deployments.
+func (r *ControllerFinder) getPodReplicaSet(ref ControllerReference, namespace string) (*ScaleAndSelector, error) {
+	// This error is irreversible, so there is no need to return error
+	ok, _ := verifyGroupKind(ref, controllerKindRC.Kind, []string{controllerKindRC.Group})
```
FillZpp, line 110: The function is `getPodReplicaSet ` which should get ReplicaSet in`apps/v1`, why it try to get the deprecated ReplicationController?

PR #651, opened 2021-06-23
remove diff pods for managePreparingDelete

shiyan2016:
It has checked whether the pod is isSpecifiedDelete for the range of
podsInPreDelete in managePreparingDelete func.

Remove diff pods for managePreparingDelete since it has checked whether the pod is isSpecifiedDelete for the range of
podsInPreDelete in managePreparingDelete func.


FillZpp (approved):
LGTM

PR #653, opened 2021-06-25
fix inplace_update for daemonset and support inplace

Noisyes:
fix inlace_update bug and add test case. 
Before modification, ObserveUpdated function to refresh oldPods status, but should refresh newPods status to pass SatisfiedExpectations. 
and  ds.updateExp  should have own implementation not default.

NONE

compare container image for new pod and old pod, then compare old nodes and new nodes.
  add e2e daemonset test case


FillZpp:
/hold

Rebase and check this after #711 has been merged.

PR #654, opened 2021-06-28
Add proposal of WorkloadSpread, a bypass scheme for elasticity and multi-domain management 

BoltsLei:
Signed-off-by: BoltsLei  [Caofree286@gmail.com](Caofree286@gmail.com)

Add proposal of WorkloadSpread, a bypass scheme for elasticity and multi-domain management.

WorkloadSpread supports configuration and management for elastic and multi-domain applications,
even more powerful and flexible than UnitedDeployment in some ways.

Not like the workload type, WorkloadSpread can be used with those stateless workloads, such as CloneSet, Deployment and ReplicaSet.

So users needn't learn the usage of a new workload and how to deploy and update their application by it.
They only have to create WorkloadSpread CRs related to their existing workloads.

### DIFF
--- a/docs/proposals/20210628-workloadspread.md
+++ b/docs/proposals/20210628-workloadspread.md
@@ -7,7 +7,7 @@ reviewers:
   - "@furykerry"
   - "@FillZpp"
 creation-date: 2021-06-28
-last-updated: 2021-07-15
+last-updated: 2021-07-19
 status: implementable
 ---
 
@@ -81,7 +81,6 @@ spec:
       rescheduleCriticalSeconds: 30
 status:
   observedGeneration: 12
-  observedWorkloadGeneration: 3
   subsetStatuses:
   - name: subset-a
     missingReplicas: 2
@@ -249,118 +248,12 @@ If WorkloadSpread changes its subset maxReplicas, Pods will not be recreated, bu
 
 To change scheduleStrategy of WorkloadSpread will keep the deletion-cost annotation of Pod.
 
-### Consistency
-
-Webhook selects a suitable subset to inject pod by checking whether missingReplicas of the subset > 0. However, it may have two dangerous condition to lead the new created replicas of workload can't be scheduled into the suitable subset.
-
-1.The workload scales up before the WorkloadSpread controller updates the latest CR status. It can lead the subset's missingReplicas is not newest and cause the failure spread.
-
-For example:
-   ```yaml
-   CR:
-     metadata:
-       generation: 1
-     spec:
-       subsets:
-       - name: subset-a
-         maxReplicas: 100
-       - name: subset-b
-         maxReplicas: 100
-     status:
-       observedGeneration: 1
-       subsetsStatuses:
-       - name: subset-a
-         missingReplicas: 0
-       - name: subset-b
-         missingReplicas: 0
-   ---
-   workload:
-     replicas: 200
-   ```
-   ```yaml
-   CR:
-     metadata:
-       generation: 2
-     spec:
-       subsets:
-       - name: subset-a
-         maxReplicas: 200 # 100 + 100
-       - name: subset-b
-         maxReplicas: 100 # 100 + 0
-     status:
-       observedGeneration: 1 # should equal 2
-       subsetsStatuses:
-       - name: subset-a
-         missingReplicas: 0 # should equal 100
-       - name: subset-b
-         missingReplicas: 0
-   ---
-   workload:
-     replicas: 300 # 200 + 100
-   ```
-
-The workload scales up from 200 to 300, but the WorkloadSpread controller has not completed updating the CR status by comparing observedGeneration with generation.
-#### solution:
-Before the workload controller scales up, which should check whether the missingReplicas is true.
-
-2.The workload scales up when it's maxReplicas type is percent, but the WorkloadSpread controller doesn't update the latest workloadReplicas, the corresponding missingReplicas can't be scaled up in time.
-The possible reason is the race between the kruise informer and the controller-manager informer.
-
-For example:
-   ```yaml
-   CR:
-     metadata:
-       generation: 1
-     spec:
-       subsets:
-       - name: subset-a
-         maxReplicas: 50%
-       - name: subset-b
-         missingReplicas: 50%
-     status:
-       observedGeneration: 1
-       observedWorkloadGeneration: 2
-       subsetsStatuses:
-       - name: subset-a
-         missingReplicas: 0
-       - name: subset-b
-         missingReplicas: 0
-   ---
-   workload:
-     generation: 2
-     replicas: 200
-   ```
-   ```yaml
-   CR:
-     metadata:
-       generation: 1
-     spec:
-       subsets:
-       - name: subset-a
-         maxReplicas: 50% # 100 + 100
-       - name: subset-b
-         missingReplicas: 50% # 100 + 0
-     status:
-       observedGeneration: 1
-       observedWorkloadGeneration: 2
-       subsetsStatuses:
-       - name: subset-a
-         missingReplicas: 0 # should equal 50
-       - name: subset-b
-         missingReplicas: 0 # should equal 50
-   ---
-   workload:
-     generation: 3
-     replicas: 300 # 200 + 100
-   ```
-The workload scales up from 200 to 300 and doesn't change the subset maxReplicas. The generation of WorkloadSpread is ture, but the CR doesn't adjust missingReplicas to 50 because the informer race.
-
-#### Solution
-Add a observedWorkloadGeneration into status represents the controller reconciles the latest CR with corresponding workload' generation. Webhook should mutate Pod only if the observedWorkloadGeneration is equal the workload's generation.
-However, it is a minor possibility that two generation are equal but are all the old generation because the kruise informer' watch is slower than the controller-manager's informer.
-
 ## Caution
 When you adjust the subset's maxReplicas, you need to trigger the workload's rollout to make the existing Pods meeting the new topology spread.
+
+The workload scales up when it's maxReplicas type is percent, but the new Pods don't meet the desired spread.
+The possible reason is the race between the kruise informer and the controller-manager informer.
+The controller doesn't update the latest workloadReplicas, leading to the corresponding missingReplicas can't be scaled up in time.
 
 ## Alternative Considered
 
--- a/docs/proposals/20210628-workloadspread.md
+++ b/docs/proposals/20210628-workloadspread.md
@@ -24,7 +24,7 @@ any additional information provided beyond the standard proposal template.
 Define a CRD named WorkloadSpread, which can support configuration and management for elastic and multi-domain applications,
 even more powerful and flexible than [UnitedDeployment](https://openkruise.io/en-us/docs/uniteddeployment.html) in some ways.
 
-Not like the workload type, WorkloadSpread can be used with those stateless workloads, such as CloneSet, Deployment and ReplicaSet.
+Not like the workload type, WorkloadSpread can be used with those stateless workloads, such as CloneSet, Deployment, ReplicaSet and even Job.
 
 So users needn't learn the usage of a new workload and how to deploy and update their application by it.
 They only have to create WorkloadSpread CRs related to their existing workloads. Everything is just great!
@@ -232,15 +232,22 @@ And then controller cleans up all unscheduable Pods to trigger workload creating
 The unscheduled subset can be kept for 10 minutes and then should be recovered schedulable to schedule Pod again by controller.
 
 ### Deletion priority
+We use deletion-cost feature to restrict replica numbers of each subset when scaling down.
+If `controller.kubernetes.io/pod-deletion-cost` annotation is set, then the pod with the lower value will be deleted first.
 
 We have three types for subset's Pod deletion-cost
 1. the number of active Pods in this subset <= maxReplicas, deletion-cost = 100. indicating the priority of Pods \
    in this subset is higher than other Pods in workload.
-2. subset's maxReplicas is nil, deletion-cost = 0. indicating the priority of Pods in this subset is lower than \
-   other subsets that have a not nil maxReplicas.
+2. subset's maxReplicas is nil, deletion-cost = 0.
 3. the number of active Pods in this subset > maxReplicas, two class: (a) deletion-cost = -100, (b) deletion-cost = +100. \
    indicating we prefer deleting the Pod have -100 deletion-cost in this subset in order to control the instance of subset \
    meeting up the desired maxReplicas number.
+
+If WorkloadSpread changes its subset maxReplicas, Pods will not be recreated, but WorkloadSpread will adjust deletion-cost annotation through the above algorithm.
+**We shouldn't change maxReplicas frequently because this operation may increase the pressure of api-server.**
+
+To change scheduleStrategy of WorkloadSpread will keep the deletion-cost annotation of Pod.
+
 ## Alternative Considered
 
 ### UnitedDeployment
--- a/docs/proposals/20210628-workloadspread.md
+++ b/docs/proposals/20210628-workloadspread.md
@@ -7,7 +7,7 @@ reviewers:
   - "@furykerry"
   - "@FillZpp"
 creation-date: 2021-06-28
-last-updated: 2021-07-05
+last-updated: 2021-07-06
 status: implementable
 ---
 
@@ -53,14 +53,14 @@ spec:
         operator: In
         values:
         - zone-a
-    preferredNodeSelectorTerms:
-    - weight: 50
-      preference:
-      matchExpressions:
-      - key: topology.kubernetes.io/zone
-        operator: In
-        values:
-        - zone-c
+#   preferredNodeSelectorTerms:
+#   - weight: 50
+#     preference:
+#     matchExpressions:
+#     - key: topology.kubernetes.io/zone
+#       operator: In
+#       values:
+#       - zone-a
     maxReplicas: 3
     tolertions: []
     patch:
--- a/docs/proposals/20210628-workloadspread.md
+++ b/docs/proposals/20210628-workloadspread.md
@@ -1,0 +1,237 @@
+---
+title: WorkloadSpread
+authors:
+  - "@BoltsLei"
+
+reviewers:
+  - "@Fei-Guo"
+  - "@furykerry"
+  - "@FillZpp"
+
+creation-date: 2021-06-28 
+last-updated: 2021-06-28 
+status: implementable
+---
+
+# WorkloadSpread
+
+## Table of Contents
+
+A table of contents is helpful for quickly jumping to sections of a proposal and for highlighting
+any additional information provided beyond the standard proposal template.
+[Tools for generating](https://github.com/ekalinin/github-markdown-toc) a table of contents from markdown are available.
+
+
+## Motivation
+
+Define a CRD named WorkloadSpread, which can support configuration and management for elastic and multi-domain applications,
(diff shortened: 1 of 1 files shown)


docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,223 @@
...
+      The key in the map is the name of the pod and the value is the time when the webhook
+      handler process the creation request. If the real creation didn't happen and a pod is
+      still in this map, it will be removed from the list automatically by WorkloadSpread controller
+      after some time.\
+      If everything goes smooth this map should be empty for the most of the time.
+      Large number of entries in the map may indicate problems with pod creations.
+  - `deletingPods`: the map of Pods been deleted or evicted that just noticed by webhook
+    > deletingPods is similar with creatingPods and it contains information about pod deletion.
+  - `subsetUnscheduledStatus`: contains the details for the unscheduled subset.
+     - `unschedulable`: unschedulable is true indicates this Subset cannot be scheduled. The default is false.
```
zmberg, line 142: Please indicate that "unschedulable" is temporary, and the conditions for setting and restoring.

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,223 @@
...
+     - `unscheduledTime`: last time for unscheduled.
+     - `failedCount`: the number of subset was marked with unschedulable.
```
zmberg, line 144: Does failedCount have any real meaning in the actual processing logic, other than recording the number of failures?

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,223 @@
...
+   (3). if it succeeds, inject the rules of this subset into the Pod. 
+2. if `missingReplicas` = 0, go to the next subset
+3. if there is no available subset left, just let the Pod go
+
+##### Pod deletion 
+Also, when a Pod that belongs to a subset is deleting or evicting, webhook will put it into the `deletingPods` map in subsetStatus and update `missingReplicas` += 1.
+
+##### Update Pod Status
+When a Pod that belongs to a subset changes status phase to 'succeed' or 'failed', which means Pod has been terminated lifecycle, \
+webhook will clean it from `creatingPods` map and update `missingReplicas` -= 1.
```
zmberg, line 174: Shouldn't the "Update pod status" be implemented by the controller? Please describe in detail the controller for handling subset status

docs/proposals/20210628-workloadspread.md
BoltsLei: In this moment, failedCount just records the number of failures. Every failure can lead subset unschedulable for 10 minutes. In the future, the failCount may influence the unschedulable duration.

docs/proposals/20210628-workloadspread.md
BoltsLei: The process of "**update pod status**"  is every fast, controller may not update `missingReplicas` in time, which 
can cause error for Pods spread in multiple subset.

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,223 @@
...
+
+**spec**:
+- `targetRef`: reference to a specific workload, it supports `Deployment` in `apps` group and `CloneSet` in `apps.kruise.io` group
+- `subsets`: list of subsets which defines different topologies
+  - `name`: name of this subset
+  - `requiredNodeSelectorTerm`: the additional node selector of this subset, it will be injected into each term of `pod.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms`
+  - `preferredNodeSelectorTerms`: the additional node selector of this subset, it will be injected into each term of `pod.spec.affinity.nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution`
+  - `tolerations`: the additional tolerations of this subset, it will be injected into `pod.spec.tolerations`
+  - `patch`: the specified strategic patch body of this subset, it will be patched to Pods that belong to this subset
+  - `maxReplicas`: the maximum number of Pods in this subset, it can be an absolute number or percentage string. maxReplicas
```
Fei-Guo, line 107: I suggest to change `maxReplicas` to `replicas`. `maxReplica` means the controller will remove the extra Pods proactively which seems not to be the case for WorkloadSpread controller.  It should be a preferred number, not a max number.

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,223 @@
...
+   (1). update `missingReplicas` -= 1 and put this pod into `creatingPods` map and try to update subsetStatus. \
+   (2). if it conflicts, get the WorkloadSpread again and go back to step (1). \
+   (3). if it succeeds, inject the rules of this subset into the Pod. 
+2. if `missingReplicas` = 0, go to the next subset
+3. if there is no available subset left, just let the Pod go
+
+##### Pod deletion 
+Also, when a Pod that belongs to a subset is deleting or evicting, webhook will put it into the `deletingPods` map in subsetStatus and update `missingReplicas` += 1.
+
+##### Update Pod Status
```
Fei-Guo, line 172: Shouldn't a Pod be removed from the `creatingPod` map when its status is changed to "running"?

Fei-Guo (commented):
High level question:

1) Should WorkloadSpread CR be associated with a particular Workload instance meaning
 setting OwnerReference to the workload CR; or the targetRef needs to contain UID and the  controller needs to match the exact workload CR instance.
2) Will both webhook and controller update the `missingReplicas` field in the status? If so, how to avoid conflict?

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,223 @@
...
+The controller is responsible for calculating the real status, and the webhook mainly counts missingReplicas and records the creation or deletion entry of Pod into map.
+
+#### Replicas control
+
+##### Pod creation 
+When a Pod is creating and its workload has a related WorkloadSpread, webhook will check the subsetStatuses in the WorkloadSpread one by one:
+
+1. if `missingReplicas` > 0 or `missingReplicas` = -1 and this subset is not unschedulable, choose this subset temporarily \
+   (1). update `missingReplicas` -= 1 and put this pod into `creatingPods` map and try to update subsetStatus. \
+   (2). if it conflicts, get the WorkloadSpread again and go back to step (1). \
```
furykerry, line 164: if it conflicts -> if update to WorkloadSpread conflicts with other updates

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,223 @@
...
+   (3). if it succeeds, inject the rules of this subset into the Pod.
```
furykerry, line 165: if it succeeds ->  if update to WorkloadSpread succeeds

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,223 @@
...
+
+1. if `missingReplicas` > 0 or `missingReplicas` = -1 and this subset is not unschedulable, choose this subset temporarily \
+   (1). update `missingReplicas` -= 1 and put this pod into `creatingPods` map and try to update subsetStatus. \
+   (2). if it conflicts, get the WorkloadSpread again and go back to step (1). \
+   (3). if it succeeds, inject the rules of this subset into the Pod. 
+2. if `missingReplicas` = 0, go to the next subset
+3. if there is no available subset left, just let the Pod go
+
+##### Pod deletion 
+Also, when a Pod that belongs to a subset is deleting or evicting, webhook will put it into the `deletingPods` map in subsetStatus and update `missingReplicas` += 1.
```
furykerry, line 170: when a Pod that belongs to a subset is deleting or evicting ->
 when a Pod that belongs to a subset is being deleted or evicted

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,223 @@
...
+
+##### Update Pod Status
+When a Pod that belongs to a subset changes status phase to 'succeed' or 'failed', which means Pod has been terminated lifecycle, \
+webhook will clean it from `creatingPods` map and update `missingReplicas` -= 1.
+
+#### Reschedule strategy
+rescheduleSubset will delete some unscheduled Pods that still in pending status. Some subsets have no sufficient resource can lead to some Pods scheduled failed.
```
furykerry, line 177: plz be more specific about how many pods will deleted and who will delete the pods, all unscheduled pods will be deleted?

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,223 @@
...
+
+WorkloadSpread has multiple subset, so some Pods scheduled failed should be rescheduled to other subsets. \
```
furykerry, line 179: Pods scheduled failed -> unschedulable pods
controller will mark the subset contains Pods scheduled failed  -> controller will mark the subset containing unschedulable pods

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,223 @@
...
+### Requirements
+
+WorkloadSpread requires Pod webhook to inject rules into Pods and notice the deletion and eviction of Pods.
+
+So if the `PodWebhook` feature-gate is set to `false`, WorkloadSpread will also be disabled.
+
+### Implementation Details/Notes/Constraints
+
+WorkloadSpread have both webhook and controller. Controller should collaborate with webhook to maintain WorkloadSpread status together. \
+The controller is responsible for calculating the real status, and the webhook mainly counts missingReplicas and records the creation or deletion entry of Pod into map.
```
furykerry, line 155: The webhook is responsible for injecting rules into pod, reducing missingReplicas and recording the creation or deletion entry of Pod into map.
The controller is responsible for updating the missingReplicas along with other statics, and clean creatingpods and deletingPods.

furykerry (changes requested):

docs/proposals/20210628-workloadspread.md
BoltsLei: Controller will remove the extra Pods when the number of Pods belongs to this subset is less than maxReplicas, which is in line with our expectations.  Of course,  controller can't remove Pods directly rather than setting different deletion-cost annotation for Pods.
User can scale down subset's maxReplicas and workload's replicas to reduce the pressure on the subset or only release resources .

docs/proposals/20210628-workloadspread.md
BoltsLei: Controller will remove Pod entry that has been created from `creatingPods` map. The definition of `create` is that the controller discovers the Pod, regardless of whether its status is running, pending or otherwise.

BoltsLei: In addition, the reason we update missingReplicas and remove entry from creatingMap is that the kubelet can update Pod phase to `failed`  in a very short time due to unsatisfied resources. Controller can't update the missingReplicas in time, which will cause error for Pods spread in multiple subset. Therefore，webhook may be better in this scenario.

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,287 @@
...
+      The key in the map is the name of the pod and the value is the time when the webhook
+      handler process the creation request. If the real creation didn't happen and a pod is
+      still in this map, it will be removed from the list automatically by WorkloadSpread controller
+      after some time.\
+      If everything goes smooth this map should be empty for the most of the time.
+      Large number of entries in the map may indicate problems with pod creations.
+  - `deletingPods`: the map of Pods been deleted or evicted that just noticed by webhook
+    > deletingPods is similar with creatingPods and it contains information about pod deletion.
+  - `subsetUnscheduledStatus`: contains the details for the unscheduled subset.
+     - `unschedulable`: unschedulable is true indicates this Subset cannot be scheduled. The default is false.
```
furykerry, line 172: why the spread have to track the unschedulable status explicitly, why not translate the spread into preferredDuringSchedulingIgnoredDuringExecution

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,287 @@
...
+        values:
+        - zone-a
+    preferredNodeSelectorTerms:
+    - weight: 50
+      preference:
+      matchExpressions:
+      - key: topology.kubernetes.io/zone
+        operator: In
+        values:
+        - zone-b
```
furykerry, line 63: the sample here is problematic, the prefer terms conflicts with the required terms

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,287 @@
...
+| 50  | 50 |
+| 30  | 0  |
+
+The subset-b will only be allocated Pods when the application load is higher (100 replicas), and not provide resources during the rest of the time.
+This will reduce a lot of resources belongs to subset-b for us, and the application can get better elastic support.
+
+#### NodeAffinity
+
+Add nodeAffinity to the workload template, such as preferredDuringSchedulingIgnoredDuringExecution, which can be scheduled in multiple regions.
+However, It cannot specify the replica numbers of the subset, and it is only valid when the Pod is scheduled.
```
furykerry, line 271: >  It cannot specify the replica numbers of the subset, and it is only valid when the Pod is scheduled.
it cannot limit the replica numbers of a subset, and it is only effective when scaling out the workload, not effective when scaling in

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,287 @@
...
+    controller.kubernetes.io/pod-deletion-cost: -100 / 100 / 0
+```
+
+(In 1.21, users need to enable `PodDeletionCost` feature-gate, and since 1.22 it will be enabled by default)
+
+### Alternative Considered
+
+#### UnitedDeployment
+
+Both UnitedDeployment and WorkloadSpread allow the workload to be distributed in different subsets.
```
furykerry, line 247: i think the key difference between UnitedDeployment and WorkloadSpread is that WorkloadSpread requires NO workload api change, one can use vanilla CloneSet even Deployment while attaching extra topology constraints. The UnitedDeployment can also be improved to understand the subset preferences

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,287 @@
...
+      failedCount: 1
+  - name: subset-b
+    missingReplicas: -1
+```
+
+**spec**:
+- `targetRef`: reference to a specific workload, it supports `Deployment` in `apps` group and `CloneSet` in `apps.kruise.io` group
+- `subsets`: list of subsets which defines different topologies
+  - `name`: name of this subset
+  - `requiredNodeSelectorTerm`: the additional node selector of this subset, it will be injected into each term of `pod.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms`
```
furykerry, line 103: > it will be injected into each term of

the nodes corresponding to each term in requiredDuringSchedulingIgnoredDuringExecution are intersected, so is it more appropriate to just append the term into the  requiredNodeSelectorTerm?

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,287 @@
...
+  - `preferredNodeSelectorTerms`: the additional node selector of this subset, it will be injected into each term of `pod.spec.affinity.nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution`
```
furykerry, line 104: it will be injected into each term 
-> 
it will be appended into the preferredDuringSchedulingIgnoredDuringExecution

furykerry (changes requested):

docs/proposals/20210628-workloadspread.md
BoltsLei: `pod.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms ` is a list of node selector terms and the terms are ORed. 
So we need to inject `requiredNodeSelectorTerm` into each terms to enforce scheduling Pod to subset.

It is more appropriate to append the term into the `preferredNodeSelectorTerms` .

Fei-Guo:
The elastic capacity use case looks reasonable. Thanks

LGTM.

One side note: We should alert users that they should use this feature with caution. If the workload is scaled but the workload spread CR is not adjusted accordingly, unexpected behaviors may happen.

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,291 @@
...
+A table of contents is helpful for quickly jumping to sections of a proposal and for highlighting
+any additional information provided beyond the standard proposal template.
+[Tools for generating](https://github.com/ekalinin/github-markdown-toc) a table of contents from markdown are available.
+
+## Motivation
+
+Define a CRD named WorkloadSpread, which can support configuration and management for elastic and multi-domain applications,
+even more powerful and flexible than [UnitedDeployment](https://openkruise.io/en-us/docs/uniteddeployment.html) in some ways.
+
+Not like the workload type, WorkloadSpread can be used with those stateless workloads, such as CloneSet, Deployment and ReplicaSet.
```
furykerry, line 27: is it possible to support job workloads?

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,291 @@
...
+### Deletion priority
+
+We have three types for subset's Pod deletion-cost
+1. the number of active Pods in this subset <= maxReplicas, deletion-cost = 100. indicating the priority of Pods \
+   in this subset is higher than other Pods in workload.
+2. subset's maxReplicas is nil, deletion-cost = 0. indicating the priority of Pods in this subset is lower than \
+   other subsets that have a not nil maxReplicas.
+3. the number of active Pods in this subset > maxReplicas, two class: (a) deletion-cost = -100, (b) deletion-cost = +100. \
+   indicating we prefer deleting the Pod have -100 deletion-cost in this subset in order to control the instance of subset \
+   meeting up the desired maxReplicas number.
```
furykerry, line 243: what if workload spread is changed, for example change maxreplicas or scheduleStrategy. will the pod be recreated ?

furykerry (changes requested):

docs/proposals/20210628-workloadspread.md
BoltsLei: If workload spread changes maxReplicas, the Pods will not be recreated, and workload spread will adjust  deletion-cost annotation through the above algorithm.
To change scheduleStrategy of workload spread will keep the deletion-cost annotation for Pods.

docs/proposals/20210628-workloadspread.md
BoltsLei: In my opinion,  workload spread can also support the Job workload. We have considered the Pod status for 
`succeed` or `failed`.

furykerry (approved):

docs/proposals/20210628-workloadspread.md
```diff
@@ -0,0 +1,310 @@
...
+
+If WorkloadSpread changes its subset maxReplicas, Pods will not be recreated, but WorkloadSpread will adjust deletion-cost annotation through the above algorithm.
+
+To change scheduleStrategy of WorkloadSpread will keep the deletion-cost annotation of Pod.
+
+### MissingReplicas consistency
+
```
zmberg, line 252: MissingReplicas consistency -> Spread.Spec and Spread.Status consistency issues

zmberg: Also clarify the consistency issues and how to deal with it

furykerry (commented):
/lgtm

Fei-Guo:
I believe we have agreed on removing WorkloadGeneration and SpreadGeneration from status? We should just explain the possible race in the design doc instead.

BoltsLei:
@Fei-Guo, we should retain the `ObservedGeneration` to indicates the most recent generation observed for this WorkloadSpread. Most workloads have this filed in status(such as StatefulSet, CloneSet).
And we just provide the recent generation and don't attach any logic to it.

PR #655, opened 2021-06-30
adjust sidecarset control api interface UpdateSidecarContainerToLatest

zmberg:

Function UpdateSidecarContainerToLatest can't completely block the difference between the group and the community, so the method is adjusted to UpgradeSidecarContainer, which is used to determine and generate the latest version of sidecar container.


furykerry (commented):
lgtm

PR #659, opened 2021-07-06
workloadspread api and validating webhook

BoltsLei:
Signed-off-by: bolstlei <1606880753@qq.com>

workloadspread api and validating webhook


pkg/webhook/workloadspread/validating/workloadspread_validation.go
```diff
@@ -0,0 +1,253 @@
...
+					subsetMaxReplicas, err := intstr.GetValueFromIntOrPercent(subset.MaxReplicas, 100, true)
+					if err != nil || subsetMaxReplicas < 0 {
+						allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("maxReplicas"), subset.MaxReplicas, "maxReplicas is not valid for subset"))
+					}
+					if subset.MaxReplicas.Type == intstr.String && subsetMaxReplicas > 100 {
+						allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("maxReplicas"), subset.MaxReplicas, "maxReplicas exceeds 100% for subset"))
+					}
+				} else {
+					// validate maxReplicas of the default subset
+					allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("maxReplicas"), subset.MaxReplicas, "maxReplicas of default subset must be nil"))
```
Fei-Guo, line 202: Do this mean if WorkloadSpread CR has only one subset, then its replica number has to be nil, which means the CR does not affect any scheduling decision? 

If so, should we enforce that a WorkloadSpread CR has to have at least two subsets?

BoltsLei: One subset also affects scheduling decisions because the webhook injects the subset topology rules into the Pod, and the maxReplicas is nil, which means there is no limit for replica numbers in the subset.
All in all, a subset can distribute the workload in a fixed zone.

Fei-Guo: This makes no difference comparing to applying node selector and tolerations in workload Pod Spec directly, isn't it?

BoltsLei: Indeed, one subset has no meaning for workload spread, and should has at least two subset. I will adjust the validate rules to enforce CR at least two subset.

pkg/webhook/workloadspread/validating/workloadspread_validation_test.go
```diff
@@ -0,0 +1,635 @@
...
+	"strconv"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
```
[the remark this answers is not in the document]
BoltsLei: ignore

pkg/webhook/workloadspread/validating/workloadspread_validation.go
```diff
@@ -0,0 +1,253 @@
...
+	corevalidation "k8s.io/kubernetes/pkg/apis/core/validation"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
+)
+
+var (
+	controllerKruiseKindCS = appsv1alpha1.SchemeGroupVersion.WithKind("CloneSet")
+	controllerKindRS       = appsv1.SchemeGroupVersion.WithKind("ReplicaSet")
+	controllerKindDep      = appsv1.SchemeGroupVersion.WithKind("Deployment")
```
furykerry, line 40: add job kind

pkg/webhook/workloadspread/validating/workloadspread_validation.go
```diff
@@ -0,0 +1,253 @@
...
+			case controllerKindDep.Kind:
+				ok, err := verifyGroupKind(spec.TargetReference, controllerKindDep.Kind, []string{"apps"})
+				if !ok || err != nil {
+					allErrs = append(allErrs, field.Invalid(fldPath.Child("targetRef"), spec.TargetReference, "TargetReference is not valid for Deployment."))
+				}
+			case controllerKindRS.Kind:
+				ok, err := verifyGroupKind(spec.TargetReference, controllerKindRS.Kind, []string{"apps"})
+				if !ok || err != nil {
+					allErrs = append(allErrs, field.Invalid(fldPath.Child("targetRef"), spec.TargetReference, "TargetReference is not valid for ReplicaSet."))
+				}
```
furykerry, line 104: add job kind and batch group related logic

pkg/webhook/workloadspread/validating/workloadspread_validation.go
```diff
@@ -0,0 +1,253 @@
...
+
+	// validate subsets
+	allErrs = append(allErrs, validateWorkloadSpreadSubsets(spec.Subsets, fldPath.Child("subsets"))...)
+
+	// validate scheduleStrategy
+	if spec.ScheduleStrategy.Type != "" &&
+		spec.ScheduleStrategy.Type != appsv1alpha1.FixedWorkloadSpreadScheduleStrategyType &&
+		spec.ScheduleStrategy.Type != appsv1alpha1.AdaptiveWorkloadSpreadScheduleStrategyType {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("scheduleStrategy").Child("type"),
+			spec.ScheduleStrategy.Type, "ScheduleStrategy's type is not valid"))
```
furykerry, line 119: add the input schedulestrategy type in the message

pkg/webhook/workloadspread/validating/workloadspread_validation.go
```diff
@@ -0,0 +1,253 @@
...
+	return false, nil
+}
+
+func (h *WorkloadSpreadCreateUpdateHandler) validatingWorkloadSpreadFn(obj *appsv1alpha1.WorkloadSpread) field.ErrorList {
+	// validate ws.spec.
+	allErrs := validateWorkloadSpreadSpec(obj, field.NewPath("spec"))
+
+	// validate whether ws.spec.targetRef is in conflict with others.
+	wsList := &appsv1alpha1.WorkloadSpreadList{}
+	if err := h.Client.List(context.TODO(), wsList, &client.ListOptions{Namespace: obj.Namespace}); err != nil {
```
furykerry, line 69: add disable deepcopy options

pkg/webhook/workloadspread/validating/workloadspread_validation.go
```diff
@@ -0,0 +1,253 @@
...
+				allErrs = append(allErrs, corevalidation.ValidateTolerations(coreTolerations, fldPath.Index(i).Child("tolerations"))...)
+			}
+
+			if subset.MaxReplicas != nil {
+				if i < len(subsets)-1 {
+					subsetMaxReplicas, err := intstr.GetValueFromIntOrPercent(subset.MaxReplicas, 100, true)
+					if err != nil || subsetMaxReplicas < 0 {
+						allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("maxReplicas"), subset.MaxReplicas, "maxReplicas is not valid for subset"))
+					}
+					if subset.MaxReplicas.Type == intstr.String && subsetMaxReplicas > 100 {
```
furykerry, line 197: 1. do we need to check whether the total maxreplicas is less than 1 if all maxreplicas is specified as percent
2. what if the total maxreplicas is greater than the workload replicas?

pkg/webhook/workloadspread/validating/workloadspread_validation.go
```diff
@@ -0,0 +1,253 @@
...
+						allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("maxReplicas"), subset.MaxReplicas, "maxReplicas exceeds 100% for subset"))
+					}
+				} else {
+					// validate maxReplicas of the end subset
+					allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("maxReplicas"), subset.MaxReplicas, "maxReplicas of end subset must be nil"))
```
furykerry, line 202: is the last subset valid if maxreplicas is specifed as percent?

pkg/webhook/workloadspread/validating/workloadspread_validation.go
```diff
@@ -0,0 +1,253 @@
...
+					if err := corev1.Convert_v1_Toleration_To_core_Toleration(&toleration, coreToleration, nil); err != nil {
+						allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("tolerations"), subset.Tolerations, fmt.Sprintf("Convert_v1_Toleration_To_core_Toleration failed: %v", err)))
+					} else {
+						coreTolerations = append(coreTolerations, *coreToleration)
+					}
+				}
+				allErrs = append(allErrs, corevalidation.ValidateTolerations(coreTolerations, fldPath.Index(i).Child("tolerations"))...)
+			}
+
+			if subset.MaxReplicas != nil {
```
furykerry, line 191: is it valid if ScheduleStrategy is fixed and max replicas is not specified

furykerry (changes requested):

pkg/webhook/workloadspread/validating/workloadspread_validation.go
BoltsLei: The official controller-runtime pkg may not support the disable  deepcopy options?

pkg/webhook/workloadspread/validating/workloadspread_validation.go
BoltsLei: We enforce user to set the last subset' maxReplicas with nil.

pkg/webhook/workloadspread/validating/workloadspread_validation.go
BoltsLei: 1. Less than 1 should be permitted because the last subset's maxReplicas is nil, which will contain the remaining Pods.
2. We can validate the **percent** condition if the total maxReplicas > workload Replicas. The other condition should be valid because we can't know the actual workload replicas when validating.

pkg/webhook/workloadspread/validating/workloadspread_validation.go
BoltsLei: It should be valid.

pkg/webhook/workloadspread/validating/workloadspread_validation.go
zmberg: @Fei-Guo 
This is considered because there is a moment of inconsistency between spec and status (the user adjusts the spec and then the controller takes some time to adjust the status), and at this point it is hoped that there will be a default subset to give the webhook the ability to do the injection.

Now this is not really appropriate, so webhook will determine whether the spec.generation and status.generation are equal when injecting, if not, etc. temporarily reject the request, and then let the workload controller retry to ensure that the webhook injection when the spec and status is consistent. So it is not necessary to set the last subset to nil.

furykerry (changes requested):

pkg/webhook/workloadspread/validating/workloadspread_validation.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+	}
+
+	return allErrs
+}
+
+func validateWorkloadSpreadSubsets(subsets []appsv1alpha1.WorkloadSpreadSubset, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if len(subsets) == 0 {
```
furykerry, line 141: is it more valid if len is greater than 1

furykerry (changes requested):

apis/apps/v1alpha1/workloadspread_types.go
```diff
@@ -0,0 +1,210 @@
...
+	// Contains the status of each subset. Each element in this array represents one subset
+	// +optional
+	SubsetStatuses []WorkloadSpreadSubsetStatus `json:"subsetStatuses,omitempty"`
+}
+
+// SubsetUnscheduledStatus contains the details for the unscheduled subset.
+type SubsetUnscheduledStatus struct {
+	// Unschedulable is true indicates this Subset cannot be scheduled. The default is false.
+	// +optional
+	Unschedulable bool `json:"unschedulable,omitempty"`
```
Fei-Guo, line 136: It is not clear how to define a subset is schedulable or not? Will a single Pod scheduling failure triggers this status update? What if out of x replicas, only one Pod fails. Is this subset scheduable or not?

apis/apps/v1alpha1/workloadspread_types.go
```diff
@@ -0,0 +1,210 @@
...
+	// +optional
+	Patch runtime.RawExtension `json:"patch,omitempty"`
+}
+
+// WorkloadSpreadStatus defines the observed state of WorkloadSpread.
+type WorkloadSpreadStatus struct {
+	// ObservedGeneration is the most recent generation observed for this WorkloadSpread. It corresponds to the
+	// WorkloadSpread's generation, which is updated on mutation by the API Server.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
Fei-Guo, line 121: Is Generation equal to ResourceVersion in object meta?

apis/apps/v1alpha1/workloadspread_types.go
BoltsLei: Scheduler will try to reschedule Pod when scheduled failure by preempting lower priority pods if possible.. In our design, controller will tag this subset unschulable when the subset has one or more than one failed Pod when failed time more than a giving duration.

pkg/webhook/workloadspread/validating/workloadspread_validation.go
BoltsLei: Because the maxReplicas is nil means no replica limits, we only permit the end subset' maxReplicas is nil.
If the nil subset is not the last, the back subset is invalid.

apis/apps/v1alpha1/workloadspread_types.go
BoltsLei: #### How to define a subset is unschedulable
The subset has one or more than one Pod by checking it's status `condition.type=PodScheduled`, 'condition.status=False', and `condition.reason=Unschedulable`

#### Why is it designed
The subset nodes has no available resources to receive all Pods, and it is possible that the failure pods is one or more than one.

zmberg: 1. When a pod in a subset has condition[PodScheduled] = false, the subset is considered temporarily unschedulable. The condition[PodScheduled] = false, means that the cluster does not have enough resources to schedule at this moment, even if only one Pod scheduling fails.
2. This state is temporary, and after a period of time, the spread controller will mark it schedulable.

pkg/webhook/workloadspread/validating/workloadspread_validation.go
zmberg: 1. all subset maxReplicas must be the same type: int or percent
2. only last subset max replicas can be not specified
2. Fixed
- if all maxreplicas is specified as percent, the total maxreplicas must equals 1, except the last subset is not specified.
3. Adaptive
- the last subset must be not specified

apis/apps/v1alpha1/workloadspread_types.go
zmberg: Generation equals to generation in object meta

zmberg:
LGTM

apis/apps/v1alpha1/workloadspread_types.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+// SubsetUnscheduledStatus contains the details for the unscheduled subset.
+type SubsetUnscheduledStatus struct {
+	// LastScheduleFailureTime indicates the last time for unscheduled subset.
+	// It's valid only when the WorkloadSpreadScheduleStrategy is 'adaptive', otherwise it is always null.
+	// When one or more one pods in a subset have condition[PodScheduled] = false, the subset is considered temporarily unschedulable.
+	// The condition[PodScheduled] = false, which means the cluster does not have enough resources to schedule at this moment,
+	// even if only single Pod scheduled fails.
+	// The unschedulable state is temporary, and after a period of time(e.g. 5m), the spread controller will make it null to schedulable.
```
Fei-Guo, line 135: Please fix the comments to match with the field.

Fei-Guo (commented):
LGTM. Please fix the minor comment issue.

apis/apps/v1alpha1/workloadspread_types.go
```diff
@@ -0,0 +1,204 @@
...
+type SubsetUnscheduledStatus struct {
+	// LastScheduleFailureTime indicates the last time for unscheduled subset.
+	// It's valid only when the WorkloadSpreadScheduleStrategy is 'adaptive', otherwise it is always null.
+	// When one or more one pods in a subset have condition[PodScheduled] = false, the subset is considered temporarily unschedulable.
+	// The condition[PodScheduled] = false, which means the cluster does not have enough resources to schedule at this moment,
+	// even if only single Pod scheduled fails. Controller will set this field with a unschedulable time.
+	// The LastScheduleFailureTime is valid in a period of time(e.g. 5m), and the webhook will skip this subset by checking whether time.Now() < LastScheduleFailureTime + 5m.
+	// After a period of time，the controller will set this filed to empty, indicating that it is schedulable, and the webhook can't ignore this subset.
+	// +optional
+	LastScheduleFailureTime *metav1.Time `json:"lastScheduleFailureTime,omitempty"`
```
furykerry, line 138: plz consider make the status fields more extensible(replace fixed schedule condition with conditions array), so that we can record other status in the subset here, for example, we can record that creating a pods in a subset will be denied by the quota

furykerry (changes requested):

apis/apps/v1alpha1/workloadspread_types.go
```diff
@@ -0,0 +1,204 @@
...
+
+// SubsetUnscheduledStatus contains the details for the unscheduled subset.
+type SubsetUnscheduledStatus struct {
+	// LastScheduleFailureTime indicates the last time for unscheduled subset.
+	// It's valid only when the WorkloadSpreadScheduleStrategy is 'adaptive', otherwise it is always null.
+	// When one or more one pods in a subset have condition[PodScheduled] = false, the subset is considered temporarily unschedulable.
+	// The condition[PodScheduled] = false, which means the cluster does not have enough resources to schedule at this moment,
+	// even if only single Pod scheduled fails. Controller will set this field with a unschedulable time.
+	// The LastScheduleFailureTime is valid in a period of time(e.g. 5m), and the webhook will skip this subset by checking whether time.Now() < LastScheduleFailureTime + 5m.
+	// After a period of time，the controller will set this filed to empty, indicating that it is schedulable, and the webhook can't ignore this subset.
```
Fei-Guo, line 136: "// After a period of time，the controller will set this filed to empty, indicating that it is schedulable, and the webhook can't ignore this subset."  The controller should not clear this field.

apis/apps/v1alpha1/workloadspread_types.go
Fei-Guo: +1 to add condition array

Fei-Guo (commented):
Overall LGTM

zmberg:
Lack of PEOJECT, kustomization.yaml WorkloadSpread

pkg/webhook/workloadspread/validating/workloadspread_create_update_handler.go
```diff
@@ -0,0 +1,90 @@
...
+	obj := &appsv1alpha1.WorkloadSpread{}
+	oldObj := &appsv1alpha1.WorkloadSpread{}
+
+	switch req.AdmissionRequest.Operation {
+	case admissionv1beta1.Create:
+		if err := h.Decoder.Decode(req, obj); err != nil {
+			return admission.Errored(http.StatusBadRequest, err)
+		}
+		if allErrs := h.validatingWorkloadSpreadFn(obj); len(allErrs) > 0 {
+			return admission.Errored(http.StatusUnprocessableEntity, allErrs.ToAggregate())
```
furykerry, line 56: it sounds better the error is http.StatusBadRequest

pkg/webhook/workloadspread/validating/workloadspread_create_update_handler.go
```diff
@@ -0,0 +1,90 @@
...
+		if err := h.Decoder.Decode(req, obj); err != nil {
+			return admission.Errored(http.StatusBadRequest, err)
+		}
+		if err := h.Decoder.DecodeRaw(req.AdmissionRequest.OldObject, oldObj); err != nil {
+			return admission.Errored(http.StatusBadRequest, err)
+		}
+
+		validationErrorList := h.validatingWorkloadSpreadFn(obj)
+		updateErrorList := validateWorkloadSpreadUpdate(obj, oldObj)
+		if allErrs := append(validationErrorList, updateErrorList...); len(allErrs) > 0 {
```
furykerry, line 68: it sounds better the error is http.StatusBadRequest, especially when validationErrorList is not empty

pkg/webhook/workloadspread/validating/workloadspread_validation.go
```diff
@@ -0,0 +1,295 @@
...
+	if err := h.Client.List(context.TODO(), wsList, &client.ListOptions{Namespace: obj.Namespace}); err != nil {
+		allErrs = append(allErrs, field.InternalError(field.NewPath(""), fmt.Errorf("query other WorkloadSpread failed, err: %v", err)))
+	} else {
+		allErrs = append(allErrs, validateWorkloadSpreadConflict(obj, wsList.Items, field.NewPath("spec"))...)
+	}
+
+	return allErrs
+}
+
+func validateWorkloadSpreadSpec(obj *appsv1alpha1.WorkloadSpread, fldPath *field.Path) field.ErrorList {
```
furykerry, line 80: can we valid the patch field, some concerns:
1. patch should not change some fields, such as image of podtemplate, labels that are part of workload selector
2. patch should at least be a valid part of podtemplate

furykerry (changes requested):

furykerry (commented):
/lgtm

pkg/webhook/workloadspread/validating/workloadspread_validation.go
```diff
@@ -0,0 +1,297 @@
...
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("targetRef"), spec.TargetReference, "TargetReference's GroupKind is not permitted."))
+			}
+		}
+	}
+
+	// validate subsets
+	allErrs = append(allErrs, validateWorkloadSpreadSubsets(spec.Subsets, fldPath.Child("subsets"))...)
+
+	// validate scheduleStrategy
+	if spec.ScheduleStrategy.Type != "" &&
```
FillZpp, line 122: According to the definition in `workloadspread_types.go`, the `spec.scheduleStrategy.type` field is optional. We should not raise error here if it is empty.
This may need a mutating webhook to inject default fields into workloadspread.

pkg/webhook/workloadspread/validating/workloadspread_validation.go
```diff
@@ -0,0 +1,297 @@
...
+					coreTolerations = append(coreTolerations, *coreToleration)
+				}
+			}
+			allErrs = append(allErrs, corevalidation.ValidateTolerations(coreTolerations, fldPath.Index(i).Child("tolerations"))...)
+		}
+
+		//TODO validate patch
+
+		//1. All subset maxReplicas must be the same type: int or percent.
+		//2. Only the last subset maxReplicas can be not specified, indicates subset replicas is not limited.
```
FillZpp, line 218: I'm not sure if we should limit only the last one can be defined with no maxReplicas.
What if users need to use it with `Adaptive` strategy? Which means always use the former subset if it has enough resource, otherwise try the later subsets.

pkg/webhook/workloadspread/validating/workloadspread_validation.go
zmberg: Is webhook too heavy if it's just to set the type field? If type is not filled, default fixed OK?

pkg/webhook/workloadspread/validating/workloadspread_validation.go
zmberg: Look at the fourth point below, When adaptive, the last subset must be not specified.

pkg/webhook/workloadspread/validating/workloadspread_validation.go
FillZpp: That will be fine. The `spec.scheduleStrategy` is not a pointer, so the yaml we get using kubectl will be

```yaml
kind: WorkloadSpread
spec:
  # ...
  scheduleStrategy: {}
```

Is it a little strange? (I'm ok for it)

pkg/webhook/workloadspread/validating/workloadspread_validation.go
FillZpp: I mean the non-last subset can also be defined with no `maxReplicas`.
For example,

```yaml
  - name: subset-a
    # ...
  - name: subset-b
    # ...
```

With Adaptive strategy, it means always use `subset-a` if it has enough resource. Otherwise, try the `subset-b`. So the former one do not need `maxReplicas`.

zmberg: subset-a maxReplicas=100% is ok?

apis/apps/v1alpha1/workloadspread_types.go
```diff
@@ -0,0 +1,222 @@
...
+	// +optional
+	Reason string `json:"reason,omitempty"`
+
+	// A human readable message indicating details about the transition.
+	// +optional
+	Message string `json:"message,omitempty"`
+}
+
+// WorkloadSpreadSubsetStatus defines the observed state of subset
+type WorkloadSpreadSubsetStatus struct {
```
zmberg, line 160: add Replicas filed indicates the number of Pods managed by the subset.

pkg/webhook/workloadspread/validating/workloadspread_validation.go
FillZpp: what if there are a,b,c three subsets?

furykerry (approved):

PR #667, opened 2021-07-12
ImagePullSecrets Proposal for sidecarSet and more things.

veophi:
1. Proposal: ImagePullSecrets for sidecarSet;
2. Update debug guide;

We provide two schemes for this feature, so we need a discuss.


docs/debug/README.md
```diff
@@ -39,43 +39,21 @@ Follow the [official minikube installation guide](https://kubernetes.io/docs/tas
...
-              value: kruise-webhook-server-secret
-          image: $DOCKERID/kruise:test
-          imagePullPolicy: Always
-          name: manager
-```
-
-* step 6: `kubectl delete sts kruise-controller-manager -n kruise-system` to remove the old statefulset if any;
-* step 7: `kubectl apply -f config/manager/all_in_one.yaml` to install the new statefulset with the customized controller manager image;
+* step 5: `export KUBECONFIG=<your_kube_config>` to choose the target k8s cluster, e.g., `export KUBECONFIG=~/.kube/config`;
+* step 6: `kubectl delete deployment kruise-controller-manager -n kruise-system` to remove the old Kruise Deployment if any;
```
FillZpp, line 55: There is no need to delete the deployment first, `make deploy` will apply the changes to the old resources.

And to delete all resources installed by `make deploy`, you should use `./scripts/uninstall.sh`. You can add this tip below.

docs/proposals/20210712-sidecarsetimagespullsecrets.md
```diff
@@ -0,0 +1,138 @@
...
+creation-date: 2021-07-12
+  
+last-updated: 2021-07-12
+  
+status: implementable
+
+---
+# SidecarSetImagePullSecrets
+
+- Provide a way to pull images using `Secret` from private repositories for sidecarSet.
```
FillZpp, line 21: Please replace all `sidecarSet` to `SidecarSet` or `sidecarset`. The former is kind, and the later is singular. I recommend the former one.
Usually we don't use word like `sidecarSet` except the object variable name in golang.

docs/proposals/20210712-sidecarsetimagespullsecrets.md
```diff
@@ -0,0 +1,138 @@
...
+Created by [gh-md-toc](https://github.com/ekalinin/github-markdown-toc)
+
+## Motivation
+
+In k8s, secret is an easy and safe way to store and manage sensitive information, such as passwords, OAuth tokens, and ssh keys.
+One of the most common-used features of it is to pull images from private repositories. However, sidecarSet does not support this feature so far.
+
+## Proposal
+
+When a sidecarSet is injected into the matched `Namespace.Pod`, the references of the secrets it needs also will be injected into the pod.
```
FillZpp, line 49: Do not use `Namespace.Pod` or `NS.POD` or `Namespace.Secret`. Just `Pod` and `Secret`, of cause they are in namespaced scope.

docs/proposals/20210712-sidecarsetimagespullsecrets.md
```diff
@@ -0,0 +1,138 @@
...
+This is a rough but effective way to solve these two problems. 
+
+First, we add the secret fields in `sidecarset_types.go`, it looks like this:
+```go
+type SidecarSetSpec struct {
+	...
+	SecretFields []SecretField
+}
+
+type SecretField struct {
```
FillZpp, line 71: If you want to put the secrets definition in SidecarSet, the data and name are not enough. Actually the `data` is a map and there should also have a `type` field. You may need to define `SecretTemplates []v1.Secret` in `SidecarSetSpec`.

Another question is that users should define the real data or the data after base64 encoding here?

docs/proposals/20210712-sidecarsetimagespullsecrets.md
```diff
@@ -0,0 +1,138 @@
...
+    ...
+  updateStrategy:
+    ...
+  containers:
+    - name: sidecar1
+      image: minchou/centos:6.7
+     ...
+  imagePullSecrets:
+    - name: test-secret
+      - naamespace: test-secret-namespace
```
FillZpp, line 116: `naamespace` is a typo. And why not just
```
  imagePullSecrets:
    - name: test-secret
      namespace: test-secret-namespace
```

Fei-Guo:
I don't suggest to ask SidecarSet controller to handle Secret lifecycle management (it will be complicated and out of the scope of SidecarSet). 

Hence, I vote to option 2 and just ask user to specify the secret name. The secret population/deletion should be done by another controller. If the specific secret does not exist in an injected Pod's namespace, the Pod creation should simply fail.

docs/proposals/20210713-resourcedistribution.md
```diff
@@ -0,0 +1,136 @@
...
+		if IsSecretExisted(secret, namespace) {
+			r.Client.Delete(ctx.TODO(), secret)
+		}
+	}
+}
+
+```
+
+### Risks and Mitigations
+Problem: When users delete the original secret, how to delete its copies in other namespaces?
```
hantmac, line 127: How to define `the original secret`,  I think the secrets created in `openkruise.io/sync-to` ns are equal.

Fei-Guo:
due to #671

PR #673, opened 2021-07-14
proposal: resource distribution

veophi:

proposal: resource distribution


docs/proposals/20210713-resourcedistribution.md
```diff
@@ -0,0 +1,145 @@
...
+#### Story 1: distribute to other namespaces
+```yaml
+# my-secret.yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: namespaced-secret
+  annotations:
+    openkruise.io/sync-to: "namespaces"
+    openkruise.io/sync-to/namespaces: "ns-1;ns-2;ns-3"
```
zmberg, line 61: 1. openkruise.io/sync-to/namespaces -> e.g. openkruise.io/sync-to-namespaces
2. whether the user can configure only one annotations, e.g. openkruise.io/sync-to-namespaces
3. support all-namespaces, except kube-system

docs/proposals/20210713-resourcedistribution.md
```diff
@@ -0,0 +1,145 @@
...
+  password: t0p-Secret
+```
+When `kubectl apply -f my-secret.yaml`,  the secret will be created or updated in **all namespaces**.
+
+### Implementation
+We will watch all the events about the `Secret`, and sync it when reconciling it.
+The main logic looks like this:
+
+#### WebHook Validation
+The value　of `Annotations["openkruise.io/sync-to"]` is limited in `["cluster", "namespace"]` , and it should be validated.
```
zmberg, line 88: Is there no need to do webhook validate checks in this place

FillZpp (commented):
Is it a good idea to implement resource distribution through the annotations or labels?

If so, we should specify the kinds that support to distribute, and kruise-manager has to list watch all objects of these kinds.

Comparing to define a CRD named ResourceDistribution and users put the object into CR, which design will be better?

hantmac:
> Is it a good idea to implement resource distribution through the annotations or labels?
> 
> If so, we should specify the kinds that support to distribute, and kruise-manager has to list watch all objects of these kinds.
> 
> Comparing to define a CRD named ResourceDistribution and users put the object into CR, which design will be better?

In my opinion, using annotations or labels is more convenient and easy to use,  and the new CRD requires learning costs for users. But define a CRD is easier to implement and has more clear code.

docs/proposals/20210713-resourcedistribution.md
veophi: ok

furykerry (approved):

PR #675, opened 2021-07-15
Add imagepullsecrets field for sidecarset

veophi:

Add imagepullsecrets for sidecarset

Add imagepullsecrets fied for sidecarset

Note: Users must  distribute their secrets to the required namespaces manually.


Fei-Guo:
LGTM

pkg/webhook/pod/mutating/sidecarset.go
```diff
@@ -239,6 +258,9 @@ func buildSidecars(isUpdated bool, pod *corev1.Pod, oldPod *corev1.Pod, matchedS
 			}
 		}
 
+		//process imagePullSecrets
```
furykerry, line 261: consider move the code to L290 so  that new pullsecrets only affect new created pod

furykerry (changes requested):

pkg/webhook/pod/mutating/sidecarset.go
zmberg: good idea

veophi: how about removing the code to L253?
if to L290, the pullsecrets will be processed repeatedly when there are more than one sidecar containers.

furykerry (approved):

PR #679, opened 2021-07-26
fix updated pod cannot inplace update again

### LINKED ISSUE #676: [BUG] Pod with Updated state cannot update/rollback
**What happened**:
Setting cloneset with inplace update lifecycle hook, when pod enter to Updated state, make cloneset rollback the image or update to new image, the pod with Updated state will skip to update.

can update Updated pod in the premise of not loss traffic

Create a cloneset with replicas 1, lifecycle hook, InPlaceIfPossible strategy. Update the cloneset image to some bad image(which will make pod not ready). When pod enter into Updated state, rollback cloneset, it will stuck.

**Environment**:
- Kruise version: v0.9.0
- Kubernetes version (use `kubectl version`): v1.19.3
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

shiyan2016:
fix updated pod cannot inplace update again

fixes #676

### DIFF
--- a/pkg/controller/cloneset/sync/cloneset_update.go
+++ b/pkg/controller/cloneset/sync/cloneset_update.go
@@ -86,9 +86,13 @@ func (c *realControl) Update(cs *appsv1alpha1.CloneSet,
 		}
 		if waitUpdate {
 			switch lifecycle.GetPodLifecycleState(pod) {
-			case appspub.LifecycleStatePreparingDelete, appspub.LifecycleStateUpdated:
+			case appspub.LifecycleStatePreparingDelete:
 				klog.V(3).Infof("CloneSet %s/%s find pod %s in state %s, so skip to update it",
 					cs.Namespace, cs.Name, pod.Name, lifecycle.GetPodLifecycleState(pod))
+			case appspub.LifecycleStateUpdated:
+				klog.V(3).Infof("CloneSet %s/%s find pod %s in state %s but not in updated revision",
+					cs.Namespace, cs.Name, pod.Name, appspub.LifecycleStateUpdated)
+				canUpdate = true
 			default:
 				if gracePeriod, _ := appspub.GetInPlaceUpdateGrace(pod); gracePeriod != "" {
 					klog.V(3).Infof("CloneSet %s/%s find pod %s still in grace period %s, so skip to update it",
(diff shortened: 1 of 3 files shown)


pkg/controller/cloneset/sync/cloneset_update.go
```diff
@@ -187,16 +196,19 @@ func (c *realControl) updatePod(cs *appsv1alpha1.CloneSet, coreControl clonesetc
...
 			switch state := lifecycle.GetPodLifecycleState(pod); state {
-			case "", appspub.LifecycleStateNormal:
-				var err error
-				var updated bool
-				if cs.Spec.Lifecycle != nil && lifecycle.IsPodHooked(cs.Spec.Lifecycle.InPlaceUpdate, pod) {
-					if updated, err = c.lifecycleControl.UpdatePodLifecycle(pod, appspub.LifecycleStatePreparingUpdate); err == nil && updated {
-						clonesetutils.ResourceVersionExpectations.Expect(pod)
-						klog.V(3).Infof("CloneSet %s update pod %s lifecycle to PreparingUpdate",
-							clonesetutils.GetControllerKey(cs), pod.Name)
+			case "", appspub.LifecycleStateNormal, appspub.LifecycleStateUpdated:
```
FillZpp, line 199: How about keep the `appspub.LifecycleStateNormal` logic, add another `appspub.LifecycleStateUpdated`:

```golang
    case "", appspub.LifecycleStateNormal:
        // ...
    case appspub.LifecycleStateUpdated:
        var inPlaceUpdateHandler *appspub.LifecycleHook
        if cs.Spec.Lifecycle != nil {
            inPlaceUpdateHandler = cs.Spec.Lifecycle. InPlaceUpdate
        }
	if updated, err = c.lifecycleControl.UpdatePodLifecycleWithHandler(pod, appspub.LifecycleStatePreparingUpdate, inPlaceUpdateHandler); err == nil && updated {
		// ...
	}
	return 0, err
```

The `UpdatePodLifecycleWithHandler` function should patch pod with not only the lifecycle state, but also the finalizers and labels in the given handler.

This will also solve the problem in this PR that the pod object is changed in `SetPodLifecycleHook` but nowhere to update it to apiserver.

pkg/controller/cloneset/sync/cloneset_update.go
```diff
@@ -86,9 +86,18 @@ func (c *realControl) Update(cs *appsv1alpha1.CloneSet,
 		}
 		if waitUpdate {
 			switch lifecycle.GetPodLifecycleState(pod) {
-			case appspub.LifecycleStatePreparingDelete, appspub.LifecycleStateUpdated:
+			case appspub.LifecycleStatePreparingDelete:
 				klog.V(3).Infof("CloneSet %s/%s find pod %s in state %s, so skip to update it",
 					cs.Namespace, cs.Name, pod.Name, lifecycle.GetPodLifecycleState(pod))
+			case appspub.LifecycleStateUpdated:
+				if clonesetutils.EqualToRevisionHash("", pod, updateRevision.Name) {
```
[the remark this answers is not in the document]
shiyan2016: done

pkg/controller/cloneset/sync/cloneset_update.go
shiyan2016: done

pkg/util/lifecycle/lifecycle_utils.go
```diff
@@ -96,6 +98,52 @@ func (c *realControl) UpdatePodLifecycle(pod *v1.Pod, state appspub.LifecycleSta
 	return true, err
 }
 
+func (c *realControl) UpdatePodLifecycleWithHandler(pod *v1.Pod, state appspub.LifecycleStateType, inPlaceUpdateHandler *appspub.LifecycleHook) (bool, error) {
+	if inPlaceUpdateHandler == nil || pod == nil {
```
FillZpp, line 102: if `inPlaceUpdateHandler` is nil, maybe we should turn to the `UpdatePodLifecycle` instead of directly return?

PR #696, opened 2021-08-06
fix sidecar name changed panic issue

zmberg:

Fix the kruise exception caused by the change of container name in sidecarSet.

When sidecarSet container namge changed, then sidecarSet controller upgrading pods logic will be exception. The controller determines if it is the same container in pod by container name.
image

The fix idea is as follows:
1. SidecarSet controller determine whether sidecar name in pod and sidecar name in sidecarSet are the same, if not, then do not perform the upgrade operation.


Fei-Guo:
It is unadvised to send a ~300 lines PR without any descriptions. Please add adequate description as template suggests.

PR #699, opened 2021-08-09
Fix the kruise exception caused by the change of container name in sidecarSet.

zmberg:

Fix the kruise exception caused by the change of container name in sidecarSet.

When sidecarSet container namge changed, then sidecarSet controller upgrading pods logic will be exception. The controller determines if it is the same container in pod by container name.
A8EA35C5-5C9B-435D-ABAF-CAFA41F3B80F

The fix idea is as follows:
1. pod annotations[kruise.io/sidecarset-hash] record injection of sidecar container list.
2. SidecarSet controller determine whether sidecar name in pod and sidecar name in sidecarSet are the same, if not, then do not perform the upgrade operation. 
3. The community version function **IsSidecarSetUpgradable** does not need to be adjusted, because it already determines whether kruise.io/sidecarset-hash-without-image is changed or not.


furykerry (commented):
/lgtm

PR #700, opened 2021-08-09
Add proposal for ContainerPriority

FillZpp:

Add proposal for ContainerLaunchPriority, which controls the sequence of containers start in a Pod.

Read link: https://github.com/FillZpp/kruise/blob/add-proposal-for-containers-starting-priority/docs/proposals/20210809-containerpriority.md

### DIFF
--- a/docs/proposals/20210809-containerlaunchpriority.md
+++ b/docs/proposals/20210809-containerlaunchpriority.md
@@ -1,0 +1,140 @@
+---
+title: ContainerPriority
+authors:
+  - "@FillZpp"
+reviewers:
+  - "@Fei-Guo"
+  - "@furykerry"
+creation-date: 2021-08-09
+last-updated: 2021-08-09
+status: implementable
+---
+
+# ContainerLaunchPriority
+
+Provide a way to help users control the sequence of containers launch in a Pod.
+
+## Table of Contents
+
+A table of contents is helpful for quickly jumping to sections of a proposal and for highlighting
+any additional information provided beyond the standard proposal template.
+[Tools for generating](https://github.com/ekalinin/github-markdown-toc) a table of contents from markdown are available.
+
+- [ContainerPriority](#containerpriority)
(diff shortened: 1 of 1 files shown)


docs/proposals/20210809-containerpriority.md
```diff
@@ -0,0 +1,126 @@
...
+      value: "1"
+    - name: KRUISE_CONTAINER_BARRIER
+      valueFrom:
+        configMapKeyRef:
+          name: {pod-name}-barrier
+          key: "c_1"   # its KRUISE_CONTAINER_PRIORITY is 1
+    # ...
+```
+
+*Note that the key of containers will be `c_{priority}` and the key of init containers will be `i_{priority}`.*
```
furykerry, line 107: the start sequence of init container is already deterministic, i don't think it is necessary to mess with the init containers

docs/proposals/20210809-containerpriority.md
```diff
@@ -0,0 +1,126 @@
...
+    # ...
+    env:
+    - name: KRUISE_CONTAINER_BARRIER
+      valueFrom:
+        configMapKeyRef:
+          name: {pod-name}-barrier
+          key: "c_0"   # no KRUISE_CONTAINER_PRIORITY, so defaults to 0
+  - name: sidecar
+    env:
+    - name: KRUISE_CONTAINER_PRIORITY
```
furykerry, line 97: is it necessary to include integer value priority in the environment? is a simple boolean value such as KRUISE_CONTAINER_STRONG_ORDER suffice ? the containers list of the pod spec is already ordered, and kubelet will start the containers in order, if the orders in containers list is conflicts with the one specified by environment variable , the orders can be hard to reason about.

furykerry (changes requested):

docs/proposals/20210809-containerpriority.md
FillZpp: Removed, don't need to deal with init containers.

PR #712, opened 2021-08-17
fix: ds status is outdate, apiserver will refuse this request

Noisyes:
Signed-off-by: Noisyes <kyf19970408@163.com>

It's easy to get update conflict when processing a DaemonSet continuously. storeDaemonSetStatus had a retry logic that it gets the resource from api server to update the DaemonSet with the latest resource. However, version.daemonset.metadata.resourceVersion is outdate, api server will  refuse this request and always made an useless get call.

little fix

### DIFF
--- a/pkg/controller/daemonset/daemonset_util.go
+++ b/pkg/controller/daemonset/daemonset_util.go
@@ -301,7 +301,6 @@ func storeDaemonSetStatus(dsClient kubeClient.Client, ds *appsv1alpha1.DaemonSet
 		}
 
 		klog.Errorf("update DaemonSet status %v failed: %v", ds.Status, updateErr)
-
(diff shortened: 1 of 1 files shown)


pkg/controller/daemonset/daemonset_util.go
```diff
@@ -282,7 +282,7 @@ func storeDaemonSetStatus(dsClient kubeClient.Client, ds *appsv1alpha1.DaemonSet
 	klog.V(6).Infof("toUpdate is %v", ds)
 
 	var updateErr, getErr error
-	for i := 0; i < StatusUpdateRetries; i++ {
+	for i := 0; ; i++ {
```
Fei-Guo, line 285: I suggest to remove the StatusUpdateRetries and use the client-go retry package. i.e., the retry.RetryOnConflict method.

PR #722, opened 2021-08-25
fix(daemon): daemon logs nothing when get kubeconfig fails

yttan:
SetLogger should be called before GetConfigOrDie

When daemon fails to get kubeconfig, the pod turns to Error and there is no log.

According to https://github.com/kubernetes-sigs/controller-runtime/issues/1475 , SetLogger should be called before GetConfigOrDie.


FillZpp (commented):
/approve

PR #734, opened 2021-09-02
Optimize workloadspread feature-gate and object decode for pod deletion

FillZpp:

Optimize workloadspread feature-gate and object decode for pod deletion.

/cc @BoltsLei


furykerry (commented):
/lgtm

PR #742, opened 2021-09-07
fix(sidecarset): filter active pods with kubecontroller.IsPodActive

### LINKED ISSUE #741: [BUG] sidecarset update not work when Evicted pod exists
**What happened**:
After some Pods selected by sidecarset were evicted by kubelet, updating sidecarset won't update all selected pods.

Sidecarset update should work for only normal Pods but evicted Pods.

1. apply the following manifests (Pods in example will be evicted in minutes) 
2. update sidecarset image form 1.17 to 1.16
3. observe the sidecarset and Pods

sidecarset

pods

Here is the sidecarset and deployment example:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
spec:
  replicas: 10
  selector:
    matchLabels:
      app: test
  template:
    metadata:
      name: test
      labels:
        app: test
        sidecarset: test
    spec:
      containers:
       - name: test
          image: registry.cn-shanghai.aliyuncs.com/docker_mirror_image/random-logger:pypypy
          resources:
            limits:
              ephemeral-storage: 1Mi
---
kind: SidecarSet
metadata:
  name: test-sidecarset
spec:
  selector:
    matchLabels:
      sidecarset: test
  updateStrategy:
    type: RollingUpdate
    maxUnavailable: 1
  containers:
    - name: sidecar
      image: nginx:1.16
```
The test container in deployment will print log to file, when the ephemeral-storage exceed the limit, the Pod will be evicted.
The hostname hash determines whether log or not, **this will only make part of Pods evicted**.
```python
import hashlib
import socket
import time

hostname = socket.gethostname()
hostname_hash = bin(int(hashlib.md5(hostname.encode('utf-8')).hexdigest(), 16))
need_log_to_file = hostname_hash[-1] == "1"

def log():
    s = "a"
    i = 1
    while True:
        if need_log_to_file:
            print("log to file")
            with open("a.log", "a") as f:
                f.write("{}\n".format(s*i))
                i += 10
                time.sleep(0.1)
        else:
            print("no need log to file")
            time.sleep(1)
if __name__ == '__main__':
    log()
```

The problem is related to the `.Spec.UpdateStrategy.MaxUnavailable`. Because of updated evicted Pods will never be ready,  the upgradeAndNotReadyCount will always great than MaxUnavailable, so some Pods will never be updated.
https://github.com/openkruise/kruise/blob/34c1e98ad70e0b5d8b0a18815bc6c02d8cb72ea8/pkg/controller/sidecarset/sidecarset_strategy.go#L153

Refer to the replicaset controller, they filter the active Pods by the following func.
```go
func IsPodActive(p *v1.Pod) bool {
    return v1.PodSucceeded != p.Status.Phase &&
        v1.PodFailed != p.Status.Phase &&
	p.DeletionTimestamp == nil
}
```
So I think the IsActivePod in kruise should be improved.
https://github.com/openkruise/kruise/blob/8b4f7d5af23b8ca965fc633f728fad7ec26121e9/pkg/control/sidecarcontrol/util.go#L85

**Environment**:
- Kruise version: 0.8.1
- Kubernetes version (use `kubectl version`): v1.17.12
- OS (e.g: `cat /etc/os-release`): Ubuntu 18.04.5 LTS
- Kernel (e.g. `uname -a`): 5.4.0-73-generic
- Install tools: helm charts
- Others:

ciiiii:
Improve the IsActivePod logic in sidecarset controller and fix #741.

#741 

follow the reproduce  procedure in issue.

### DIFF
--- a/pkg/control/sidecarcontrol/util.go
+++ b/pkg/control/sidecarcontrol/util.go
@@ -88,10 +88,7 @@ func IsActivePod(pod *corev1.Pod) bool {
 			return false
 		}
 	}
-	if pod.ObjectMeta.GetDeletionTimestamp() != nil {
-		return false
-	}
-	return true
+	return kubecontroller.IsPodActive(pod)
 }
 
 func GetSidecarSetRevision(sidecarSet *appsv1alpha1.SidecarSet) string {


zmberg:
It's good. 
/approve

FillZpp (approved):
Thanks for fixing this.

/approve

PR #748, opened 2021-09-10
fix unpredictable numbers of pods when interval of reserveOrdinals > …

### LINKED ISSUE #746: [BUG] unpredictable numbers of pods when interval of reserveOrdinals > ordinals of pods
**What happened**:
    unpredictable numbers of pods when interval of reserveOrdinals > ordinals of pods

    when interval of reserveOrdinals > ordinals of pods, if delete item of reserveOrdinals, don`t create pods.
    
    for example

     step1: create a statefulset with 3 replicas ---> [pod-0, pod-1, pod-2]
     step2: update statefulset.Spec.ReserveOrdinals to [3, 4, 5]
     step3: update statefulset.Spec.ReserveOrdinals to [4, 5]

     After the above operation， the pods will become [pod-0, pod-1, pod-2, pod-3], but Spec.Replicas is 3.
    
     

**Environment**:
- Kruise version: v0.9.0
- Kubernetes version (use `kubectl version`): v1.19.3

vegazmaoning:
…ordinals of pods

  don`t create pods when interval of reserveOrdinals > max(pod.ordinal) and replicas stay the same

fixed #746

### DIFF
--- a/pkg/controller/statefulset/stateful_set_control.go
+++ b/pkg/controller/statefulset/stateful_set_control.go
@@ -479,7 +479,10 @@ func (ssc *defaultStatefulSetControl) updateStatefulSet(
 				i, replicas)
 		}
 		// If we find a Pod that has not been created we create the Pod
-		if !isCreated(replicas[i]) {
+		// 1. If `condemned` has item represent the reverseOrdinals has changed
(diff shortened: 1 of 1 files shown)


FillZpp:
/hold

We should firstly discuss this in the issue #746

PR #751, opened 2021-09-13
Fix e2e bug for workloadspread

BoltsLei:
Signed-off-by: bolstlei <1606880753@qq.com>

We should wait for the controller to complete a reconcile when the workloadspread is updated.
image


FillZpp (approved):
/lgtm

PR #752, opened 2021-09-13
Support WorkloadSpread to manage the existing pods

veophi:

Support workloadspread to manage the existing pods,  which were created before the workloadspread.
**WorkloadSpread will count these existing pods for each subset when reconciling, and will no longer ignore them.**

### Ⅱ. Describe how to verify it
Unit test and E2E test.

### Ⅲ. Special notes for reviews
**For those existing pods that do not match any subset of the workloadspread, their deletion cost will be zero.**


pkg/controller/workloadspread/workloadspread_controller.go
```diff
@@ -327,16 +328,49 @@ func groupPod(ws *appsv1alpha1.WorkloadSpread, pods []*corev1.Pod) map[string][]
...
+
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, err
+	} else if err == nil && node != nil {
+		for i := range ws.Spec.Subsets {
+			subset := &ws.Spec.Subsets[i]
+			// in case of that this pod was scheduled to the node which matches a subset of workloadSpread
+			if matched, err := nodeMatchesSubset(node, subset); err != nil {
+				return nil, err
+			} else if matched {
```
furykerry, line 365: shall we also check the subset patch ? One can have subset that only have patch but without any affinity or toleration changes. maybe we should leave it as a TODO

pkg/controller/workloadspread/workloadspread_controller.go
```diff
@@ -327,16 +328,49 @@ func groupPod(ws *appsv1alpha1.WorkloadSpread, pods []*corev1.Pod) map[string][]
...
+func (r *ReconcileWorkloadSpread) getSuitableSubsetForOldPod(ws *appsv1alpha1.WorkloadSpread, pod *corev1.Pod) (*appsv1alpha1.WorkloadSpreadSubset, error) {
+	var err error
+	var node *corev1.Node
+
+	if len(pod.Spec.NodeName) > 0 {
+		node = &corev1.Node{}
+		err = r.Get(context.TODO(), types.NamespacedName{Name: pod.Spec.NodeName}, node)
+	}
+
+	if err != nil && !errors.IsNotFound(err) {
```
furykerry, line 357: consider move the subset iteration out of the if clause to make the main logic clear. 
```
  if len(pod.Spec.NodeName) == 0 {
      return nil, nil
   }
  if err != nil {
        if errors.IsNotFound(err) {
            return nil, nil 
        }
        return nil, err
   }
  for i := range ws.Spec.Subsets {
  }

```

pkg/controller/workloadspread/workloadspread_controller_test.go
```diff
@@ -1620,6 +1620,201 @@ func TestDelayReconcile(t *testing.T) {
...
+		name                 string
+		getPods              func() []*corev1.Pod
+		getNodes             func() []*corev1.Node
+		getCloneSet          func() *appsv1alpha1.CloneSet
+		getWorkloadSpread    func() *appsv1alpha1.WorkloadSpread
+		expectWorkloadSpread func() *appsv1alpha1.WorkloadSpread
+	}{
+		{
+			name: "manager the pods that were created before workloadSpread",
+			getPods: func() []*corev1.Pod {
```
furykerry, line 1634: add a case for existing pods with injected workload spread annotation, but the spread in annotation cannot match the pod again

pkg/controller/workloadspread/workloadspread_controller.go
veophi: done

pkg/controller/workloadspread/workloadspread_controller.go
veophi: done

pkg/controller/workloadspread/workloadspread_controller_test.go
veophi: done

pkg/controller/workloadspread/workloadspread_controller.go
veophi: done

pkg/controller/workloadspread/workloadspread_controller.go
veophi: I left it as a TODO

pkg/controller/workloadspread/workloadspread_controller_test.go
veophi: done

pkg/controller/workloadspread/workloadspread_controller.go
```diff
@@ -318,23 +322,81 @@ func getInjectWorkloadSpreadFromPod(pod *corev1.Pod) *wsutil.InjectWorkloadSprea
...
+				subset.Name, ws.Namespace, ws.Name, err.Error())
+		} else if matched {
+			// add MatchedWorkloadSpreadSubsetAnnotations for this pod based on the matched subset
+			// the purpose of this operation is to avoid to do this match again in the next reconcile
+			by, _ := json.Marshal(&wsutil.InjectWorkloadSpread{Name: ws.Name, Subset: subset.Name})
+			if pod.Annotations == nil {
+				pod.Annotations = make(map[string]string)
+			}
+			pod.Annotations[wsutil.MatchedWorkloadSpreadSubsetAnnotations] = string(by)
+			if err := r.Update(context.TODO(), pod); err != nil {
```
FillZpp, line 389: You'd better use `patch` instead of `update` to avoid conflict.

veophi: done

pkg/controller/workloadspread/workloadspread_controller_utils.go
```diff
@@ -78,3 +80,14 @@ func filterOutCondition(conditions []appsv1alpha1.WorkloadSpreadSubsetCondition,
 	}
 	return newConditions
 }
+
+func nodeMatchesSubset(node *corev1.Node, subset *appsv1alpha1.WorkloadSpreadSubset) (bool, error) {
```
veophi, line 84: Here, should we first merge the affinity and tolerations of pod with subset, and then match the node? @FillZpp @furykerry

pkg/util/workloadspread/workloadspread.go
```diff
@@ -252,79 +252,84 @@ func (h *Handler) mutatingPod(matchedWS *appsv1alpha1.WorkloadSpread,
...
-		var err error
-		start := time.Now()
-		if refresh {
-			// TODO: shall we set metav1.GetOptions{resourceVersion="0"} so that we get the cached object in apiServer memory instead of etcd?
-			wsClone, err = kubeClient.GetGenericClient().KruiseClient.AppsV1alpha1().
-				WorkloadSpreads(matchedWS.Namespace).Get(context.TODO(), matchedWS.Name, metav1.GetOptions{})
-			if err != nil {
-				if errors.IsNotFound(err) {
-					return nil
+	if len(matchedWS.Spec.Subsets) == 1 &&
```
veophi, line 255: Add line 255-259

PR #762, opened 2021-09-23
export sync-seconds to control resource resync frequency

shiyan2016:
export sync-seconds to control resource resync frequency

### DIFF
--- a/main.go
+++ b/main.go
@@ -75,6 +75,7 @@ func main() {
 	var enableLeaderElection, enablePprof, allowPrivileged bool
 	var leaderElectionNamespace string
 	var namespace string
+	var syncPeriodStr string
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&healthProbeAddr, "health-probe-addr", ":8000", "The address the healthz/readyz endpoint binds to.")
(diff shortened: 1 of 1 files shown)


shiyan2016:
@dco

main.go
```diff
@@ -129,6 +132,7 @@ func main() {
 		LeaderElectionNamespace:    leaderElectionNamespace,
 		LeaderElectionResourceLock: resourcelock.ConfigMapsResourceLock,
 		Namespace:                  namespace,
+		SyncPeriod:                 &syncPeriod,
```
FillZpp, line 135: The type of `SyncPeriod` is `*time.Duration`, which means the default value `36000` is `36us`...

How about:
```golang
        var syncPeriodStr string
	flag.StringVar(&syncPeriodStr, "sync-period", "", "Determines the minimum frequency at which watched resources are reconciled.")

	var syncPeriod *time.Duration
	if syncPeriodStr != "" {
		d, err := time.ParseDuration(syncPeriodStr)
		if err != nil {
			setupLog.Error(err, "invalid sync period flag")
		}
		syncPeriod = &d
	}

            // ...
            SyncPeriod:                 syncPeriod,
```

Then you can set the flag as `--sync-perioid=10h` or `--sync-perioid=0` to disable resync.

shiyan2016: done

FillZpp (approved):
/lgtm

PR #797, opened 2021-11-09
move manager main.go to cmd/manager/main.go and update dockerfile,mak…

flyaways:
…efile


FillZpp:
@flyaways I'm afraid we can not move main.go into cmd/, for it is the main file that generated by kubebuilder. Moving it might cause failure of subsequent `kubebuild create api ...` command.

flyaways:
thanks for your reply @FillZpp

PR #799, opened 2021-11-10
Add featureGate to enable restartPolicy onFailure for CloneSet

### LINKED ISSUE #794: [BUG] CloneSet validating webhook requests pod with restartPolicy `onFailure` even though it is supported
**What happened**:
We are deploying a workload using CloneSet with restartPolicy as onFailure.
because one of the colocating containers is not a long running service and it needs to exit after its job is done.

We configured the validating webhook cloneset (` vcloneset.kb.io`) failurePolicy as `Ignore`.

However, the pod request was rejected given the error: 

```
"admission webhook \"vcloneset.kb.io\" denied the request: spec.template.spec.restartPolicy: Unsupported value: \"OnFailure\": supported values: \"Always\""}
```

Then we somehow disabled the webhook completely (hacks such as giving the webhook a wrong cert in the field `caBundle `, controller-manager happily proceeds but just the webhook will not take effect), and we see our workloads deployed successfully with the expected behavior (no restarting the colocating container in the pod if exited successfully)

I noticed something in the code base specified the webhook to fail the request once the restartPolicy is not Always.

We expect that
1. when the webhook failurePolicy says Ignore, then just don't fail the request when the pod restart policy is not Always. 
2. The onFailure is expected to behave, why adding the guards such as  [here](https://github.com/openkruise/kruise/blob/master/pkg/webhook/cloneset/validating/validation.go#L79)? Can we remove it?

Just create the Cloneset with a pod restartPolicy as onFailure

FillZpp:

Add featureGate to enable restartPolicy onFailure for CloneSet

fixes #794

### DIFF
--- a/pkg/features/kruise_features.go
+++ b/pkg/features/kruise_features.go
@@ -79,6 +79,9 @@ const (
 	// InPlaceUpdateEnvFromMetadata enables Kruise to in-place update a container in Pod
 	// when its env from labels/annotations changed and pod is in-place updating.
 	InPlaceUpdateEnvFromMetadata featuregate.Feature = "InPlaceUpdateEnvFromMetadata"
+
+	// EnableCloneSetRestartPolicyOnFailure enables webhook to allow CloneSet with restartPolicy onFailure.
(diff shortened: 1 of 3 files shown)


FillZpp:
closed by #794

PR #812, opened 2021-11-18
UnitedDeployment subset add patch filed change workload template

### LINKED ISSUE #811: [feature request] UnitedDeployment support patch workload by subset
In UnitedDeployment, allow subset patch workload template, can customize the workload by subset.

**Why is this needed**:
We need manage multiple deployment or statefulset, most of their configurations are the same, but some environment variables or image is different, so UnitedDeployment need patch workload feature by subset.

wxyh:
In UnitedDeployment, subset add patch field, can customize each workload.

Fixes #811 

Add test subsetWithPatch.

Deploy UnitedDeployment, change subset.patch filed, observe changes of workload.


pkg/controller/uniteddeployment/adapter/cloneset_adapter.go
```diff
@@ -135,6 +139,35 @@ func (a *CloneSetAdapter) ApplySubsetTemplate(ud *alpha1.UnitedDeployment, subse
 	attachNodeAffinity(&set.Spec.Template.Spec, subSetConfig)
 	attachTolerations(&set.Spec.Template.Spec, subSetConfig)
 
+	if len(subSetConfig.Patch.Raw) != 0 {
+		cloneBytes, _ := json.Marshal(set)
+		modified, err := strategicpatch.StrategicMergePatch(cloneBytes, subSetConfig.Patch.Raw, &alpha1.StatefulSet{})
```
FillZpp, line 144: Have you test it in a real cluster? (or you can write an e2e case in `test/e2e/apps`) I'm not sure whether you can use `strategicpatch.StrategicMergePatch` for cloneset and advanced statefulset, since they are defined by CRD.

pkg/controller/uniteddeployment/adapter/advanced_statefulset_adapter.go
```diff
@@ -172,6 +176,35 @@ func (a *AdvancedStatefulSetAdapter) ApplySubsetTemplate(ud *alpha1.UnitedDeploy
...
+			return fmt.Errorf("subset[%s] patch changed labels, mismatch UnitedDeployment selector ", subsetName)
+		}
+
+		patchedSet.DeepCopyInto(set)
+		klog.V(2).Infof("AdvancedStatefulSet [%s/%s] has patches configure successfully: %s", set.Namespace, set.GenerateName, subSetConfig.Patch.Raw)
+	}
+	if set.Annotations == nil {
+		set.Annotations = make(map[string]string)
+	}
+	set.Annotations[alpha1.AnnotationSubsetPatchKey] = string(subSetConfig.Patch.Raw)
```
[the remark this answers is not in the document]
wxyh: If the patch field is now empty and the field had a value before, need update annotations

pkg/controller/uniteddeployment/adapter/cloneset_adapter.go
wxyh: I have tested and I add test in `pkg/controller/uniteddeployment`. Do you mean need to  add `uniteddeployment.go `in `test/e2e/apps `?

FillZpp: Yes, add an e2e case so we can test it in a real k8s cluster.

shijieqin: @wxyh Hi boy, can you write an e2e case in "test/e2e/apps" as FillZpp said? I really like the feature you submitted, and I want to use it in the upcoming new version of Kruise, otherwise I would need to resubmit the pr myself.

wxyh: @shijieqin OK. I'll add it in the next few days

wxyh: @FillZpp I found some problems in the test.
1.  For `cloneSet` and `advancedStatefulSet`, some fields such as `ScaleStrategy`,`MinReadySeconds`,`Lifecycle`  don't assign value from `CloneSetTemplate` and `AdvancedStatefulSetTemplate` .Is this a mistake? should I fix it in another pr?
2. `strategicpatch.StrategicMergePatch` can be used in `cloneSet` and `advancedStatefulSet`.But for list like `VolumeClaimTemplates`  patch strategy is `replace` because `patchStrategy` is not defined. I think that's acceptable,what do you think ?

perek:
checking in on this PR, is this still reasonable to add? This is very useful for many different usecases

PR #815, opened 2021-11-19
feat: create daemon pod according to progressive flag in manage phase…

jetmuffin:
Signed-off-by: jetmuffin<jeffchen328@gmail.com>

This PR fixs 2 problems:

1. DaemonSet controller never deletes failed pod.

Daemonset-controller figures out the pods to be created and deleted on given nodes in podsShouldBeOnNodes. During this process, if any daemon pod failed, daemonset-controller push it into podsToDelete slice and set a backoff key for this pod, in order to delete it and recreate a new pod in the next sync loop.

But in manage func, daemonset-controller gets the backoff time, simply returns with RequeueAfter option and never deletes this failed pod.

```golang
	delay, err := dsc.manage(ds, hash)
	if err != nil {
	}
	if delay != 0 {
		return reconcile.Result{RequeueAfter: delay}, nil
	}
```

Further, daemonset-controller founds this failed pod in next sync loop, increases the backoff time and returns once again which blocks the upgrading. If there are some failed pods in cluster, all worker will be blocked due to the backoff and no upgrading continues.

2. Not all nodes runs daemon pod in rolling update.

During the upgration, the partition is set to a specified number so that only some of the replicas will be update to new version. If I delete some pod manually, new pod will not be created on that node because it is controlled by partition to progressively create pod in first deployment.

Add some test cases for calculating nodes that need pods in manage phase.

### DIFF
--- a/pkg/controller/daemonset/daemonset_controller.go
+++ b/pkg/controller/daemonset/daemonset_controller.go
@@ -96,6 +96,10 @@ const (
 	IsIgnoreNotReady     = "daemonset.kruise.io/ignore-not-ready"
 	IsIgnoreUnscheduable = "daemonset.kruise.io/ignore-unscheduable"
 
+	// ProgressiveCreatePod indicates daemon pods created in manage phase will be controlled by partition.
+	// This annotation will be added to DaemonSet when it is created, and removed if partition is set to 0.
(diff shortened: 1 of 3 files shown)


FillZpp (approved):
/lgtm

PR #820, opened 2021-11-25
Mark revision stable when all pods updated to it

FillZpp:

Consider the update revision as stable if revisions of all pods are consistent to it, no need to wait all of them ready.


furykerry (commented):
/lgtm

PR #821, opened 2021-11-25
add ephemeral job finalizer

hellolijj:

when ephemeral job deleted,  the ephemeralcontainer which owns will be removed.

### DIFF
--- a/pkg/controller/ephemeraljob/ephemeraljob_controller.go
+++ b/pkg/controller/ephemeraljob/ephemeraljob_controller.go
@@ -49,6 +49,8 @@ var (
 	scaleExpectations    = expectations.NewScaleExpectations()
 )
 
+const EphemeralContainerFinalizer = "apps.kruise.io/ephemeralcontainers-cleanup"
+
 // Add creates a new ImagePullJob Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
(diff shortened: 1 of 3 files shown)


pkg/controller/ephemeraljob/ephemeraljob_controller.go
```diff
@@ -49,6 +49,8 @@ var (
 	scaleExpectations    = expectations.NewScaleExpectations()
 )
 
+const EphemeralContainerFinalizer = "apps.kruise.io/delete-ephemeralcontainers"
```
FillZpp, line 52: `const EphemeralContainerCleanupFinalizer = "apps.kruise.io/ephemeralcontainers-cleanup"`

pkg/controller/ephemeraljob/ephemeraljob_controller.go
```diff
@@ -375,3 +388,20 @@ func (r *ReconcileEphemeralJob) updateJobStatus(job *appsv1alpha1.EphemeralJob)
 	klog.V(5).Infof("Updating job %s status %#v", job.Name, job.Status)
 	return r.Status().Update(context.TODO(), job)
 }
+
+func (r *ReconcileEphemeralJob) removeEphemeralContainers(job *appsv1alpha1.EphemeralJob) error {
+	targetPods, err := r.filterPods(job)
```
FillZpp, line 393: `filterPods` will get all pods that **can add the ephemeral containers into it**. But you might have to get pods that **already added ephemeral containers by this job**.

pkg/controller/ephemeraljob/ephemeraljob_controller.go
```diff
@@ -375,3 +388,20 @@ func (r *ReconcileEphemeralJob) updateJobStatus(job *appsv1alpha1.EphemeralJob)
...
+	if err != nil {
+		klog.Errorf("Failed to get ephemeral job %s/%s related target pods: %v", job.Namespace, job.Name, err)
+		return err
+	}
+
+	var errors error
+	for _, pod := range targetPods {
+		if e := econtainer.New(job).RemoveEphemeralContainer(pod); e != nil {
```
FillZpp, line 401: `RemoveEphemeralContainer` should not get the single pod as parameter. If the pod has ephemeral containers that owned by multiple jobs, you should not remove them all in this place.

pkg/controller/ephemeraljob/ephemeraljob_controller.go
hellolijj: fixed

furykerry (commented):
/lgtm

PR #823, opened 2021-11-26
Make workloadSpread more robust and easily-used.

veophi:
Two things:
1. Relax webhook validating restriction about subset: No limit for name-only subset.
2. Improve the management for the pods that were created before workloadspread:
   - consider PreferredNodeSelectorTerms and Patch of subset when matching these pods and  subsets;
   - update the deletion cost of the pods that cannot match any subset to -100 * (len(subsets)+1), so that make sure these pods can be deleted preferentially.

Unit Test & E2E Test

### DIFF
--- a/pkg/controller/workloadspread/update_pod_deletion_cost.go
+++ b/pkg/controller/workloadspread/update_pod_deletion_cost.go
@@ -42,6 +42,13 @@ func (r *ReconcileWorkloadSpread) updateDeletionCost(ws *appsv1alpha1.WorkloadSp
 			return err
 		}
 	}
+	// update the deletion-cost annotation for such pods that do not match any real subsets.
+	// these pods will have the minimum deletion-cost, and will be deleted preferentially.
+	if len(podMap[FakeSubsetName]) > 0 {
+		if err := r.syncSubsetPodDeletionCost(ws, nil, len(ws.Spec.Subsets), podMap[FakeSubsetName], workloadReplicas); err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
@@ -83,7 +90,10 @@ func (r *ReconcileWorkloadSpread) syncSubsetPodDeletionCost(
 	replicas := len(activePods)
 
 	// First we partition Pods into two lists: positive, negative list.
-	if subset.MaxReplicas == nil {
+	if subset == nil {
+		// for the scene of FakeSubsetName, where the pods don't match any subset and will be deleted preferentially.
+		negativePods = activePods
+	} else if subset.MaxReplicas == nil {
 		// maxReplicas is nil, which means there is no limit to the number of Pods in this subset.
 		positivePods = activePods
 	} else {
@@ -116,11 +126,11 @@ func (r *ReconcileWorkloadSpread) syncSubsetPodDeletionCost(
 		}
 	}
 
-	err = r.updateDeletionCostForSubsetPods(ws, subset.Name, positivePods, strconv.Itoa(wsutil.PodDeletionCostPositive*(len(ws.Spec.Subsets)-subsetIndex)))
+	err = r.updateDeletionCostForSubsetPods(ws, subset, positivePods, strconv.Itoa(wsutil.PodDeletionCostPositive*(len(ws.Spec.Subsets)-subsetIndex)))
 	if err != nil {
 		return err
 	}
-	err = r.updateDeletionCostForSubsetPods(ws, subset.Name, negativePods, strconv.Itoa(wsutil.PodDeletionCostNegative*(subsetIndex+1)))
+	err = r.updateDeletionCostForSubsetPods(ws, subset, negativePods, strconv.Itoa(wsutil.PodDeletionCostNegative*(subsetIndex+1)))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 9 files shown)


pkg/controller/workloadspread/workloadspread_controller.go
```diff
@@ -391,35 +397,66 @@ func (r *ReconcileWorkloadSpread) getSuitableSubsetForOldPod(ws *appsv1alpha1.Wo
...
+	// patch MatchedWorkloadSpreadSubsetAnnotations to the pod;
+	// select labels/annotations form favoriteSubset.patch, and patch them to the pod;
+	if favoriteSubset != nil {
+		var patchLabels map[string]string
+		patchAnnotations := make(map[string]string)
+		patchMetadata := make(map[string]interface{})
+
+		// decode favoriteSubset.patch.raw and add their labels/annotations to the patch
+		if favoriteSubset.Patch.Raw != nil {
+			patchField := &corev1.Pod{}
```
FillZpp, line 428: I don't really agree with the patchField used here, for:

1. Now we get the `labels` and `annotations` from it, what if there are other fields like `finalizers`?
2. If users write some special semantics of strategic patch in it, such as `... {"foo": null}...` to remove a key-value in labels/annotations or `$deleteFromPrimitiveList/finalizers` to remove finalizers, they will be lost from the unmarshal.

IMO, we should unmarshal the raw into `map[string]interface{}`. Then you get/set the `metadata.annotations` from it or initialize the fields, and submit it as the patch body.

pkg/controller/workloadspread/workloadspread_controller.go
```diff
@@ -391,35 +397,66 @@ func (r *ReconcileWorkloadSpread) getSuitableSubsetForOldPod(ws *appsv1alpha1.Wo
...
+		if patchLabels != nil {
+			patchMetadata["labels"] = patchLabels
+		}
+		patch, _ := json.Marshal(map[string]interface{}{
+			"metadata": patchMetadata,
+		})
+		if err := r.Patch(context.TODO(), pod, client.RawPatch(types.StrategicMergePatchType, patch)); err != nil {
+			// no need to retry, this operation will be retried in the next reconcile.
+			// this failure would not affect the subsequent logic
+			klog.Errorf(`Failed to patch "matched-workloadspread:{Name: %s, Subset: %s} annotation" for pod (%s/%s), but will retry in the next reconcile, err: %s`,
```
FillZpp, line 455: I'm not sure. If we failed to patch the pod, should we return the favoriteSubset with no error? (which means the pod has no annotation that belongs to the subset, but controller consider it does)

1. How to ensure the workloadspread will be triggered reconcile after this time?
2. What if the pod be allocated to another subset in next reconcile?

pkg/controller/workloadspread/workloadspread_controller.go
veophi: 1. Do you think that the `labels` and `annotations` **should not** be patched to pod?
2. Thanks. I will fix it.

pkg/controller/workloadspread/workloadspread_controller.go
veophi: emmm.... you're right,  returning this error can indeed get rid out of some problems.

pkg/controller/workloadspread/workloadspread_controller.go
```diff
@@ -369,14 +375,14 @@ func (r *ReconcileWorkloadSpread) getSuitableSubsetNameForPod(ws *appsv1alpha1.W
 		if err != nil {
 			return "", err
 		} else if matchedSubset == nil {
-			return "", nil
+			return FakeSubsetName, nil
```
furykerry, line 378: plz update the func comments ahead, now it will not return "" if suitable subset is not found

pkg/controller/workloadspread/workloadspread_controller.go
```diff
@@ -354,6 +358,8 @@ func (r *ReconcileWorkloadSpread) groupPod(ws *appsv1alpha1.WorkloadSpread, pods
 
 		if _, exist := podMap[subsetName]; exist {
 			podMap[subsetName] = append(podMap[subsetName], pods[i])
+		} else {
```
furykerry, line 361: since getSuitableSubsetNameForPod will return FakeSubsetName if no subset is found, is this else clause nesssary?

pkg/webhook/workloadspread/validating/workloadspread_validation.go
```diff
@@ -137,10 +138,14 @@ func validateWorkloadSpreadSpec(obj *appsv1alpha1.WorkloadSpread, fldPath *field
 				spec.ScheduleStrategy.Adaptive.RescheduleCriticalSeconds, "the last subset must be not specified when using adaptive scheduleStrategy"))
 		}
 
+		allowedMaxSeconds := int32(math.MaxInt32)
+		if len(spec.Subsets) > 1 {
+			allowedMaxSeconds = int32(300-5) / int32(len(spec.Subsets)-1)
```
furykerry, line 143: what is the magic number means here (e.g. 300 and 5), consider replace them with meaningful const variables

pkg/controller/workloadspread/update_pod_deletion_cost.go
```diff
@@ -83,7 +90,9 @@ func (r *ReconcileWorkloadSpread) syncSubsetPodDeletionCost(
 	replicas := len(activePods)
 
 	// First we partition Pods into two lists: positive, negative list.
-	if subset.MaxReplicas == nil {
+	if subset == nil {
+		negativePods = activePods
```
furykerry, line 94: plz comment that we prefer to delete the pods without matching any workloadspread

pkg/controller/workloadspread/workloadspread_controller.go
veophi: This `else` is for the situation where the pods' subset was deleted.

pkg/controller/workloadspread/update_pod_deletion_cost.go
veophi: I have added the comments here:
```
func (r *ReconcileWorkloadSpread) updateDeletionCost(ws *appsv1alpha1.WorkloadSpread,
	podMap map[string][]*corev1.Pod,
	workloadReplicas int32) error {
         ... ....
	// update the deletion-cost annotation for such pods that do not match any real subsets.
	// these pods will have the minimum deletion-cost, and will be deleted preferentially.
	if len(podMap[FakeSubsetName]) > 0 {
		if err := r.syncSubsetPodDeletionCost(ws, nil, len(ws.Spec.Subsets), podMap[FakeSubsetName], workloadReplicas); err != nil {
			return err
		}
	}
	return nil
}
```

pkg/webhook/workloadspread/validating/workloadspread_validation.go
veophi: done

pkg/controller/workloadspread/workloadspread_controller.go
veophi: done

furykerry (commented):
/lgtm

Issue #825, opened 2021-11-29
[BUG]Sidecarset status is incorrect when UpdateStrategy type is NotUpdate


andrewshaoyu:
**What happened**:
Sidecarset status Ready Pods is incorrect,when Sidecarset UpdateStrategy type is NotUpdateSidecarSetStrategyType.
Only update Sidecarset status when sidecarset is updated or new Pod create.
**What you expected to happen**:
when pods update trigger sidecarset reconcile, then update sidecarset status.
**How to reproduce it (as minimally and precisely as possible)**:
https://github.com/openkruise/kruise/blob/b465e29452f3b1432b3a348c129e670591d8c798/pkg/controller/sidecarset/sidecarset_pod_event_handler.go#L68-L83

I find that pods update event will trigger sidecarset reconcile ,but when sidecarset UpdateStrategy type is NotUpdateSidecarSetStrategyType, it will be ignored.

Is this by design?

**Environment**:
- Kruise version:
v.0.10.0

FillZpp:
> I find that pods update event will trigger sidecarset reconcile ,but when sidecarset UpdateStrategy type is NotUpdateSidecarSetStrategyType, it will be ignored.

Yep, it seems to be a bug that `readyPods` might not be updated in this case. PTAL @zmberg

zmberg:
@andrewshaoyu thanks, you are right. Do you have time to fix this bug?

andrewshaoyu:
> @andrewshaoyu thanks, you are right. Do you have time to fix this bug?

Ok, I can try.

PR #832, opened 2021-12-06
fix sidecarset status update

andrewshaoyu:

#825

### DIFF
--- a/pkg/controller/sidecarset/sidecarset_pod_event_handler.go
+++ b/pkg/controller/sidecarset/sidecarset_pod_event_handler.go
@@ -78,9 +78,6 @@ func (p *enqueueRequestForPod) updatePod(q workqueue.RateLimitingInterface, old,
 		return
 	}
 	for _, sidecarSet := range matchedSidecarSets {
-		if sidecarSet.Spec.UpdateStrategy.Type == appsv1alpha1.NotUpdateSidecarSetStrategyType {
-			continue
-		}
 		var isChanged bool
 		var enqueueDelayTime time.Duration
 		//check whether pod status is changed


FillZpp:
@andrewshaoyu Thanks for fixing it. The DCO check failed, which means you execute git commit without -s to sign yourself.

So you might have to `git commit --amend -s` and then force push again.

FillZpp:
/lgtm
PTAL @zmberg

FillZpp:
@zmberg You may have to apply `/approve` to merge it.

andrewshaoyu:
What's wrong？

zmberg:
@andrewshaoyu 
Because sidecarSet will manage hundreds of thousands of pods under some scenarios, for examples: logtail container. if every Pod ready changes will result in the overall calculation of all the pods, which will bring great pressure to the controller. And in the NotUpdate type, the user should not actually depend on the number of ready pods. So, we're going to have to think about this a little bit.

andrewshaoyu:
That means NotUpdate type is usually used when sidecars do not affect pod status like log container?Now I use sidecarset to inject envoy with NotUpdate type, I do not want to rolling update Pods.

To reduce update presure, I think we can try to compress event with a period, within the period, only trigger sidecarset status update once. 

And add a field statusUpdateStrategy, that indicate whether the sidecarset status need to update.

zmberg:
> And add a field statusUpdateStrategy, that indicate whether the sidecarset status need to update.
@FillZpp @furykerry What do you think, can add a type OnlyUpdateStatus to satisfy the user's scenario? For NotUpdate, sidecarSet status is no longer evaluated and triggered.

FillZpp:
> What do you think, can add a type OnlyUpdateStatus to satisfy the user's scenario? For NotUpdate, sidecarSet status is no longer evaluated and triggered.

I don't think `OnlyUpdateStatus` is a good design, for fields like `matchedPods` in sidecarset status should be correct event if it's `NotUpdate` type. What we concern is that handling events triggered by pod status changed to update `readyPods` and `updatedReadyPods` might increase controller pressure.

Compress event with a period may be an optional way, you can add the sidecarset caused by pod status changed into queue with a duration. But question is how long the duration should be and status delay could you accept? @andrewshaoyu

andrewshaoyu:
Make it configurable , default value 10s?

PR #840, opened 2021-12-08
optimize daemonset controller

hellolijj:
Signed-off-by: JunjunLi <junjunli666@gmail.com>

optimize daemonset controller


FillZpp (changes requested):
Can not simply remove `Get Node` and `CanNodeBeDeployed`.

/hold

Issue #774, opened 2021-10-15
[feature request] PodUnavailableBudget should also validate creation of ContainerRecreateRequest


FillZpp:
**What would you like to be added**:
PodUnavailableBudget should also validate creation of ContainerRecreateRequest.

**Why is this needed**:
ContainerRecreateRequest is used for recreating(restarting) containers in Pods, which also causes unavailable. So it should be validated by PodUnavailableBudget.

newhans:
I am very interested in this feature and I will try to solve it soon, can i?
support an PubOperation = "Recreate"

furykerry:
welcome, you can submit a KEP first to describe the technical consideration about this feature

newhans:
只修改crr validate webhhok即可，在创建crr时去校验一下，校验逻辑可以复用pubcontrol.PodUnavailableBudgetValidatePod。

newhans:
> welcome, you can submit a KEP first to describe the technical consideration about this feature
only update crr validate webhook can implement this feature, if there are no other problems, I will dive into code development

PR #842, opened 2021-12-10
Fix/validate crr for pub

hantmac:
PodUnavailableBudget should also validate creation of ContainerRecreateRequest.

Fix https://github.com/openkruise/kruise/issues/774

kruise/pkg/webhook/containerrecreaterequest/validating/validation_test.go


pkg/webhook/containerrecreaterequest/validating/crr_create_update_handler.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+var _ admission.Handler = &ContainerRecreateRequestHandler{}
+
+func (h *ContainerRecreateRequestHandler) validatingContainerRecreateRequestFn(ctx context.Context, req admission.Request) (allowed bool, reason string, err error) {
+	allowed = true
+
+	switch req.Operation {
+	case admissionv1.Create:
+		if utilfeature.DefaultFeatureGate.Enabled(features.PodUnavailableBudgetDeleteGate) {
```
FillZpp, line 32: Maybe we should consider container restart as an Update operation instead of Delete, which means you should check `PodUnavailableBudgetUpdateGate` feature-gate.

PR #843, opened 2021-12-13
do not exit when webhook ready timeout

shiyan2016:
do not exit when webhook ready timeout


main.go
```diff
@@ -177,7 +177,6 @@ func main() {
 		setupLog.Info("wait webhook ready")
 		if err = webhook.WaitReady(); err != nil {
 			setupLog.Error(err, "unable to wait webhook ready")
-			os.Exit(1)
```
FillZpp, line 180: Should we remove the `return err` in `WaitReady` instead of ignoring error here?

If controllers started before webhook ready, they may fail to create or update resources because apiserver can not connect to webhook.

FillZpp (approved):
/lgtm

PR #845, opened 2021-12-13
fix #844 When updating statefullset, the total number of pods being u…

### LINKED ISSUE #844: [BUG] When updating statefullset, the total number of pods being updated may exceed the expected value
**What happened**:
commit  id `02685e224be0215f48d8bb7b23d7892bbfaa9c5f` remove [`updateExpectations.ExpectUpdated`](https://github.com/openkruise/kruise/commit/02685e224be0215f48d8bb7b23d7892bbfaa9c5f#diff-a1e25eb50e4b2aed8947c2caa5d6d1ddfef1a5f2f95fdf71f2388c657823ee13L706) in statefulset .  
This will make `updateExpectations` not working, and may cause the `reconcile function` to obtain an informer cache that has not been updated.

`updateExpectations`should not be satisfied when the informer cache has not been updated.

When informer cahce is not updated in time and the user changes the update order, it may cause the number of updated pods to exceed the expected value(maxUnavailable or partition)

**Environment**:
- Kruise version: 0.10.0
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

hzyfox:
Add `updateExpectations.ExpectUpdated` back to `StatefulSet`

fix #844

### DIFF
--- a/pkg/controller/statefulset/stateful_set_control.go
+++ b/pkg/controller/statefulset/stateful_set_control.go
@@ -858,10 +858,10 @@ func (ssc *defaultStatefulSetControl) inPlaceUpdatePod(
 			}
 
 			if res.UpdateErr == nil {
+				updateExpectations.ExpectUpdated(getStatefulSetKey(set), updateRevision.Name, pod)
 				ssc.recorder.Eventf(set, v1.EventTypeNormal, "SuccessfulUpdatePodInPlace", "successfully update pod %s in-place(revision %v)", pod.Name, updateRevision.Name)
 				return res.InPlaceUpdate, nil
 			}
-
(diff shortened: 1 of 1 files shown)


pkg/controller/statefulset/stateful_set_control.go
```diff
@@ -861,7 +861,7 @@ func (ssc *defaultStatefulSetControl) inPlaceUpdatePod(
 				ssc.recorder.Eventf(set, v1.EventTypeNormal, "SuccessfulUpdatePodInPlace", "successfully update pod %s in-place(revision %v)", pod.Name, updateRevision.Name)
 				return res.InPlaceUpdate, nil
 			}
-
+			updateExpectations.ExpectUpdated(getStatefulSetKey(set), updateRevision.Name, pod)
```
FillZpp, line 864: Should it be in the `if res.UpdateErr == nil { ... }` above?

hzyfox: Done.  Maybe we should add some unit tests to cover this scenario

FillZpp (approved):
/lgtm

PR #853, opened 2021-12-16
WorkloadSpread Support Deployment RollingUpdate

### LINKED ISSUE #834: [feature request] WorkloadSpread should consider the update strategy where Pod is created before deletion.
When the new Pod is created in this strategy, WorkloadSpread don't know which is the pod this new pod will replace, so cannot assign the new pod to the right subset.

- WorkloadSpread should support rollingUpdate type for Deployment；
- WorkloadSpread should support `MaxSurge` for CloneSet & Deployment;

**Why is this needed**:
For the better support to CloneSet & Deployment.

veophi:

**WorkloadSpread will only care about the pods with the latest revision for Deployment:** 
- Added `Status.ObservedLatestRevision` field to record the latest pod template revision of  workload, so as to know whether the status need to be refresh;
- At webhook, refresh `Status` when observedLatest is outdated. 

fixed #834

### DIFF
--- a/pkg/util/workloadspread/workloadspread.go
+++ b/pkg/util/workloadspread/workloadspread.go
@@ -19,6 +19,7 @@ package workloadspread
 import (
 	"context"
 	"encoding/json"
+	"math"
 	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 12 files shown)


pkg/util/workloadspread/workloadspread.go
```diff
@@ -392,31 +397,204 @@ func (h *Handler) tryToGetTheLatestMatchedWS(matchedWS *appsv1alpha1.WorkloadSpr
...
+			if deployment.Spec.Replicas != nil {
+				workloadReplicas = *deployment.Spec.Replicas
+			} else {
+				// deployment default replicas is 1
+				workloadReplicas = 1
+			}
+
+			replicaset := h.GetOwnerReplicaSetFor(pod)
+			// if the replicaset is not the latest revision, return false
+			if replicaset != nil && !util.EqualIgnoreHash(&deployment.Spec.Template, &replicaset.Spec.Template) {
```
veophi, line 444: We should check who is the newer one between deployment and replicaset via `Deployment Revision Annotation`.

PR #854, opened 2021-12-16
optimization to determine whether the pod state is consistent logic

dafu-wu:
when we to determine whether the pod state is consistent logic and Is Image digest, call func IsPodContainerDigestEqual。
The pod container statuses may be evaluated multiple times by the func。
I'm trying to optimize it, maybe there's a better way to optimize it, maybe there's no need to optimize it.

### DIFF
--- a/pkg/control/sidecarcontrol/sidecarset_control.go
+++ b/pkg/control/sidecarcontrol/sidecarset_control.go
@@ -155,6 +155,7 @@ func (c *commonControl) IsPodStateConsistent(pod *v1.Pod, sidecarContainers sets
 	}
 
 	allDigestImage := true
+	cImageIDs := util.GetPodContainerImageIDs(pod)
 	for _, container := range pod.Spec.Containers {
 		// only check whether sidecar container is consistent
 		if !sidecarContainers.Has(container.Name) {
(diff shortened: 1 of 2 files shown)


pkg/util/pods.go
```diff
@@ -203,7 +203,7 @@ func IsRunningAndReady(pod *v1.Pod) bool {
 	return pod.Status.Phase == v1.PodRunning && podutil.IsPodReady(pod)
 }
 
-func IsPodContainerDigestEqual(containers sets.String, pod *v1.Pod) bool {
+func GetPodContainerStatusesMap(pod *v1.Pod) map[string]string {
```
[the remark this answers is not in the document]
dafu-wu: ok, your suggestion is better. I correct it.

FillZpp:
/lgtm
PTAL @zmberg

pkg/control/sidecarcontrol/sidecarset_control.go
```diff
@@ -168,7 +169,11 @@ func (c *commonControl) IsPodStateConsistent(pod *v1.Pod, sidecarContainers sets
 			continue
```
zmberg, line 169: continue -> break ?

dafu-wu: I think so too, let me optimize it.

PR #860, opened 2021-12-20
Fix log error when extract container from fieldpath failed

pigletfly:
Signed-off-by: pigletfly <wangbing.adam@gmail.com>


FillZpp (approved):
/lgtm

PR #863, opened 2021-12-21
remove dead code

dafu-wu:
remove dead code when updatePodSidecarContainer.

maybe, when updating pod sidecarContainer Only Image fields are updated，so I don't think these merge fields make sense at the moment.
Will these fields be updated in the future？


dafu-wu:
> /cc @zmberg

@zmberg  Could this code be removed？If I don't have to, I'll close the PR

zmberg:
@dafu-wu tks for you. I think the optimization is reasonable, but because of the impact with the internal logic, I need to spend some time to verify it first.

dafu-wu:
> 

ok brother, got it

PR #864, opened 2021-12-24
allow to update asts RevisionHistoryLimit

shiyan2016:
allow to update asts RevisionHistoryLimit


FillZpp:
It seems the built-in StatefulSet does not allow to modify the `revisionHistoryLimit` field. But I think it will not break anything if it is mutable?

shiyan2016:
> It seems the built-in StatefulSet does not allow to modify the `revisionHistoryLimit` field. But I think it will not break anything if it is mutable?

yes, and I think it is reasonable for revisionHistoryLimit mutable.

FillZpp (approved):
/lgtm

PR #884, opened 2022-01-10
Fix validateSidecarConflict

pigletfly:
Signed-off-by: pigletfly <wangbing.adam@gmail.com>

scenario: a same sidecar container exists in multiple other sidecarset


zmberg:
tks, you are right. But I'd like to think about it in a different way and see if you can:
func validateSidecarConflict(xxxx) field.ErrorList {
	allErrs := field.ErrorList{}
        matchedList := make([]*appsv1alpha1.SidecarSet, 0)
	for i := range sidecarSets.Items {
		obj := &sidecarSets.Items[i]
		if util.IsSelectorOverlapping(sidecarSet.Spec.Selector, obj.Spec.Selector) {
			matchedList = append(matchedList, obj)
		}
	}

       // This matchedList can be facilitated later. 
      // Because only two sidecar overlaps are meaningful for comparison, and this situation does not exist "a same sidecar container exists in multiple other sidecarset"?

PR #893, opened 2022-01-18
CloneSet refresh pod states before skipping update when paused

### LINKED ISSUE #892: [BUG] extra ReadinnesGate with ReCreate updateStrategy
**What happened**:
CloneSet pods do not become ready due to  `ReadinnesGate` ` InPlaceUpdateReady`. Pod status shows
```
    message: corresponding condition of pod readiness gate "InPlaceUpdateReady" does
      not exist.
```

CloneSet pods are not affected by `ReadinnesGate` ` InPlaceUpdateReady` when `type` is `ReCreate`

* Create s `CloneSet` with
```
  updateStrategy:
    paused: true
    type: ReCreate
```
* get a couple of pods with added `ReadinnesGate` ` InPlaceUpdateReady` which doesn't become ready.
Temporary fix is to enable and re-disable `updateStrategy` using f.e. these patches:
```
kubectl patch cloneset test --type='merge' -p '{"spec": {"updateStrategy":{"paused":false}}}'
sleep 5
kubectl patch cloneset test --type='merge' -p '{"spec": {"updateStrategy":{"paused":true}}}'
```
But I have to toggle `paused` for every new pod of this `CloneSet`

What I'm trying to achieve is to mimic StatefulSet's `updateStrategy` `OnDelete`, where I can change StatefulSet as many times as I need to, but pods are not affected until pod is created after scale up or pod is re-created after delete. That's why I need to pause `updateStrategy ` in `CloneSet`  - I have to ensure that new pod will be re-created after delete with the same name to use existing PVC. At the same time I need to control which pods have to be scaled down instead of latest pods in StatefulSet. PV& PVC auto removal feature of `CloneSet` is useful too.

**Environment**:
GKE 1.20
- Kruise version:
1.0.0
Install via helm with args
```
installation:
  namespace: kruise-system
```
- Others:

FillZpp:

CloneSet refresh pod states before skipping update when paused

fixes #892

### DIFF
--- a/pkg/controller/cloneset/sync/cloneset_update.go
+++ b/pkg/controller/cloneset/sync/cloneset_update.go
@@ -48,10 +48,6 @@ func (c *realControl) Update(cs *appsv1alpha1.CloneSet,
 	requeueDuration := requeueduration.Duration{}
 	coreControl := clonesetcore.New(cs)
 
-	if cs.Spec.UpdateStrategy.Paused {
-		return requeueDuration.Get(), nil
-	}
-
 	// 1. refresh states for all pods
 	var modified bool
 	for _, pod := range pods {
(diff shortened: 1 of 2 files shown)


furykerry (commented):
/lgtm

PR #894, opened 2022-01-18
Improve Kruise E2E image registry and fix image pulling

FillZpp:

1. Improve Kruise E2E cases
2. Make E2E image registry to be configurable
3. Fix image pulling for specific images like `docker.io/library/nginx:1.14-alpine`


test/e2e/apps/broadcastjob.go
```diff
@@ -82,7 +82,7 @@ var _ = SIGDescribe("BroadcastJob", func() {
 
 	framework.KruiseDescribe("BroadcastJob dispatching", func() {
 
-		ginkgo.It("succeeds for parallelism < number of node", func() {
+		framework.ConformanceIt("succeeds for parallelism < number of node", func() {
```
furykerry, line 85: why change the test to conformance test ?

furykerry (commented):
/lgtm

test/e2e/apps/broadcastjob.go
yjjy0921: I think the e2e framework is based on k8s e2e framework. But we should not use ConformanceIt here.Need fix it

PR #909, opened 2022-02-21
Support in-place update containers with launch priority

### LINKED ISSUE #901: [feature request] Support ordered execution of the application containers during in-place updates in CloneSet

We would like CloneSet to support ordered execution of the application containers during in-place updates.

Specifically, 
- user can specify the priority for each application container in a pod. 
- When in-place updates happen  (e.g version change) in the multiple applications containers, the application containers stop and start based on the priority/ordering container by container.
- in-place update to _a_ container (not multiple containers) with specified ordering just have effect to that container, and no impact to the other containers even though they are specified with orders. 
- For containers that are not specified with the orders/priority, no guarantees of the execution order is fine.
- In the case of containers with specified orders fail at kubelet probes and gets restarted by kubelet, the ordering might not apply anymore. But it will be a good-to-have to ensure ordering restarting as well. 
 
**Why is this needed**:

We believe this is a critical feature that can help us to leverage the CloneSet in-place updates successfully.

We have business logic have to be lined up for execution. While we can (and are) leveraging init container to ensure the execution ordering for some business logic. However, updates to the init containers will cause the whole pod to be recreated.
Even though we can do in-place update in the application container, the nature of the init container updates defeat the whole purpose of in-place updates. 

While we can separate the business logic in init container per pod lifecycle to avoid updates in init container, but in practice, there are business logic is not per pod lifecycle but per container lifecycle and we have to line them up for ordered execution. With the feature support, we can put the business logic into long running service as in application container and achieve ordered in-place updates

FillZpp:

Support in-place update containers with launch priority.

- For pods without container orders, no guarantees of the execution order during in-place update multiple containers.
- For pods with container orders:
  - keep execution order during in-place update multiple containers with different priorities.
  - no guarantees of the execution order during in-place update multiple containers with the same priority.

fixes #901


apis/apps/pub/inplace_update.go
```diff
@@ -61,6 +61,23 @@ type InPlaceUpdateState struct {
 
 	// UpdateEnvFromMetadata indicates there are envs from annotations/labels that should be in-place update.
 	UpdateEnvFromMetadata bool `json:"updateEnvFromMetadata,omitempty"`
+
+	// NextContainerImages is the containers with lower priority that waiting for in-place update in next batch.
+	NextContainerImages map[string]string `json:"nextContainerImages,omitempty"`
+
+	// NextContainerImages is the containers with lower priority that waiting for in-place update in next batch.
+	NextContainerRefMetadata map[string]metav1.ObjectMeta `json:"nextContainerRefMetadata,omitempty"`
```
furykerry, line 69: plz correct the comment about NextContainerRefMetadata

FillZpp: updated

pkg/util/inplaceupdate/inplace_update_defaults.go
```diff
@@ -100,55 +230,91 @@ func defaultCalculateInPlaceUpdateSpec(oldRevision, newRevision *apps.Controller
...
 		idx, _ := strconv.Atoi(words[3])
 		if len(oldTemp.Spec.Containers) <= idx {
 			return nil
 		}
-		updateSpec.ContainerImages[oldTemp.Spec.Containers[idx].Name] = jsonPatchOperation.Value.(string)
+		updateSpec.ContainerImages[oldTemp.Spec.Containers[idx].Name] = op.Value.(string)
 	}
-	if metadataChanged {
+
+	if len(metadataPatches) > 0 {
```
furykerry, line 266: no need to check len, iteration through empty map will do nothing

pkg/util/inplaceupdate/inplace_update_defaults.go
```diff
@@ -100,55 +230,91 @@ func defaultCalculateInPlaceUpdateSpec(oldRevision, newRevision *apps.Controller
...
+		if utilfeature.DefaultFeatureGate.Enabled(features.InPlaceUpdateEnvFromMetadata) {
+			for _, op := range metadataPatches {
```
furykerry, line 268: plz add a path example here

pkg/util/inplaceupdate/inplace_update_test.go
```diff
@@ -24,13 +24,16 @@ import (
 	"time"
 
 	appspub "github.com/openkruise/kruise/apis/apps/pub"
+	"github.com/openkruise/kruise/pkg/features"
```
furykerry, line 27: plz add ut for defaultPatchUpdateSpecToPod

pkg/util/inplaceupdate/inplace_update_defaults.go
FillZpp: There is not only the iteration, but also calculate strategic patch using old and new templates.

pkg/util/inplaceupdate/inplace_update_defaults.go
FillZpp: done

pkg/util/inplaceupdate/inplace_update_test.go
FillZpp: done

furykerry (commented):
/lgtm

PR #916, opened 2022-03-02
Disable DeepCopy for some specific cache list

### LINKED ISSUE #872: [feature request] Disable cache DeepCopy to reduce memory usage for kruise-manager
After #871 has been done, we should use the new feature `UnsafeDisableDeepCopy` in controller-runtime v0.10.

This feature allows developers to avoid DeepCopy for some kinds of resources when they get or list from cache. We can start with Pod, which is the most cost to memory cache in large-scale clusters.

**Why is this needed**:

Reduce memory usage for kruise-manager.

Note that once we disable DeepCopy for a resource kind, when we get or list objects of this kind from cache, we have to deep copy first before modify them.

FillZpp:

1. A custom delegating client has been added, which supports to use custom cache reader with DisableDeepCopy option.
2. Disable DeepCopy for some specific cache list.

fixes #872

### DIFF
--- a/.github/workflows/e2e-1.16.yaml
+++ b/.github/workflows/e2e-1.16.yaml
@@ -70,7 +70,20 @@ jobs:
         run: |
           export KUBECONFIG=/home/runner/.kube/config
           make ginkgo
+          set +e
           ./bin/ginkgo -timeout 60m -v --focus='\[apps\] StatefulSet' test/e2e
+          retVal=$?
+          restartCount=$(kubectl get pod -n kruise-system -l control-plane=controller-manager --no-headers | awk '{print $4}')
+          if [ "${restartCount}" -eq "0" ];then
+              echo "$out"
+              echo "Kruise-manager has not restarted"
+          else
+              echo "$out"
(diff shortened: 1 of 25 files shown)


pkg/util/client/cache_reader.go
```diff
@@ -0,0 +1,263 @@
...
+func (_ disableDeepCopy) ApplyToList(_ *client.ListOptions) {
+}
+
+type cacheReader struct {
+	cache  cache.Cache
+	scheme *runtime.Scheme
+}
+
+func (r *cacheReader) Get(ctx context.Context, key client.ObjectKey, out client.Object) error {
+	gvk, err := apiutil.GVKForObject(out, r.scheme)
```
furykerry, line 55: is it neccessary to copy the CacheReader function here? delegate reader will not paas Get request to cacheReader, it seems this func is useless

FillZpp: A Reader interface contains Get and List methods. The cacheReader here is going to be a complete Reader without any deepcopy, even if we just use the List for now.

furykerry (commented):
/lgtm

PR #919, opened 2022-03-05
fix WaitReady error return

yxxhero:
Signed-off-by: yxxhero <aiopsclub@163.com>

fix WaitReady error return

fix WaitReady error return

### DIFF
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -42,6 +42,8 @@ var (
 	handlerGates = map[string]GateFunc{}
 )
 
+const waitReadyTimeout = 5 * time.Second
+
 func addHandlers(m map[string]admission.Handler) {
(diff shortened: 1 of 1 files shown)


FillZpp:
@yxxhero The returning error was removed in https://github.com/openkruise/kruise/pull/843 , for sometimes we don't know how long it costs to initialize cache in large-scale clusters.

yxxhero:
oh. I known. Will the startup of this webhook affect the mainstream program? @FillZpp Thanks very much.

FillZpp:
> oh. I known. Will the startup of this webhook affect the mainstream program? @FillZpp Thanks very much.

Yes, if the mainstream program you mean controllers of kruise. Controllers will fail to create/update some CRs if the webhook has not started.

PR #923, opened 2022-03-09
Support preDelete lifecycle for Advanced DaemonSet

### LINKED ISSUE #911: [feature request] support lifecycle in daemonset
Add lifestyle in advanced daemonset such as cloneset, PreparingDelete needed first.

**Why is this needed**:
We need to expand memory request for advanced daemonset, but not all node has enough memory.If delete pod unconditionally，new pod may pending a longtime.
we can use preparingDelete lifestyle to intercept delete  action when node resource is not enough , then wait other modules such as descheduler to make node resource enough.

FillZpp:

Support preDelete lifecycle for Advanced DaemonSet

fixes #911


pkg/controller/cloneset/sync/cloneset_update.go
```diff
@@ -255,6 +256,7 @@ func (c *realControl) updatePod(cs *appsv1alpha1.CloneSet, coreControl clonesetc
 			if res.InPlaceUpdate {
 				if res.UpdateErr == nil {
 					c.recorder.Eventf(cs, v1.EventTypeNormal, "SuccessfulUpdatePodInPlace", "successfully update pod %s in-place(revision %v)", pod.Name, updateRevision.Name)
+					clonesetutils.ResourceVersionExpectations.Expect(&metav1.ObjectMeta{UID: pod.UID, ResourceVersion: res.NewResourceVersion})
```
furykerry, line 259: do we still need UpdateExpectations here ?

pkg/controller/daemonset/daemonset_controller.go
```diff
@@ -785,6 +816,9 @@ func (dsc *ReconcileDaemonSet) podsShouldBeOnNode(
 				// Emit an event so that it's discoverable to users.
 				dsc.eventRecorder.Eventf(ds, corev1.EventTypeWarning, FailedDaemonPodReason, msg)
 				podsToDelete = append(podsToDelete, pod.Name)
+			} else if isPodPreDeleting(pod) {
```
furykerry, line 819: why we need add PreparingDelete pods into podsToDelete, if there is some logic which make the pods in the podsToDelete in the first place , is it still the case when next reconcile?

pkg/controller/cloneset/sync/cloneset_update.go
FillZpp: No, I will remove it in next PR.

pkg/controller/daemonset/daemonset_controller.go
FillZpp: Yeah, we should put it into podsToDelete even if it has been set to PreparingDelete in previous reconcile.

furykerry (commented):
/lgtm

PR #850, opened 2021-12-15
SidecarSet hot upgrade containers should be consistent with pod's

veophi:
Signed-off-by: veophi <vec.g.sun@gmail.com>

### Ⅰ. Describe what this PR does

SidecarSet hot upgrade containers should be consistent with pod's, otherwise, will not match the pod/sidecarset.

Ⅱ. Does this pull request fix one issue?
fixed #855

PR #927, opened 2022-03-15
Support in place upgrade Pod injection sidecar container

zmberg:

Fixed in PR, https://github.com/openkruise/kruise/pull/850


FillZpp (approved):
/lgtm

PR #931, opened 2022-03-18
Optimize inplace update and support pod-template-hash label for cloneset

FillZpp:

Optimize inplace update and support `pod-template-hash` label for cloneset.

Now there are two hash labels in pods that managed by cloneset:
- `controller-revision-hash` is still the important one that indicates the pod version. It defaults to be `"{CLONESET_NAME}-{HASH}"`, and it will be `"{HASH}"` if user enabled the `CloneSetShortHash` feature-gate.
- `pod-template-hash` should always be the short format `"{HASH}"`. cloneset-controller will not rely on it, but some other controllers, such as kruise-rollout, may use it to be the service selector.

### DIFF
--- a/pkg/controller/cloneset/sync/cloneset_update.go
+++ b/pkg/controller/cloneset/sync/cloneset_update.go
@@ -17,6 +17,7 @@ limitations under the License.
 package sync
 
 import (
+	"context"
 	"fmt"
 	"sort"
 	"time"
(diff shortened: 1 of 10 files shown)


pkg/controller/cloneset/sync/cloneset_update.go
```diff
@@ -198,6 +206,22 @@ func (c *realControl) refreshPodState(cs *appsv1alpha1.CloneSet, coreControl clo
 	return false, res.DelayDuration, nil
 }
 
+func (c *realControl) fixPodTemplateHashLabel(cs *appsv1alpha1.CloneSet, pod *v1.Pod) (bool, error) {
```
furykerry, line 209: please comment the history and reason of why we need "fix" template hash. It seems that previously we only has controllerrevisionhash in the pod

furykerry (commented):
/lgtm

Issue #944, opened 2022-04-01
[feature request] set pod not ready in PreparingUpdate and PreparingDelete lifecycle hook


ZYecho:
**What would you like to be added**:
set pod not ready in PreparingUpdate and PreparingDelete lifecycle hook

**Why is this needed**:
ccm（https://github.com/kubernetes/cloud-provider-alibaba-cloud） is responsible for create loadbalancer in cloud environment. ccm starts working when pod not ready, it removes the pod from LB.

we want to use cloneset PreparingUpdate and PreparingDelete lifecycle hook to inject some code to implement pod's graceful offline. without pod's not ready, ccm can't start working.

FillZpp:
closed by #979 
/close

PR #949, opened 2022-04-12
Improve Lifecycle

veophi:

- Fix invalid lifecycle PreDelete hook for StatefulSet.
- Add `RequiredPodNotReady` lifecycle policy for CloneSet/StatefulSet/DaemonSet.

---
**Explanation:** `RequiredPodNotReady = true` means:
- Pod will be set to 'NotReady' at preparingDelete/preparingUpdate state.
- Pod will be restored to 'Ready' at Updated state if it was set to 'NotReady' at preparingUpdate state.

---

#944 
#950

### DIFF
--- a/apis/apps/pub/lifecycle.go
+++ b/apis/apps/pub/lifecycle.go
@@ -40,4 +40,9 @@ type Lifecycle struct {
 type LifecycleHook struct {
 	LabelsHandler     map[string]string `json:"labelsHandler,omitempty"`
 	FinalizersHandler []string          `json:"finalizersHandler,omitempty"`
+	// MarkPodNotReady = true means:
+	// - Pod will be set to 'NotReady' at preparingDelete/preparingUpdate state.
+	// - Pod will be restored to 'Ready' at Updated state if it was set to 'NotReady' at preparingUpdate state.
+	// Default to false.
+	MarkPodNotReady bool `json:"markPodNotReady,omitempty"`
 }
--- a/pkg/controller/cloneset/sync/cloneset_scale.go
+++ b/pkg/controller/cloneset/sync/cloneset_scale.go
@@ -165,6 +165,14 @@ func (r *realControl) Scale(
 }
 
 func (r *realControl) managePreparingDelete(cs *appsv1alpha1.CloneSet, pods, podsInPreDelete []*v1.Pod, numToDelete int) (bool, error) {
+	//  We do not allow regret once the pod enter PreparingDelete state if MarkPodNotReady is set.
+	// Actually, there is a bug cased by this transformation from PreparingDelete to Normal,
+	// i.e., Lifecycle Updated Hook may be lost if the pod was transformed from Updating state
+	// to PreparingDelete.
+	if lifecycle.IsRequiredPodNotReady(cs.Spec.Lifecycle) {
+		return false, nil
+	}
+
 	diff := int(*cs.Spec.Replicas) - len(pods) + numToDelete
 	var modified bool
 	for _, pod := range podsInPreDelete {
@@ -177,7 +185,7 @@ func (r *realControl) managePreparingDelete(cs *appsv1alpha1.CloneSet, pods, pod
 
 		klog.V(3).Infof("CloneSet %s patch pod %s lifecycle from PreparingDelete to Normal",
 			clonesetutils.GetControllerKey(cs), pod.Name)
-		if updated, gotPod, err := r.lifecycleControl.UpdatePodLifecycle(pod, appspub.LifecycleStateNormal); err != nil {
+		if updated, gotPod, err := r.lifecycleControl.UpdatePodLifecycle(pod, appspub.LifecycleStateNormal, false); err != nil {
 			return modified, err
 		} else if updated {
 			modified = true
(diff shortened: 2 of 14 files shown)


pkg/util/inplaceupdate/inplace_update.go
```diff
@@ -161,6 +163,29 @@ func (c *realControl) Refresh(pod *v1.Pod, opts *UpdateOptions) RefreshResult {
 	return RefreshResult{RefreshErr: err}
 }
 
+func (c *realControl) SetPodNotReady(pod *v1.Pod) (bool, error) {
+	if isPodAlreadyNotReady(pod) {
```
furykerry, line 167: pod readiness status can be unstable,  pod can return to be ready but still waiting for deletion. Is it more safe to update the readiness condition even if the pod is not ready

furykerry (commented):
/lgtm

apis/apps/pub/lifecycle.go
```diff
@@ -31,6 +31,8 @@ type LifecycleStateType string
 
 // Lifecycle contains the hooks for Pod lifecycle.
 type Lifecycle struct {
+	// Policy declares the policy when managing lifecycle.
+	Policy *LifecyclePolicy `json:"policy,omitempty"`
```
FillZpp, line 35: How about this, if we will also add NotReady for InPlaceUpdate hook?

```golang
type LifecycleHook struct {
	LabelsHandler     map[string]string `json:"labelsHandler,omitempty"`
	FinalizersHandler []string          `json:"finalizersHandler,omitempty"`
        // ...
        RequiredNotReady bool `json:"requiredNotReady,omitempty"`
}
```

pkg/controller/cloneset/sync/cloneset_scale.go
```diff
@@ -278,6 +290,17 @@ func (r *realControl) deletePods(cs *appsv1alpha1.CloneSet, podsToDelete []*v1.P
...
 				clonesetutils.ResourceVersionExpectations.Expect(gotPod)
 			}
+
+			if lifecycle.IsPodNotReadyDuringPreparingDeletePolicy(cs.Spec.Lifecycle) {
+				if updated, err := podreadiness.SetPodNotReadyDuringPreparingDelete(r.Client, pod, getReadinessMessage(cs)); err != nil {
+					return false, err
+				} else if updated {
+					klog.V(3).Infof("CloneSet %s scaling update pod %s not ready condition due to PreparingDelete",
+						clonesetutils.GetControllerKey(cs), pod.Name)
+					modified = true
```
FillZpp, line 300: All these `modified` are unnecessary. We only use the `modified` set by `UpdatePodLifecycle` before with `ResourceVersionExpectations` to avoid duplicate operation in next reconciles.

pkg/controller/cloneset/sync/cloneset_scale.go
```diff
@@ -278,6 +290,17 @@ func (r *realControl) deletePods(cs *appsv1alpha1.CloneSet, podsToDelete []*v1.P
 				modified = true
 				clonesetutils.ResourceVersionExpectations.Expect(gotPod)
 			}
+
+			if lifecycle.IsPodNotReadyDuringPreparingDeletePolicy(cs.Spec.Lifecycle) {
+				if updated, err := podreadiness.SetPodNotReadyDuringPreparingDelete(r.Client, pod, getReadinessMessage(cs)); err != nil {
+					return false, err
```
FillZpp, line 296: 1. If it fails, the pod in PreparingDelete state will still be ready. So maybe add an error log to warn this.
2. Should this `SetPodNotReady` logic be before or after the `UpdatePodLifecycle` logic? Should we rollback the first operation if the second failed?

pkg/controller/cloneset/sync/cloneset_sync_utils.go
```diff
@@ -241,3 +243,7 @@ func isPodAvailable(coreControl clonesetcore.Control, pod *v1.Pod, minReadySecon
 	}
 	return coreControl.IsPodUpdateReady(pod, minReadySeconds)
 }
+
+func getReadinessMessage(cs *appsv1alpha1.CloneSet) podreadiness.Message {
+	return podreadiness.Message{UserAgent: "CloneSet", Key: fmt.Sprintf("%s/%s", cs.Namespace, cs.Name)}
```
FillZpp, line 248: The Message must be unique. So maybe `podreadiness.Message{UserAgent: "CloneSet", Key: "PreDeleteHook"}`

(This is different with the CRR. There might be multiple CRRs for one Pod, so we have to put the name of CRR in key to declare which one triggers not-ready. But one Pod only belongs to a CloneSet, so we don't have to put the name of CloneSet in it)

pkg/util/podreadiness/pod_readiness_utils.go
```diff
@@ -30,11 +30,18 @@ import (
...
 )
 
-func AddNotReadyKey(c client.Client, pod *v1.Pod, msg Message) error {
+var (
+	// SetPodNotReadyDuringPreparingDelete is func that set pod not ready at
+	// lifecycle preparingDelete state for CloneSet/StatefulSet/DaemonsSet.
+	SetPodNotReadyDuringPreparingDelete = AddNotReadyKey
+)
+
+func AddNotReadyKey(c client.Client, pod *v1.Pod, msg Message) (bool, error) {
```
FillZpp, line 39: `modified` may be unnecessary.

pkg/util/podreadiness/pod_readiness_utils.go
```diff
@@ -61,8 +68,13 @@ func AddNotReadyKey(c client.Client, pod *v1.Pod, msg Message) error {
 			condition.LastTransitionTime = metav1.Now()
 		}
 
-		return c.Status().Update(context.TODO(), newPod)
+		err := c.Status().Update(context.TODO(), newPod)
```
FillZpp, line 71: The old logic here can be optimized that we'd better update the Ready condition to False after set KruisePodReady condition to False. Like what we do in inplace update:

https://github.com/openkruise/kruise/blob/5b52b6eff3ff93210206eeaeb79751a16043bfd9/pkg/util/inplaceupdate/inplace_update.go#L172-L175

PR #952, opened 2022-04-18
optmize the logic of listNamespacesForDistributor func

hantmac:

Optmize the logic of listNamespacesForDistributor func.

NONE

### DIFF
--- a/pkg/controller/resourcedistribution/utils.go
+++ b/pkg/controller/resourcedistribution/utils.go
@@ -293,8 +293,15 @@ func listNamespacesForDistributor(handlerClient client.Client, targets *appsv1al
 		for _, namespace := range namespacesList.Items {
 			matchedSet.Insert(namespace.Name)
 		}
-	} else if len(targets.NamespaceLabelSelector.MatchLabels) != 0 || len(targets.NamespaceLabelSelector.MatchExpressions) != 0 {
-		// 2. select the namespaces via targets.NamespaceLabelSelector
+	} else {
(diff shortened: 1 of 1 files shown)


pkg/controller/resourcedistribution/utils.go
```diff
@@ -293,8 +293,15 @@ func listNamespacesForDistributor(handlerClient client.Client, targets *appsv1al
...
-	} else if len(targets.NamespaceLabelSelector.MatchLabels) != 0 || len(targets.NamespaceLabelSelector.MatchExpressions) != 0 {
-		// 2. select the namespaces via targets.NamespaceLabelSelector
+	} else {
+		// 2. select the namespaces via targets.IncludedNamespaces
+		for _, namespace := range targets.IncludedNamespaces.List {
+			matchedSet.Insert(namespace.Name)
+		}
+	}
+
+	if len(targets.NamespaceLabelSelector.MatchLabels) != 0 || len(targets.NamespaceLabelSelector.MatchExpressions) != 0 {
```
veophi, line 303: if `target.AllNamespaces`  is selected, this `list` operation may be redundant.

hantmac: You mean user will set `target.AllNamespaces` and `targets.NamespaceLabelSelector` in the same time?

veophi: > You mean user will set `target.AllNamespaces` and `targets.NamespaceLabelSelector` in the same time?

yeah, though it is not a common settings.

PR #954, opened 2022-04-18
[Feat]: Make sure at least one pod is upgraded if CloneSet Partition != 100% 

### LINKED ISSUE #926: [BUG] Percent partitions always round up causing no updates
**What happened**:
The use case I am wishing to model is allowing for deploying only 1 pod on the new revision and then removing that split. I would expect that setting a partition to 99% should effectively behave like the maxSurge in Deployment and default to at least 1 replica, it seems we round up tho, so we end up with no replicas.

99% Partition
3 pods, expect 1 pod to get new revision

Make a cloneset with under 100 replicas
updateStrategy:
    maxSurge: 1
    maxUnavailable: 50%
    partition: 99%
    type: ReCreate

**Environment**: 
- Kruise version: Latest
- Kubernetes version (use `kubectl version`):  v1.19.2
- Install details (e.g. helm install args): helm install kruise openkruise/kruise --version 1.0.1
- Others:

### LINKED ISSUE #953: [feature request] Make sure at least one pod is upgraded if CloneSet Partition != 100%
This feature request is for CloneSet Partition with percentage type:
- CloneSet controller should make sure that at least one pod is upgraded if `replicas>0 && partition != "100%"`;
- CloneSet should show the absolute value of Partition in status if it has complex computational logic;

**Why is this needed**:
If we don't, when users guide traffic (e.g., 10% traffic) to the upgraded pods via ingress configurations,  this 10% traffic may be lost due to CloneSet scaling. For example, assuming that `replicas=10, partition=90%`, now we scale down the replicas to 9, the only upgraded pod will be deleted.

veophi:

Make sure at least one pod is upgraded if CloneSet Partition != 100%:
- If `replicas >= 2 && partition != 100%`, at least ONE pod will be upgraded;
- If `replicas >= 1` && partition != 0%, at least ONE pod will keep old revision (same behaviors as before);

fixes #953
fixes #926

### DIFF
--- a/apis/apps/v1alpha1/cloneset_types.go
+++ b/apis/apps/v1alpha1/cloneset_types.go
@@ -168,6 +168,10 @@ type CloneSetStatus struct {
 	// indicated by updateRevision and have a Ready Condition.
 	UpdatedReadyReplicas int32 `json:"updatedReadyReplicas"`
 
+	// ExpectedUpdatedReplicas is the number of Pods that should be updated by CloneSet controller.
+	// This field is calculated via Replicas - Partition.
+	ExpectedUpdatedReplicas int32 `json:"expectUpdatedReplicas,omitempty"`
+
 	// UpdateRevision, if not empty, indicates the latest revision of the CloneSet.
 	UpdateRevision string `json:"updateRevision,omitempty"`
 
--- a/pkg/controller/cloneset/cloneset_status.go
+++ b/pkg/controller/cloneset/cloneset_status.go
@@ -22,6 +22,7 @@ import (
 	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
 	clonesetcore "github.com/openkruise/kruise/pkg/controller/cloneset/core"
 	clonesetutils "github.com/openkruise/kruise/pkg/controller/cloneset/utils"
+	"github.com/openkruise/kruise/pkg/util"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
@@ -100,5 +101,13 @@ func (r *realStatusUpdater) calculateStatus(cs *appsv1alpha1.CloneSet, newStatus
 	// Consider the update revision as stable if revisions of all pods are consistent to it, no need to wait all of them ready
 	if newStatus.UpdatedReplicas == newStatus.Replicas {
 		newStatus.CurrentRevision = newStatus.UpdateRevision
+	}
+
+	if newStatus.UpdateRevision == newStatus.CurrentRevision {
+		newStatus.ExpectedUpdatedReplicas = *cs.Spec.Replicas
+	} else {
+		if partition, err := util.CalculatePartitionReplicas(cs.Spec.UpdateStrategy.Partition, cs.Spec.Replicas); err == nil {
+			newStatus.ExpectedUpdatedReplicas = *cs.Spec.Replicas - int32(partition)
+		}
 	}
 }
--- a/pkg/controller/cloneset/sync/cloneset_sync_utils.go
+++ b/pkg/controller/cloneset/sync/cloneset_sync_utils.go
@@ -76,8 +76,12 @@ func calculateDiffsWithExpectation(cs *appsv1alpha1.CloneSet, pods []*v1.Pod, cu
 	replicas := int(*cs.Spec.Replicas)
 	var partition, maxSurge, maxUnavailable, scaleMaxUnavailable int
 	if cs.Spec.UpdateStrategy.Partition != nil {
-		partition, _ = intstrutil.GetValueFromIntOrPercent(cs.Spec.UpdateStrategy.Partition, replicas, true)
-		partition = integer.IntMin(partition, replicas)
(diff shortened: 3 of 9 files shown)


pkg/controller/cloneset/sync/cloneset_sync_utils.go
```diff
@@ -241,3 +240,26 @@ func isPodAvailable(coreControl clonesetcore.Control, pod *v1.Pod, minReadySecon
...
+	if replicasPointer != nil {
+		replicas = int(*replicasPointer)
+	}
+
+	pValue, err := intstrutil.GetScaledValueFromIntOrPercent(partition, replicas, true)
+	if err != nil {
+		return pValue, err
+	}
+
+	if replicas > 1 && pValue == replicas && !reflect.DeepEqual(*partition, intstrutil.FromString("100%")) {
```
furykerry, line 259: what if partition is of int type?

pkg/controller/cloneset/sync/cloneset_sync_utils.go
```diff
@@ -241,3 +240,26 @@ func isPodAvailable(coreControl clonesetcore.Control, pod *v1.Pod, minReadySecon
 	}
 	return coreControl.IsPodUpdateReady(pod, minReadySeconds)
 }
+
+func CalculatePartitionReplicas(partition *intstrutil.IntOrString, replicasPointer *int32) (int, error) {
```
furykerry, line 244: consider move the function to  a package that is common to all controllers, so that statefulset & daemonset can reuse this function

pkg/controller/cloneset/sync/cloneset_sync_utils.go
furykerry, line 259: plz comment why replicas > 1 here

apis/apps/v1alpha1/cloneset_types.go
```diff
@@ -168,6 +168,11 @@ type CloneSetStatus struct {
 	// indicated by updateRevision and have a Ready Condition.
 	UpdatedReadyReplicas int32 `json:"updatedReadyReplicas"`
 
+	// ExpectedPartitionReplicas is the number of Pods that should be partitioned by CloneSet controller. This field
+	// is designed to let users know the absolute number of Partition calculated by CloneSet controller if they use
+	// percentage as their Partition.
+	ExpectedPartitionReplicas int32 `json:"expectedPartitionReplicas,omitempty"`
```
furykerry, line 174: plz rename the field to expectUpdatedReplicas which should be more understandable for users

pkg/controller/cloneset/sync/cloneset_sync_utils.go
veophi: > what if partition is of int type?

In such case, we just return `pValue`.

furykerry:
this patch should also fix #926

pkg/controller/cloneset/cloneset_status.go
```diff
@@ -101,4 +102,12 @@ func (r *realStatusUpdater) calculateStatus(cs *appsv1alpha1.CloneSet, newStatus
 	if newStatus.UpdatedReplicas == newStatus.Replicas {
 		newStatus.CurrentRevision = newStatus.UpdateRevision
 	}
+
+	if newStatus.UpdateRevision == newStatus.CurrentRevision {
+		newStatus.ExpectedUpdatedReplicas = 0
```
furykerry, line 107: if UpdateRevision= CurrentRevision, ExpectedUpdatedReplicas=replicas sounds more reasonable to me

pkg/controller/cloneset/sync/cloneset_sync_utils_test.go
```diff
@@ -466,6 +466,43 @@ func TestCalculateDiffsWithExpectation(t *testing.T) {
 			},
 			expectResult: expectationDiffs{},
 		},
+		{
+			name: "update recreate partition=99% with maxSurge=2 (step 1/3)",
```
furykerry, line 470: plz consider add cases for percentage type maxUnavailable?

pkg/controller/cloneset/sync/cloneset_sync_utils_test.go
```diff
@@ -466,6 +466,80 @@ func TestCalculateDiffsWithExpectation(t *testing.T) {
...
+				createTestPod(oldRevision, appspub.LifecycleStateNormal, true, false),
+				createTestPod(oldRevision, appspub.LifecycleStateNormal, true, false),
+				createTestPod(oldRevision, appspub.LifecycleStateNormal, true, false),
+				createTestPod(newRevision, appspub.LifecycleStateNormal, false, false), // new creation
+			},
+			expectResult: expectationDiffs{},
+		},
+		{
+			name: "update recreate partition=99% with maxUnavailable=40%, maxSurge=30% (step 1/3)",
+			set:  createTestCloneSet(5, intstr.FromString("99%"), intstr.FromString("40%"), intstr.FromString("30%")),
```
furykerry, line 508: 5*0.4=2, plz add other cases to test the rounding logic of maxUnavailable

veophi: done

PR #959, opened 2022-04-20
Proposal: support multi-version daemonset

LinFCai:
Sometimes users may want to upgrade some daemon pods to version 1.1 and others to version 1.2 on specified nodes. For example, upgrade daemon pods in different CPU architecture nodes independently. A UnitedDaemonSet which is related to the existing DaemonSet can help users to achieve that. The controller will distribute multi-version pods of the existing DaemonSet according to UnitedDaemonSet subsets configuration.

None
None

None


LinFCai:
@FillZpp @furykerry 
we want to add a CRD to deploy multi-version daemon pods on different nodes and this is our idea,  thanks.

Issue #948, opened 2022-04-11
[feature request] sidecarSet support namespace selector


zmberg:
**What would you like to be added**:
Support for injecting sidecar via namespace selector, e.g., injecting envoy container by giving ns istio-injection=enabled

**Why is this needed**:

PR #969, opened 2022-05-07
sidecarSet support namespace selector

hantmac:

Support for injecting sidecar via namespace selector.

Fix https://github.com/openkruise/kruise/issues/948

### DIFF
--- a/pkg/control/sidecarcontrol/util.go
+++ b/pkg/control/sidecarcontrol/util.go
@@ -17,6 +17,7 @@ limitations under the License.
 package sidecarcontrol
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"regexp"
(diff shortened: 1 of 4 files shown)


FillZpp (commented):
Maybe we should also get namespaces from selector here to get pods in them.

https://github.com/openkruise/kruise/blob/cf95157d707acbfd0fea4a1d850eda57ea7cfe15/pkg/controller/sidecarset/sidecarset_processor.go#L269-L274

FillZpp:
> @FillZpp I just have a question, if the sidecarSet has `spec.Namespace=test-1` and `SidecarSetInjectNamespaceLabelSelectorAnnotation="xx=yy"` in the same time, the sidecarSet will inject sidecar into both `test-1` and namespaces with `xx=yy` label? Or which way has higher priority? For example, if set the annotation, we will use the namespaces with label first and omit the pod in ns `test-1`.

I think SidecarSet with `spec.Namespace` field should be forbidden to have `SidecarSetInjectNamespaceLabelSelectorAnnotation`. The former one means it should only be applied for pods in the specific namespace.

WDYT @zmberg

zmberg:
Is it possible to add NamespaceSelector filed in such a way that the spec.Namespace can be gradually deprecated later.

hantmac:
> 

Make sense.

hantmac:
> > 
> 
> That make sense. @FillZpp

hantmac:
> Is it possible to add NamespaceSelector filed in such a way that the spec.Namespace can be gradually deprecated later.

You mean add the `NamespaceSelector` field to `SidecarSet`?

zmberg:
> > Is it possible to add NamespaceSelector filed in such a way that the spec.Namespace can be gradually deprecated later.
> 
> You mean add the `NamespaceSelector` field to `SidecarSet`?

yes, spec.NamespaceSelector

hantmac:
> > > Is it possible to add NamespaceSelector filed in such a way that the spec.Namespace can be gradually deprecated later.
> > 
> > 
> > You mean add the `NamespaceSelector` field to `SidecarSet`?
> 
> yes, spec.NamespaceSelector

I'm afraid such a change is too big and will lead to incompatibility. And I think `spec.Namespace` and `spec.NamespaceSelector` also exist in the same time, because not all the namespace have labels.

FillZpp:
How about

```golang
type SidecarSetSpec struct {
    // ...
    DeprecatedNamespace string `json:"namespace,omitempty"`
    // ...
    NamespaceSelector *SidecarSetNamespaceSelector `json:"namespaceSelector,omitempty"`
    ...
}

// ...
// Namespaces and LabelSelector are mutually exclusive.
type SidecarSetNamespaceSelector struct {
    // ...
    Namespaces []string `json:"namespaces,omitempty"`
    // ...
    metav1.LabelSelector `json:",inline"`
}
```

It will be compatible with the existing `spec.namespace`, which is marked as deprecated and will be removed in future release.

PR #1007, opened 2022-06-27
fix bug: allow optional filed maxunavilable to be nil in ads, and set default …

### LINKED ISSUE #1006: [BUG] Why Optional Field <MaxUnavailable> can not be set nil?
**What happened**:
I use advanced daemonset and update with partition feature. However, when I update to v1.1.0, I can not update when I don't set MaxUnavailable and MaxSurge with kruise controller log error msg "couldn't get unavailable numbers: invalid value for MaxUnavailable: nil value for IntOrString".

As api doc said: "// The maximum number of DaemonSet pods that can be unavailable during the
	// update. Value can be an absolute number (ex: 5) or a percentage of total
	// number of DaemonSet pods at the start of the update (ex: 10%). Absolute
	// number is calculated from percentage by rounding up.
	// This cannot be 0 if MaxSurge is 0
	// Default value is 1.
	// Example: when this is set to 30%, at most 30% of the total number of nodes
	// that should be running the daemon pod (i.e. status.desiredNumberScheduled)
	// can have their pods stopped for an update at any given time. The update
	// starts by stopping at most 30% of those DaemonSet pods and then brings
	// up new DaemonSet pods in their place. Once the new pods are available,
	// it then proceeds onto other DaemonSet pods, thus ensuring that at least
	// 70% of original number of DaemonSet pods are available at all times during
	// the update.
	// +optional".
This is an optional field and default value is 1.

update daemon with nil MaxUnavailable.

**Environment**:
- Kruise version: v1.1.0
- Kubernetes version (use `kubectl version`): 1.16
- Others:

ABNER-1:
fix bug: allow optional filed max unavilable in ads, and set default value 1

1. allow optional filed max unavilable to be nil in ads
2. set max unavilable default value to 1

fixes #1006

update ads with nil MaxUnavailable.

### DIFF
--- a/pkg/controller/daemonset/daemonset_util.go
+++ b/pkg/controller/daemonset/daemonset_util.go
@@ -254,7 +254,10 @@ func unavailableCount(ds *appsv1alpha1.DaemonSet, numberToSchedule int) (int, er
 	}
 	r := ds.Spec.UpdateStrategy.RollingUpdate
 	if r == nil {
-		return 0, nil
+		return 1, nil
+	}
+	if r.MaxUnavailable == nil {
+		return 1, nil
 	}
 	return intstrutil.GetScaledValueFromIntOrPercent(r.MaxUnavailable, numberToSchedule, true)
 }
--- a/pkg/controller/daemonset/daemonset_util.go
+++ b/pkg/controller/daemonset/daemonset_util.go
@@ -254,10 +254,10 @@ func unavailableCount(ds *appsv1alpha1.DaemonSet, numberToSchedule int) (int, er
 	}
 	r := ds.Spec.UpdateStrategy.RollingUpdate
 	if r == nil {
-		return 1, nil
+		return 0, nil
 	}
 	if r.MaxUnavailable == nil {
-		return 1, nil
+		return 0, nil
 	}
 	return intstrutil.GetScaledValueFromIntOrPercent(r.MaxUnavailable, numberToSchedule, true)
 }


pkg/controller/daemonset/daemonset_util.go
```diff
@@ -254,7 +254,10 @@ func unavailableCount(ds *appsv1alpha1.DaemonSet, numberToSchedule int) (int, er
 	}
 	r := ds.Spec.UpdateStrategy.RollingUpdate
 	if r == nil {
-		return 0, nil
+		return 1, nil
```
FillZpp, line 257: This should not be changed.

pkg/controller/daemonset/daemonset_util.go
```diff
@@ -254,7 +254,10 @@ func unavailableCount(ds *appsv1alpha1.DaemonSet, numberToSchedule int) (int, er
...
+	}
+	if r.MaxUnavailable == nil {
+		return 1, nil
```
FillZpp, line 260: It should return `0` which is same as maxSurge ([here](https://github.com/openkruise/kruise/blob/master/pkg/controller/daemonset/daemonset_util.go#L266)), and then controller will set `maxUnavailable` to `1` if they are all `0` ([here](https://github.com/openkruise/kruise/blob/master/pkg/controller/daemonset/daemonset_update.go#L258)).

pkg/controller/daemonset/daemonset_util.go
ABNER-1: I just refer to the implementation of max surge and return the default value of maxunavilable in the doc.
Sorry to change this because I did not read all codes in ds reconcile function.

ABNER-1: Why I can not set default value when rollingupdate is nil?
This code is necessary when the webhook is unreachabole in my case.

FillZpp (approved):
/lgtm

Thanks!

PR #1008, opened 2022-06-28
add KruiseDaemon feature check in imagepulljob mutating webhook

YoghurtFree:
When the featuregate "KruiseDaemon" is not enabled, the imagepullljob mutating webhook should not take effect, the same as imagepullljob validating webhook does.


FillZpp:
> When the featuregate "KruiseDaemon" is not enabled, the imagepullljob mutating webhook should not take effect, the same as imagepullljob validating webhook does.

Question is, since imagepullljob validating webhook has already validated the feature-gate, why should we also validate it in mutating webhook?

PR #1022, opened 2022-07-14
pod state from updating to Normal should all hooked

shiyan2016:
pod state from updating to Normal should all hooked


FillZpp (approved):
/lgtm

PR #1029, opened 2022-07-17
docs: add resourcedistribution generator proposal

### LINKED ISSUE #943: [feature request]ResourceDistribution should consider configmapGenerator
ResourceDistribution should support other resource like configmapgenerator with Kustomize
**Why is this needed**:
In k8s, configmap only used as create.
This is not a good design.
So if we  want to change configmap we should delete and then create [of course we should pay attention to check exists]
So we cannot just simplily to

dong4325:

Add proposal for ResourceDistributionGenerator

fixes #943

### DIFF
--- a/docs/proposals/20220717-resourcedistribution-generator.md
+++ b/docs/proposals/20220717-resourcedistribution-generator.md
@@ -1,0 +1,285 @@
+---
+title: ResourceDistribution Generator
+authors:
+  - "@dong4325"
+reviewers:
+  - "@Fei-Guo"
+  - "@FillZpp"
+creation-date: 2022-07-17
+last-updated: 2022-07-19
+status: implementable
+---
+
+# resourcedistribution-generator
+
+## Table of Contents
+
+A table of contents is helpful for quickly jumping to sections of a proposal and for highlighting
+any additional information provided beyond the standard proposal template.
+[Tools for generating](https://github.com/ekalinin/github-markdown-toc) a table of contents from markdown are available.
+
+- [Title](#title)
+  - [Table of Contents](#table-of-contents)
+  - [Motivation](#motivation)
+  - [Proposal](#proposal)
+    - [API Definition](#api-definition)
+    - [NameSuffixHash](#namesuffixhash)
+  - [User Cases](#user-Cases)
+    - [A Sample Case](#a-sample-case)
+      - [Options and ResourceOptions](#options-and-resourceoptions)
+      - [Resource Source](#resource-source)
+      - [Type](#type)
+      - [Targets](#targets)
+      - [Behavior](#behavior)
+  - [Implementation History](#implementation-history)
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


docs/proposals/20220717-resourcedistribution-generator.md
```diff
@@ -0,0 +1,293 @@
...
+  AllNamespaces bool `json:"allNamespaces,omitempty" yaml:"allNamespaces,omitempty"`
+  
+  // ExcludedNamespaces is a list of excluded namespaces name.
+  ExcludedNamespaces []string `json:"excludedNamespaces,omitempty" yaml:"excludedNamespaces,omitempty"`
+  
+  // IncludedNamespaces is a list of included namespaces name.
+  IncludedNamespaces []string `json:"includedNamespaces,omitempty" yaml:"includedNamespaces,omitempty"`
+ 
+  // NamespaceLabelSelector for the generator.
+  NamespaceLabelSelector *namespaceLabelSelector `json:"namespaceLabelSelector,omitempty" yaml:"namespaceLabelSelector,omitempty"`
```
hantmac, line 92: How about using `NamespaceLabelSelector metav1.LabelSelector `json:"namespaceLabelSelector,omitempty"`` directly?

docs/proposals/20220717-resourcedistribution-generator.md
```diff
@@ -0,0 +1,293 @@
...
+}
+
+type NamespaceLabelSelector struct{
```
hantmac, line 95: Use  `metav1.LabelSelector` to replace `NamespaceLabelSelector `  so that many functions can re-use.

docs/proposals/20220717-resourcedistribution-generator.md
```diff
@@ -0,0 +1,293 @@
...
+apiVersion: apps.kruise.io/v1alpha1
+kind: ResourceDistributionGenerator
+metadata:
+  name: rdname
+distributionObject: ConfigMap # Only configmap and secret are available
+distributionName: cmname 
+# The sources of the resource must be literals, files and envs
+files: 
+  - file.properties
+literals:
```
hantmac, line 167: What is the `literals ` mean?

docs/proposals/20220717-resourcedistribution-generator.md
dong4325: Good idea. I will modify it.

docs/proposals/20220717-resourcedistribution-generator.md
dong4325: I got it. Thank you.

docs/proposals/20220717-resourcedistribution-generator.md
dong4325: > literals is a list of literal pair sources. Each literal source should be a key and literal value, e.g. `key=value`. 

I will add this comment to it

docs/proposals/20220717-resourcedistribution-generator.md
```diff
@@ -0,0 +1,287 @@
...
+  NamespaceLabelSelector *metav1.LabelSelector `json:"namespaceLabelSelector,omitempty" yaml:"namespaceLabelSelector,omitempty"`
+}
+```
+
+### NameSuffixHash
+If both resourcedistribution and resource's names are suffixed with hash values, they will have the same hash value.
+
+### Weakness
+
+Using the go plugin requires that the main program and the plugin program have the same build environment. So, End users using shared plugins must compile both kustomize and plugin.
```
zmberg, line 108: Is it possible to use the [Containerized KRM Function](https://kubectl.docs.kubernetes.io/guides/extending_kustomize/containerized_krm_functions/) method? Does this mean that the user does not need to compile extra

PR #1034, opened 2022-07-23
Fix doc for `BroadcastJob`

astraw99:
The `RestartLimit` specifies the number of retries before marking the job failed, not the pod.
Also add the `Default is FailFast` doc for the `FailurePolicyType`.

NONE

### DIFF
--- a/apis/apps/v1alpha1/broadcastjob_types.go
+++ b/apis/apps/v1alpha1/broadcastjob_types.go
@@ -37,7 +37,7 @@ type BroadcastJobSpec struct {
 	Template v1.PodTemplateSpec `json:"template" protobuf:"bytes,2,opt,name=template"`
 
 	// CompletionPolicy indicates the completion policy of the job.
-	// Default is Always CompletionPolicyType
+	// Default is Always CompletionPolicyType.
 	// +optional
 	CompletionPolicy CompletionPolicy `json:"completionPolicy" protobuf:"bytes,3,opt,name=completionPolicy"`
 
@@ -52,8 +52,8 @@ type BroadcastJobSpec struct {
 
 // CompletionPolicy indicates the completion policy for the job
 type CompletionPolicy struct {
-	// Type indicates the type of the CompletionPolicy
-	// Default is Always
+	// Type indicates the type of the CompletionPolicy.
+	// Default is Always.
 	Type CompletionPolicyType `json:"type,omitempty" protobuf:"bytes,1,opt,name=type,casttype=CompletionPolicyType"`
 
 	// ActiveDeadlineSeconds specifies the duration in seconds relative to the startTime that the job may be active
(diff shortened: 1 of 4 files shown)


apis/apps/v1alpha1/broadcastjob_types.go
```diff
@@ -153,9 +153,10 @@ const (
 // FailurePolicy indicates the behavior of the job, when failed pod is found.
 type FailurePolicy struct {
 	// Type indicates the type of FailurePolicyType.
+	// Default is FailurePolicyTypeFailFast.
 	Type FailurePolicyType `json:"type,omitempty" protobuf:"bytes,1,opt,name=type,casttype=FailurePolicyType"`
 
-	// RestartLimit specifies the number of retries before marking the pod failed.
+	// RestartLimit specifies the number of retries before marking the job failed.
```
FillZpp, line 159: That is the failed pod, not the whole job.

FillZpp (commented):
Mostly LG, thanks.

apis/apps/v1alpha1/broadcastjob_types.go
astraw99: The pod is the worker, when the pod failed, it will be deleted and the job will create another new pod to retry.
If the pod failed times > `RestartLimit`, the parent job will be marked failed.

Is something missing or misunderstood?

FillZpp: No, broadcastjob will not recreate the pod when it fails.

It only compares `restartLimit` with the `restartCount` of the pod status. When a pod has a `restartCount` bigger than `restartLimit`, it will be considered as a failed pod.

https://github.com/openkruise/kruise/blob/286792678707b67bc9abff2a342195ececd3d8f2/pkg/controller/broadcastjob/broadcastjob_utils.go#L66-L85

astraw99: After checking the code and verify the demo yaml in my cluster, we can see the pod phase will be `Failed` and deleted immediately, not by calculating `RestartCount` in status.

The logic to get failed pod is here:
https://github.com/openkruise/kruise/blob/286792678707b67bc9abff2a342195ececd3d8f2/pkg/controller/broadcastjob/broadcastjob_utils.go#L44-L64

The logic goes into L50-L51: checking the pod phase is `Failed`, then the pod was added into failedPods, then
will go to the delete pod logic on code L282 and L294:
https://github.com/openkruise/kruise/blob/286792678707b67bc9abff2a342195ececd3d8f2/pkg/controller/broadcastjob/broadcastjob_controller.go#L273-L297

Please check the logic again, thanks.

FillZpp: Yeah, so in this doc it should be `pod` instead of `job`.

astraw99: OK, updated. PTAL thanks.
Also updated the website doc: https://github.com/openkruise/openkruise.io/pull/68

FillZpp (approved):
/lgtm

Thanks!

PR #1036, opened 2022-07-24
Fix go get in Makefile with go 1.18

### LINKED ISSUE #1035: [BUG] Makefile works fail on go 1.18
**What happened**:
In my env, the go version is `go 1.18.2`, when do the command `make manifests` in the master branch, got the error:
```
make manifests

rm -rf /Users/xxx/kruise/bin/controller-gen
go: creating new go.mod: module tmp
Downloading sigs.k8s.io/controller-tools/cmd/controller-gen@v0.7.0
go: added github.com/fatih/color v1.12.0
...
go: added sigs.k8s.io/controller-tools v0.7.0
go: added sigs.k8s.io/structured-merge-diff/v4 v4.1.2
go: added sigs.k8s.io/yaml v1.2.0
/Users/xxx/kruise/bin/controller-gen "crd:crdVersions=v1" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
bash: /Users/xxx/kruise/bin/controller-gen: No such file or directory
make: *** [manifests] Error 127
```
After check, it is the `controller-gen` binary not generated by `go get` in go 1.18.

To make the Makefile both work well on go version below or above go 1.18

1. Git clone the kruise master branch.
2. Set the go version with go 1.18.
3. Do `make manifests` or `make generate manifests`, then will get the error.

**Environment**:
- OS: Mac
- Go: go 1.18.2

astraw99:
From the Go 1.18 [release notes](https://tip.golang.org/doc/go1.18), we know that the `go get -d` is enabled by default,
then the target package binary will not install.

So this PR fix it to work well both below or above go 1.18 by using `go install` instead of `go get`.

fixes #1035

Change to defferent go version, like go 1.17 and go 1.18, `make manifests` will both work ok.

Updated `kustomize` version to make the `go install` work ok on go 1.18.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ endif
 
 KUSTOMIZE = $(shell pwd)/bin/kustomize
 kustomize: ## Download kustomize locally if necessary.
-	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v3@v3.8.7)
+	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4@v4.5.5)
 
 GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
 golangci-lint: ## Download golangci-lint locally if necessary.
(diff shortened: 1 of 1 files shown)


hantmac (approved):

FillZpp (approved):
/lgtm

PR #1042, opened 2022-07-28
WorkloadSpread support StatefulSet

veophi:

1. WorkloadSpread support StatefulSet
 - Only works when pod creating;
 - Will not interfere with the sequence of scale down
2. Consider missing replicas quota when managing existing pod


warmchang:
duplicate with https://github.com/openkruise/kruise/pull/1056?

PR #1043, opened 2022-07-28
Fix expectedUpdatedReplicas logic when CloneSet rolling back

veophi:

`ExpectedUpdatedReplicas` just is calculated based on `Partition`.


furykerry (commented):
/lgtm

PR #1046, opened 2022-08-02
Remove pod name label from CRR

veophi:

If pod name is more than 63 chars, the CRR creation will fail due to its pod name label. Hence, we remove this label from CRR.


FillZpp (approved):
/lgtm

PR #1065, opened 2022-08-25
feat(imagePullJob): add type for pullPolicy like k8s imagePullPolicy

mmmmmmmxl:
Signed-off-by: s1ark <maxinlin@vip.qq.com>

When you need to overwrite/preheat the existing image or latest tag image on the machine, you need to set a PullPolicy that is the same as the ImagePullPolicy of k8s.This PR provides configuration to solve this problem.

https://github.com/openkruise/kruise/issues/1267

Everythings is ok in our cluster.

### DIFF
--- a/apis/apps/v1alpha1/imagepulljob_types.go
+++ b/apis/apps/v1alpha1/imagepulljob_types.go
@@ -83,6 +83,11 @@ type ImagePullJobNodeSelector struct {
 
 // PullPolicy defines the policy of the pulling task
 type PullPolicy struct {
+	// Specifies the type of Pod ImagePullPolicy of the pulling task.
+	// Defaults to IfNotPresent
+	// +optional
+	Type PullPolicyType `json:"type,omitempty"`
+
 	// Specifies the timeout of the pulling task.
 	// Defaults to 600
 	// +optional
--- a/pkg/daemon/imagepuller/imagepuller_worker.go
+++ b/pkg/daemon/imagepuller/imagepuller_worker.go
@@ -424,11 +424,13 @@ func (w *pullWorker) doPullImage(ctx context.Context, newStatus *appsv1alpha1.Im
 	startTime := metav1.Now()
 
 	klog.Infof("Worker is starting to pull image %s:%s version %v", w.name, tag, w.tagSpec.Version)
-
-	if _, e := w.getImageInfo(ctx); e == nil {
-		klog.Infof("Image %s:%s is already exists", w.name, tag)
-		newStatus.Progress = 100
-		return nil
+	// When PullPolicy type is always, there is no need to check whether there is an image locally
+	if w.tagSpec.PullPolicy.Type != appsv1alpha1.PullAlways {
+		if _, e := w.getImageInfo(ctx); e == nil {
+			klog.Infof("Image %s:%s is already exists", w.name, tag)
+			newStatus.Progress = 100
+			return nil
+		}
 	}
 
 	// make it asynchronous for CRI runtime will block in pulling image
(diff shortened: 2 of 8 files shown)


apis/apps/v1alpha1/imagepulljob_types.go
```diff
@@ -83,6 +83,11 @@ type ImagePullJobNodeSelector struct {
 
 // PullPolicy defines the policy of the pulling task
 type PullPolicy struct {
+	// Specifies the type of Pod ImagePullPolicy of the pulling task.
+	// Defaults to IfNotPresent
+	// +optional
+	Type PullPolicyType `json:"type,omitempty"`
```
veophi, line 89: How about just reusing `PullPolicy` of  package "k8s.io/api/core/v1".

mmmmmmmxl: That's what I did at first, but Never seemed to make no sense in this scenario, so I tried defining a type in the same way as CompletionPolicyType

furykerry:
does the user need to change the imagepulljob if using always pull policy, if imagepulljob is not changed, image version will be the same, and if the existing image is pulled by the imagepuller, there will be a status record for the tag, and the kruise-daemon will not launch new imagepuller to pull the image.

mmmmmmmxl:
> does the user need to change the imagepulljob if using always pull policy, if imagepulljob is not changed, image version will be the same, and if the existing image is pulled by the imagepuller, there will be a status record for the tag, and the kruise-daemon will not launch new imagepuller to pull the image. Please provide a use case with corresponding yaml to demonstrate how one can use the always pull policy

Yes, if the user wants to use the always pull policy, he needs to update the ImagePullJob.
And If the user wants to update the latest image, he needs to use the following yaml.
`
kind: ImagePullJob
metadata:
  name: image-update-job-product
spec:
  image: xxx/prod-builder:latest
  parallelism: 5
  completionPolicy:
    type: Always
    activeDeadlineSeconds: 1200
    ttlSecondsAfterFinished: 600
  pullPolicy:
    type: Always
    backoffLimit: 3
    timeoutSeconds: 60
`

wqj97:
Why this feature not merge into master

zmberg:
@wqj97 Sorry, can you resolve the following failed checks, for example:
image

pkg/daemon/imagepuller/imagepuller_worker.go
```diff
@@ -424,11 +424,13 @@ func (w *pullWorker) doPullImage(ctx context.Context, newStatus *appsv1alpha1.Im
 	startTime := metav1.Now()
 
 	klog.Infof("Worker is starting to pull image %s:%s version %v", w.name, tag, w.tagSpec.Version)
-
-	if _, e := w.getImageInfo(ctx); e == nil {
-		klog.Infof("Image %s:%s is already exists", w.name, tag)
-		newStatus.Progress = 100
-		return nil
+	// When PullPolicy type is always, there is no need to check whether there is an image locally
+	if w.tagSpec.PullPolicy.Type != appsv1alpha1.PullAlways {
```
zmberg, line 428: I think you need to check whether the local image digest equals the registry.

wqj97:
> @wqj97 Sorry, can you resolve the following failed checks, for example: image
> 
> And suggest your go version is 1.18.x.

Sorry, are you mention the wrong person?

zmberg:
> > @wqj97 Sorry, can you resolve the following failed checks, for example: image
> > And suggest your go version is 1.18.x.
> 
> Sorry, did you mention the wrong person?

yes.

PR #1066, opened 2022-08-26
Performance optimization for workloadSpread when suitable subset maxReplicas is niil

veophi:

Optimize workloadSpread when suitable subset maxReplicas is `nil`.

In this situation, update status is meaningless because the subset is unlimited for replicas.


furykerry (commented):
/lgtm

PR #1071, opened 2022-09-08
Add PreNormal Lifecycle Hook for CloneSet

veophi:

PreNormal lifecycle hook allow users do something before Pod become available (normal), _e.g.,_ check whether the Pod IP has been added to SLB.

Now, the Pod lifecycle is:


apis/apps/pub/lifecycle.go
```diff
@@ -20,6 +20,7 @@ const (
 	LifecycleStateKey     = "lifecycle.apps.kruise.io/state"
 	LifecycleTimestampKey = "lifecycle.apps.kruise.io/timestamp"
 
+	LifecycleStatePreparingNormal LifecycleStateType = "PreparingNormal"
```
furykerry, line 23: plz comment in code about the meaning of each lifecycle state

pkg/controller/cloneset/sync/cloneset_scale.go
```diff
@@ -172,9 +172,9 @@ func (r *realControl) managePreparingDelete(cs *appsv1alpha1.CloneSet, pods, pod
 			continue
 		}
 
-		klog.V(3).Infof("CloneSet %s patch pod %s lifecycle from PreparingDelete to Normal",
+		klog.V(3).Infof("CloneSet %s patch pod %s lifecycle from PreparingDelete to Creating",
```
furykerry, line 175: Creating ->PreNormal

Actually since the users cancel the deletions of pods, we should logging explicitly that 

"CloneSet %s cancel deletion of pod %"

pkg/controller/cloneset/sync/cloneset_scale_test.go
```diff
@@ -84,7 +84,7 @@ func TestCreatePods(t *testing.T) {
 					apps.ControllerRevisionHashLabelKey:  "revision_abc",
 					apps.DefaultDeploymentUniqueLabelKey: "revision_abc",
 					"foo":                                "bar",
-					appspub.LifecycleStateKey:            string(appspub.LifecycleStateNormal),
```
furykerry, line 87: plz add tests for the pod under preNormal state, possible with Normal, deletion, inplace-update operation cases

furykerry (commented):
/lgtm

PR #1073, opened 2022-09-13
pod probe marker apis

zmberg:


furykerry (commented):
/lgtm

apis/apps/v1alpha1/node_pod_probe_types.go
```diff
@@ -0,0 +1,117 @@
...
+	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
+	// If Status=True, Message records the return result of Probe.
+	// If Status=False, Message records Probe's error message
+	Message string `json:"message,omitempty"`
+}
+
+type ProbeState string
+
+const (
+	ProbeTrue    ProbeState = "True"
```
FillZpp, line 85: `Succeeded`/`Failed` should be more intuitive than `True`/`False` as probe state.

apis/apps/v1alpha1/pod_probe_marker_types.go
```diff
@@ -0,0 +1,113 @@
...
+	// container name
+	ContainerName string `json:"containerName"`
+	// container probe spec
+	Probe ContainerProbeSpec `json:"probe"`
+	// According to the execution result of ContainerProbe, perform specific actions,
+	// such as: patch Pod labels, annotations, ReadinessGate Condition
+	MarkerPolicy []ProbeMarkerPolicy `json:"markerPolicy,omitempty"`
+}
+
+type ContainerProbeSpec struct {
```
FillZpp, line 49: Why not just use `v1.Probe`?

apis/apps/v1alpha1/node_pod_probe_types.go
```diff
@@ -0,0 +1,117 @@
...
+	PodProbeStatuses []PodProbeStatus `json:"podProbeStatuses,omitempty"`
+}
+
+type PodProbeStatus struct {
+	// pod name
+	Name string `json:"name"`
+	// pod namespace
+	Namespace string `json:"namespace,omitempty"`
+	// pod uid
+	Uid string `json:"uid"`
```
FillZpp, line 61: `UID`

apis/apps/v1alpha1/node_pod_probe_types.go
zmberg: The probe state feeds back to the POD condition, which is also viewed by the user, so it is better to keep the probe state consistent with the condition type

furykerry (commented):
/lgtm

PR #1074, opened 2022-09-13
Reduce update pod condition operations

veophi:

Ignore some case when update pod condition.


furykerry (commented):
/lgtm

PR #1083, opened 2022-09-21
WorkloadSpread consider whether patch field is matched when assign existing pods to subset

veophi:
…subset

WorkloadSpread consider whether patch field is matched when assign existing pods to subset

`preferredScore = 100 * schedulePreferredScore + 10 * patchPreferredScore + quotaScore`

`patchPreferredScore ++` if `subset.patch.annotations` or `subset.patch.labels` is matched with pod's.

### DIFF
--- a/pkg/controller/workloadspread/workloadspread_controller.go
+++ b/pkg/controller/workloadspread/workloadspread_controller.go
@@ -364,7 +364,7 @@ func (r *ReconcileWorkloadSpread) groupPod(ws *appsv1alpha1.WorkloadSpread, pods
 	// count managed pods for each subset
 	for i := range pods {
 		injectWS := getInjectWorkloadSpreadFromPod(pods[i])
-		if injectWS == nil || injectWS.Name != ws.Name || injectWS.Subset == "" {
+		if isNotMatchedWS(injectWS, ws) {
 			continue
 		}
 		if _, exist := podMap[injectWS.Subset]; !exist {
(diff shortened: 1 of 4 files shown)


pkg/controller/workloadspread/workloadspread_controller.go
```diff
@@ -426,7 +439,7 @@ func (r *ReconcileWorkloadSpread) getAndUpdateSuitableSubsetName(ws *appsv1alpha
 	for i := range ws.Spec.Subsets {
 		subset := &ws.Spec.Subsets[i]
 		// in case of that this pod was scheduled to the node which matches a subset of workloadSpread
-		matched, preferredScore, err := matchesSubset(pod, node, subset)
+		matched, preferredNodeScore, preferredPodScore, err := matchesSubset(pod, node, subset, subsetPatchFields[subset.Name])
```
furykerry, line 442: can we move subsetMissingReplicas checking inside matchesSubset, so that matchesSubset only return whether subset can be matched and corresponding matching score. In such way, all scoring logic is in a single function, much easier to test and understand

pkg/controller/workloadspread/workloadspread_controller_utils.go
```diff
@@ -85,18 +85,53 @@ func filterOutCondition(conditions []appsv1alpha1.WorkloadSpreadSubsetCondition,
...
+	isContained := func(a, b map[string]string) bool {
+		for k, v := range b {
+			if a[k] != v {
+				return false
+			}
+		}
+		return true
+	}
+
+	// currently, we only consider labels and annotations
```
furykerry, line 124: we can test for other field easily . We just clone the pod and apply the patch to the cloned pods, can compared the patched pod and original pod to see if they're identical

pkg/controller/workloadspread/workloadspread_controller_utils.go
```diff
@@ -85,18 +89,54 @@ func filterOutCondition(conditions []appsv1alpha1.WorkloadSpreadSubsetCondition,
...
+	// preferredNodeScore is in [0, total_prefer_weight]
+	preferredNodeScore := int64(0)
+	if subset.PreferredNodeSelectorTerms != nil {
+		nodePreferredTerms, _ := nodeaffinity.NewPreferredSchedulingTerms(subset.PreferredNodeSelectorTerms)
+		preferredNodeScore = nodePreferredTerms.Score(node)
 	}
 
-	nodePreferredTerms, _ := nodeaffinity.NewPreferredSchedulingTerms(subset.PreferredNodeSelectorTerms)
-	return matched, nodePreferredTerms.Score(node), nil
+	// preferredPodScore is in [0, 2]
```
furykerry, line 106: preferredPodScore is now in [0, 1]

FillZpp (approved):
/lgtm

PR #1084, opened 2022-09-22
sidecarset support pods ns(kube-system, kube-public)

zmberg:
Currently sidecarSet ignores the pods under kube-system and kube-public. However, some system components actually have similar requirements. So, remove the restriction.


furykerry (commented):
/lgtm

PR #1097, opened 2022-10-07
Support inplace-update restart count

### LINKED ISSUE #910: [feature request] add inPlaceUpdate restart count to annotation
 Add inPlaceUpdate restart count to annotation
**Why is this needed**:
Our developers use restartCount as a way to judge the status of service, but inPlaceUpdate or Container Restart will add restart count, it may make them confused.
Since it hard to change the logic of pod restart count, it's also useful to add restart Caused by kruise in annotation, then handle this in ui。

diannaowa:
Add inPlaceUpdate restart count to annotation.
the result is shown below：
Annotations:

```
apps.kruise.io/inplace-update-containers-restart-count: '{"nginx":{"revision":"sample-d97f89dcf","restartCount":1,"timestamp":"2022-03-24T08:03:23Z"},"sidecar":{"revision":"sample-d97f89dcf","restartCount":1,"timestamp":"2022-03-24T08:03:23Z"}}'
apps.kruise.io/inplace-update-pod-restart-count: "2"
```

apps.kruise.io/inplace-update-pod-restart-count: The Pod restart count.
apps.kruise.io/inplace-update-containers-restart-count: Every container restart count of the pod.
(revision: the revision of the container updated.)

The revision and timestamp maybe is different between the containers when we just update one(or a part of containers in the pod) container of the pod.Because we just update the restart count of the container when it will be updated(inplace-update).

fixes #910


apis/apps/pub/inplace_update.go
```diff
@@ -101,6 +107,16 @@ type InPlaceUpdateStrategy struct {
 	GracePeriodSeconds int32 `json:"gracePeriodSeconds,omitempty"`
 }
 
+// InPlaceUpdateContainerRestartCount record the restart count of containers
+type InPlaceUpdateContainerRestartCount struct {
+	// Revision is the updated revision hash.
+	Revision string `json:"revision,omitempty"`
+	// RestartCount is the container restart count
+	RestartCount int `json:"restartCount,omitempty"`
```
furykerry, line 115: is this field records the count of container at some revision ? if this is the case , cloneset controller can be too busy updating the restart counts,  actually currently the event handler will skip some event if the pod does not change its UpdateReady states. 

```	if coreControl.IsPodUpdateReady(oldPod, cs.Spec.MinReadySeconds) != coreControl.IsPodUpdateReady(curPod, cs.Spec.MinReadySeconds) {
		return false
	}
```
An alternative api design is to record the initial restart count of the container at some revision, cloneset controller only need to update the count during inplace-update, and the user can read the actually counts by decrease the container restart counts by the initial restart Count

FillZpp:
@diannaowa How about a discussion on the community meeting this Thursday? You can add this topic into the agenda https://shimo.im/docs/gXqmeQOYBehZ4vqo

zmberg:
https://github.com/openkruise/kruise/issues/910

apis/apps/pub/inplace_update.go
diannaowa: work on it

pkg/util/inplaceupdate/inplace_update.go
```diff
@@ -419,3 +420,59 @@ func hasEqualCondition(pod *v1.Pod, newCondition *v1.PodCondition) bool {
...
+				RestartCount: 1,
+				Timestamp:    metav1.NewTime(Clock.Now()),
+			}
+			currentPodRestartCount += 1 // pod restart
+		}
+		containersRestartCount[cname] = containerRestartCount // containers restart
+	}
+
+	containers := make(map[string]bool)
+	if spec.ContainerImages != nil {
```
zmberg, line 460: This if judgment can be removed, just keep the for statement below

pkg/util/inplaceupdate/inplace_update.go
```diff
@@ -419,3 +420,59 @@ func hasEqualCondition(pod *v1.Pod, newCondition *v1.PodCondition) bool {
...
+		for cname := range spec.ContainerImages {
+			containers[cname] = true
+			calculateRestartCountFunc(cname)
+		}
+	}
+
+	if spec.UpdateEnvFromMetadata {
```
zmberg, line 467: as above

pkg/util/inplaceupdate/inplace_update.go
```diff
@@ -216,13 +216,13 @@ func (c *realControl) finishGracePeriod(pod *v1.Pod, opts *UpdateOptions) (time.
 				delayDuration = roundupSeconds(graceDuration - span)
 				return nil
 			}
-
 			if clone, err = opts.PatchSpecToPod(clone, &spec, &updateState); err != nil {
 				return err
 			}
+			// record restart count of the pod(containers)
+			recordPodAndContainersRestartCount(clone, &spec, updateState.Revision)
```
zmberg, line 223: It's not an in-place upgrade here, and it doesn't feel like it should be judged

ABNER-1:
Hi @diannaowa 
I have few questions:

The modifications seem to target the `InPlaceUpdateContainersRestartKey` and `InPlaceUpdatePodRestartKey` primarily in the following scenarios:
   -  When patch a pod during an in-place update.
   -  When finish Grace Period.
   -  When update Next Batch container need to in-place update.
 

   Specifically, the action taken is to increase the restart count for containers which will in-place upgrade (the revision does not match the updated revision).

**Is it correct to interpret that these annotations are intended to log the restart counts of containers specifically due to in-place upgrades?**

I feel that tracking the restart count prior to each in-place upgrade does not provide meaningful data for the current version.

My expectation was that the restart counts for each container would be logged post-completion of an in-place upgrade, effectively resetting the count to zero. Then
`abnormal restart count for the current version's containers` = `restart count in status` -  `logged restart count in annotation`

PR #1101, opened 2022-10-14
fix sidecarset inject annotations abnormal

zmberg:
fix sidecarset inject annotations abnormal:
1. If the sidecarSet is not injected for the first time or re-injected into the sidecar cosntainer, there is no need to update annotations
2. When a Pod contains multiple hot upgrade containers, the current logic will override annotations [kruise.io/sidecarset-working-hotupgrade-container]


furykerry (commented):
/lgtm

PR #1108, opened 2022-10-25
fix InPlaceUpdateEnvFromMetadata bug

yiqinguo:

fix InPlaceUpdateEnvFromMetadata bug:

When injecting environment variables via annoation like:
```
- name: PKG_VERSION
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.annotations['DeployTime']
```
annotation DeployTime value is: 2022-10-19 17:55:29.892870834 +0800 CST m=+1.020884618.

Due to the Split,the container will be restarted constantly.


FillZpp (approved):
/lgtm

Thanks for fixing this!

PR #1113, opened 2022-11-01
enable mcloneset deleting pvc when pod hanging

### LINKED ISSUE #1099: [BUG] cloneset reuse terminating pod's name when pvc exists
**What happened**:
The node is not ready and pod with pvc on this node is evicted. But as the kubelet is unready so it doesn't really delete the pod and the pod stuck in terminating state.
cloneset try to scale out a new pod and it finds there is a pvc exsits, so it reuse the pod name referred by pvc to create the pod. But the terminating pod  already exists then it fails.

A new pod can be created successfully.

1. Create a cloneset with pvc
2. Stop the kubelet and wait 5min 

**Environment**:
- Kruise version: 1.0.2
- Others:

willise:
fixes #1099

fixes #1099

1. Create a cloneset with setting "cleanUpPVC=false" and PVC
2. Edit a pod created by  cloneset with a finalizer
3. Delete the above pod and verify the new pod can't be created with name conflict
4. Edit cloneset with setting "cleanUpPVC=true" 
5. Verify a new pod with new PVC has been created and old pod hangs in terminating state.

### DIFF
--- a/apis/apps/v1alpha1/cloneset_types.go
+++ b/apis/apps/v1alpha1/cloneset_types.go
@@ -93,6 +93,10 @@ type CloneSetScaleStrategy struct {
 	// The scale will fail if the number of unavailable pods were greater than this MaxUnavailable at scaling up.
 	// MaxUnavailable works only when scaling up.
 	MaxUnavailable *intstr.IntOrString `json:"maxUnavailable,omitempty"`
+
+	// Indicate if cloneset will reuse already existed pvc to
+	// rebuild a new pod
+	DisablePVCReuse bool `json:"disablePVCReuse,omitempty"`
 }
 
 // CloneSetUpdateStrategy defines strategies for pods update.
--- a/pkg/controller/cloneset/sync/cloneset_scale.go
+++ b/pkg/controller/cloneset/sync/cloneset_scale.go
@@ -23,6 +23,11 @@ import (
 	"sync"
 	"sync/atomic"
 
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+
 	appspub "github.com/openkruise/kruise/apis/apps/pub"
 	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
 	clonesetcore "github.com/openkruise/kruise/pkg/controller/cloneset/core"
@@ -30,10 +35,6 @@ import (
 	"github.com/openkruise/kruise/pkg/util"
 	"github.com/openkruise/kruise/pkg/util/expectations"
 	"github.com/openkruise/kruise/pkg/util/lifecycle"
-	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/util/rand"
-	"k8s.io/apimachinery/pkg/util/sets"
-	"k8s.io/klog/v2"
 )
 
 const (
--- a/pkg/controller/cloneset/utils/cloneset_utils.go
+++ b/pkg/controller/cloneset/utils/cloneset_utils.go
@@ -22,12 +22,6 @@ import (
 	"strings"
 	"sync"
 
-	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
-	"github.com/openkruise/kruise/pkg/features"
-	utilclient "github.com/openkruise/kruise/pkg/util/client"
-	"github.com/openkruise/kruise/pkg/util/expectations"
-	utilfeature "github.com/openkruise/kruise/pkg/util/feature"
-	"github.com/openkruise/kruise/pkg/util/requeueduration"
 	apps "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -36,6 +30,13 @@ import (
 	kubecontroller "k8s.io/kubernetes/pkg/controller"
 	"k8s.io/utils/integer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
+	"github.com/openkruise/kruise/pkg/features"
+	utilclient "github.com/openkruise/kruise/pkg/util/client"
+	"github.com/openkruise/kruise/pkg/util/expectations"
+	utilfeature "github.com/openkruise/kruise/pkg/util/feature"
+	"github.com/openkruise/kruise/pkg/util/requeueduration"
 )
 
 var (
@@ -93,20 +94,52 @@ func GetControllerKey(cs *appsv1alpha1.CloneSet) string {
 
 // GetActivePods returns all active pods in this namespace.
 func GetActivePods(reader client.Reader, opts *client.ListOptions) ([]*v1.Pod, error) {
-	podList := &v1.PodList{}
-	if err := reader.List(context.TODO(), podList, opts, utilclient.DisableDeepCopy); err != nil {
+	podList, err := GetAllPods(reader, opts)
+	if err != nil {
 		return nil, err
 	}
 
 	// Ignore inactive pods
 	var activePods []*v1.Pod
-	for i, pod := range podList.Items {
+	for i := range podList {
+		pod := podList[i]
 		// Consider all rebuild pod as active pod, should not recreate
-		if kubecontroller.IsPodActive(&pod) {
-			activePods = append(activePods, &podList.Items[i])
+		if kubecontroller.IsPodActive(pod) {
+			activePods = append(activePods, pod)
 		}
 	}
 	return activePods, nil
+}
+
+func GetActiveAndInactivePods(reader client.Reader, opts *client.ListOptions) ([]*v1.Pod, []*v1.Pod, error) {
+	podList, err := GetAllPods(reader, opts)
+	if err != nil {
+		return nil, nil, err
+	}
+	var activePods, inactivePods []*v1.Pod
+	for i := range podList {
+		pod := podList[i]
+		if kubecontroller.IsPodActive(pod) {
+			activePods = append(activePods, pod)
+		} else {
+			inactivePods = append(inactivePods, pod)
+		}
+	}
+	return activePods, inactivePods, nil
+}
+
+// GetAllPods returns all pods in this namespace.
+func GetAllPods(reader client.Reader, opts *client.ListOptions) ([]*v1.Pod, error) {
+	podList := &v1.PodList{}
+	if err := reader.List(context.TODO(), podList, opts, utilclient.DisableDeepCopy); err != nil {
+		return nil, err
+	}
+
+	var pods []*v1.Pod
+	for i := range podList.Items {
+		pods = append(pods, &podList.Items[i])
+	}
+	return pods, nil
 }
 
 // NextRevision finds the next valid revision number based on revisions. If the length of revisions
@@ -164,6 +197,10 @@ func UpdateStorage(cs *appsv1alpha1.CloneSet, pod *v1.Pod) {
 		}
 	}
 	pod.Spec.Volumes = newVolumes
+}
+
+func GetInstanceID(obj metav1.Object) string {
+	return obj.GetLabels()[appsv1alpha1.CloneSetInstanceID]
 }
 
 // GetPersistentVolumeClaims gets a map of PersistentVolumeClaims to their template names, as defined in set. The
--- a/pkg/util/ownerref.go
+++ b/pkg/util/ownerref.go
@@ -1,0 +1,62 @@
+/*
+Copyright 2022 The Kruise Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func HasOwnerRef(target, owner metav1.Object) bool {
+	ownerUID := owner.GetUID()
+	for _, ownerRef := range target.GetOwnerReferences() {
+		if ownerRef.UID == ownerUID {
+			return true
+		}
+	}
+	return false
+}
+
+func RemoveOwnerRef(target, owner metav1.Object) bool {
+	if !HasOwnerRef(target, owner) {
+		return false
+	}
+	ownerUID := owner.GetUID()
+	oldRefs := target.GetOwnerReferences()
+	newRefs := make([]metav1.OwnerReference, len(oldRefs)-1)
+	skip := 0
+	for i := range oldRefs {
+		if oldRefs[i].UID == ownerUID {
+			skip = -1
+		} else {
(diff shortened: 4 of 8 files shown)


veophi:
How about taking the PVCs related with Competed Pods(`Phase=Failed` || `Phase=Succeeded`)?

zmberg:
@willise If the Pod is in the Terminating, it is still risky to delete the PVC directly. The difficulty with this way is that we can't quite tell if it's a normal process or a Hang. I agree with @FillZpp 's idea of adding AlwaysCreatePVC in spec.scaleStrategy to decide whether it should reuse the instanceId of existing free PVCs when creating new Pod.

willise:
> How about taking the PVCs related with Competed Pods(`Phase=Failed` || `Phase=Succeeded`)?

Thank you for your reply. I am not sure if I understand correctly. Now the `Scale` function has "active" pods and "active" pvcs as parameters. If the pod is in terminating state, it's a little hard to get it except I have another function to get all pods.

willise:
> @willise If the Pod is in the Terminating, it is still risky to delete the PVC directly. The difficulty with this way is that we can't quite tell if it's a normal process or a Hang. I agree with @FillZpp 's idea of adding AlwaysCreatePVC in spec.scaleStrategy to decide whether it should reuse the instanceId of existing free PVCs when creating new Pod.

I got what you mean and I think it is a good advice. But if `AlwaysCreatePVC` is set, cloneset controller only creates new pod with new PVC. So what to do with existing free PVCs? Like mentioned in the issue, should cloneset controller leave these PVCs unprocessed? @zmberg

zmberg:
Your current cleanup logic works, except that you need to add an additional judgment that only Pods that really don't exist anymore can be cleaned up.

apis/apps/v1alpha1/cloneset_types.go
```diff
@@ -93,6 +93,9 @@ type CloneSetScaleStrategy struct {
 	// The scale will fail if the number of unavailable pods were greater than this MaxUnavailable at scaling up.
 	// MaxUnavailable works only when scaling up.
 	MaxUnavailable *intstr.IntOrString `json:"maxUnavailable,omitempty"`
+
+	// Enable cloneset to clean up the useless pvc when pod hangs in terminating state.
+	CleanUpPVC bool `json:"cleanUpPVC,omitempty"`
```
FillZpp, line 98: How about `DisablePVCReuse`? So that controller will not reuse existing PVCs when creating new Pods and cleanup those useless PVCs whose Pods has already been deleted or in Succeeded/Failed phases.

willise: @zmberg @FillZpp Fi I update the pr with following the comments.

willise:
Also I find the main logic when scaling in, pod deletes the pvc directly. Maybe I can update the ownerreference of the pod from cloneset to the pod and if the pod is deleted, pvc will be deleted by background cascading deletion later after the pod disappears finally. Just like https://github.com/kubernetes/kubernetes/blob/4086b45af3761d59cb82af6ee427d2d6557c1cbc/pkg/controller/statefulset/stateful_set_utils.go#L231 does.

FillZpp:
> Also I find the main logic when scaling in, pod deletes the pvc directly. Maybe I can update the ownerreference of the pod from cloneset to the pod and if the pod is deleted, pvc will be deleted by background cascading deletion later after the pod disappears finally. Just like https://github.com/kubernetes/kubernetes/blob/4086b45af3761d59cb82af6ee427d2d6557c1cbc/pkg/controller/statefulset/stateful_set_utils.go#L231 does.

+1, I think it's a good idea. WDYT @zmberg @veophi

zmberg:
@willise please add UT and e2e test.

zmberg:
> Also I find the main logic when scaling in, pod deletes the pvc directly. Maybe I can update the ownerreference of the pod from cloneset to the pod and if the pod is deleted, pvc will be deleted by background cascading deletion later after the pod disappears finally. Just like https://github.com/kubernetes/kubernetes/blob/4086b45af3761d59cb82af6ee427d2d6557c1cbc/pkg/controller/statefulset/stateful_set_utils.go#L231 does.

Good idea, and please add UT.

willise:
@zmberg @FillZpp I finish the code and please review :).

pkg/controller/cloneset/sync/cloneset_scale.go
```diff
@@ -60,6 +65,19 @@ func (r *realControl) Scale(
 		return false, nil
 	}
 
+	// If cloneset doesn't want to reuse pvc, clean up
+	// the existing pvc first. Then it looks like the pod
+	// is deleted by controller, new pod can be created.
+	if updateCS.Spec.ScaleStrategy.DisablePVCReuse {
+		uselessPVCs := getUselessPVCs(pods, pvcs)
```
zmberg, line 72: Should remove uselessPVCs in pvcs? then, pvcs only contains used pvcs.

pkg/controller/cloneset/sync/cloneset_scale.go
```diff
@@ -403,3 +421,164 @@ func (r *realControl) choosePodsToDelete(cs *appsv1alpha1.CloneSet, totalDiff in
...
+				}
+				modified = true
+			}
+			// Left pvcs will be deleted directly.
+			delete(uselessPVCs, instanceID)
+		}
+	}
+
+	for _, pvc := range uselessPVCs {
+		if clonesetutils.IsTerminating(pvc) {
```
zmberg, line 478: I don’t think it’s necessary to getInactivePods in the front, just get the pod here, update the owner ref if it exists, delete it if it doesn’t exist.

pkg/controller/cloneset/sync/cloneset_scale.go
```diff
@@ -403,3 +421,164 @@ func (r *realControl) choosePodsToDelete(cs *appsv1alpha1.CloneSet, totalDiff in
...
+		clonesetutils.ScaleExpectations.ObserveScale(clonesetutils.GetControllerKey(cs), expectations.Delete, pvc.Name)
+		r.recorder.Eventf(cs, v1.EventTypeWarning, "FailedDelete", "failed to clean up PVC %s: %v", pvc.Name, err)
+		return modified, err
+	}
+	return true, nil
+}
+
+func updateClaimOwnerRefToPod(pvc *v1.PersistentVolumeClaim, cs *appsv1alpha1.CloneSet, pod *v1.Pod) bool {
+	needsUpdate := false
+	updateMeta := func(tm *metav1.TypeMeta) {
```
zmberg, line 571: Please use metav1.NewControllerRef() and you can realize the common function in utils like https://github.com/openkruise/rollouts/blob/master/pkg/util/condition.go.

pkg/controller/cloneset/sync/cloneset_scale.go
willise: OK, I will fix it.

zmberg:
@willise If it is convenient for you, you can communicate on the zoom of the community meeting on 12.22.

pkg/controller/cloneset/sync/cloneset_scale.go
```diff
@@ -403,3 +420,101 @@ func (r *realControl) choosePodsToDelete(cs *appsv1alpha1.CloneSet, totalDiff in
...
+		return modified, err
+	}
+
+	// If useless pvc owner pod does not exist, the pvc can be deleted directly,
+	// else update pvc's ownerreference to pod.
+	for _, pod := range pods {
+		instanceID := clonesetutils.GetInstanceID(pod)
+		if pvc, ok := uselessPVCs[instanceID]; ok {
+			if updateClaimOwnerRefToPod(pvc, cs, pod) {
+				if modified, err := r.updatePVC(cs, pvc); err != nil && !errors.IsNotFound(err) {
```
zmberg, line 462: if modified, err = r.updatePVC(cs, pvc); err != nil && !errors.IsNotFound(err) {

pkg/controller/cloneset/sync/cloneset_scale.go
```diff
@@ -403,3 +420,101 @@ func (r *realControl) choosePodsToDelete(cs *appsv1alpha1.CloneSet, totalDiff in
...
+	pods, err := getAllPods(r.Client, cs)
+	if err != nil {
+		klog.Errorf("Could not get cloneset %s owned pods", clonesetutils.GetControllerKey(cs))
+		return modified, err
+	}
+
+	// If useless pvc owner pod does not exist, the pvc can be deleted directly,
+	// else update pvc's ownerreference to pod.
+	for _, pod := range pods {
+		instanceID := clonesetutils.GetInstanceID(pod)
```
zmberg, line 459:
```
if kubecontroller.IsPodActive(pod) {
	continue
}
```

pkg/controller/cloneset/sync/cloneset_scale.go
```diff
@@ -403,3 +420,101 @@ func (r *realControl) choosePodsToDelete(cs *appsv1alpha1.CloneSet, totalDiff in
...
+				modified = true
+			}
+			// Left pvcs will be deleted directly.
+			delete(uselessPVCs, instanceID)
+		}
+	}
+
+	for _, pvc := range uselessPVCs {
+		// It's safe to delete pvc that has no pod found.
+		if modified, err := r.deletePVC(cs, pvc); err != nil && !errors.IsNotFound(err) {
```
zmberg, line 474: if modified, err = r.deletePVC(cs, pvc); err != nil && !errors.IsNotFound(err) {

zmberg:
@willise In addition, Can you add e2e for this?

apis/apps/v1alpha1/cloneset_types.go
```diff
@@ -93,6 +93,10 @@ type CloneSetScaleStrategy struct {
 	// The scale will fail if the number of unavailable pods were greater than this MaxUnavailable at scaling up.
 	// MaxUnavailable works only when scaling up.
 	MaxUnavailable *intstr.IntOrString `json:"maxUnavailable,omitempty"`
+
+	// Indicate if cloneset will reuse aleady existed pvc to
```
zmberg, line 97: aleady -> already

pkg/controller/cloneset/utils/cloneset_utils.go
```diff
@@ -93,22 +92,36 @@ func GetControllerKey(cs *appsv1alpha1.CloneSet) string {
...
-	if err := reader.List(context.TODO(), podList, opts, utilclient.DisableDeepCopy); err != nil {
+	podList, err := GetAllPods(reader, opts)
+	if err != nil {
 		return nil, err
 	}
 
 	// Ignore inactive pods
 	var activePods []*v1.Pod
-	for i, pod := range podList.Items {
+	for i, pod := range podList {
```
zmberg, line 102: I'm very worried that there will be problems with pointers, so I suggest:
```
       for i := range podList {
		pod := podList[i]
		// Consider all rebuild pod as active pod, should not recreate
		if kubecontroller.IsPodActive(pod) {
			activePods = append(activePods, pod)
		}
	}

```

pkg/util/ownerref_test.go
```diff
@@ -0,0 +1,154 @@
+/*
+Copyright 2020 The Kruise Authors.
```
zmberg, line 2: Copyright 2022 The Kruise Authors.

pkg/util/ownerref.go
```diff
@@ -0,0 +1,66 @@
+/*
+Copyright 2021 The Kruise Authors.
```
zmberg, line 2: Copyright 2022 The Kruise Authors.

pkg/controller/cloneset/sync/cloneset_scale.go
```diff
@@ -60,6 +64,19 @@ func (r *realControl) Scale(
...
 
+	// If cloneset doesn't want to reuse pvc, clean up
+	// the existing pvc first. Then it looks like the pod
+	// is deleted by controller, new pod can be created.
+	if updateCS.Spec.ScaleStrategy.DisablePVCReuse {
+		uselessPVCs := getUselessPVCs(pods, pvcs)
+		if len(uselessPVCs) > 0 {
+			klog.V(3).Infof("Begin to clean up cloneset %s useless PVCs", controllerKey)
+			if modified, err := r.cleanupPVCs(updateCS, uselessPVCs); err != nil || modified {
+				return modified, err
```
furykerry, line 75: if pvc is modified, why return? I think the controller can keep reconciling.

pkg/controller/cloneset/sync/cloneset_scale.go
```diff
@@ -403,3 +420,101 @@ func (r *realControl) choosePodsToDelete(cs *appsv1alpha1.CloneSet, totalDiff in
...
+			useless[id] = pvc
+		}
+	}
+	return useless
+}
+
+func (r *realControl) cleanupPVCs(cs *appsv1alpha1.CloneSet, uselessPVCs map[string]*v1.PersistentVolumeClaim) (bool, error) {
+	var modified bool
+
+	pods, err := getAllPods(r.Client, cs)
```
furykerry, line 450: why list again, can we use the pods param in Scale directly?

pkg/controller/cloneset/sync/cloneset_scale.go
```diff
@@ -403,3 +420,101 @@ func (r *realControl) choosePodsToDelete(cs *appsv1alpha1.CloneSet, totalDiff in
...
+	if err != nil {
+		klog.Errorf("Could not get cloneset %s owned pods", clonesetutils.GetControllerKey(cs))
+		return modified, err
+	}
+
+	// If useless pvc owner pod does not exist, the pvc can be deleted directly,
+	// else update pvc's ownerreference to pod.
+	for _, pod := range pods {
```
furykerry, line 458: can we really find a match pod for "useless" pvc? if we can is the pvc really useless in the begining

pkg/controller/cloneset/sync/cloneset_scale.go
willise: In the existing logic, if there is a resource change, the controller will return.

pkg/controller/cloneset/sync/cloneset_scale.go
willise: Active pods are passed in from the Scale, so there is no way to get pods that are in the terminating state.

pkg/controller/cloneset/sync/cloneset_scale.go
willise: "useless" here means that the pvc can be deleted finally, e.g. it's owner pod is stuck in terminating state, the pvc should be deleted after the pod is cleaned.

pkg/controller/cloneset/sync/cloneset_scale.go
furykerry: I suppose the "existing logic" in the comment is this code block,

'''
	podsSpecifiedToDelete, podsInPreDelete, numToDelete := getPlannedDeletedPods(updateCS, pods)
	if modified, err := r.managePreparingDelete(updateCS, pods, podsInPreDelete, numToDelete); err != nil || modified {
		return modified, err
	}
'''
It re-enqueues not because it has mark some pods to delete, but because preparing-to-delete pods has to be processed by other operators first. 

In general, if one try to modify multiple underlying cloneset resources, pods, pvc etc) , modification can be made in one reconcile loop, so as to speed the reconcilation

pkg/controller/cloneset/sync/cloneset_scale.go
furykerry: it make senses now, yet since the Scale already pass pods variables, list pods again can break the code encapsulation logic, maybe we can move the pvc cleanup logic up onto doReconcile

pkg/controller/cloneset/sync/cloneset_scale.go
willise: Thank you for your explanation, this can indeed be optimized.

pkg/controller/cloneset/sync/cloneset_scale.go
willise: I have considered this. The reason for putting it in scale is because doReconcile seems to do some data preparation work. In fact, if “all pods” can be passed into Scale then there is no need to list pods, but the cost of changing the interface is a bit high.

Issue #1038, opened 2022-07-26
[feature request] partition support config like 100% - num


shiyan2016:
**What would you like to be added**:

partition support config like 100% - num

**Why is this needed**:
In some case, we need canary release, which expect num(for example num=1) of  pods always keep the newest version even if hpa triggle scale  in the canary release time.

veophi:
sounds good!

How about to support both `100%-num` and `100%-percent` ?

FillZpp:
This should be a topic of the community meeting this week :)

shiyan2016:
After discussion，partition will support config with float to improvement precision.

PR #1124, opened 2022-11-15
cloneset partition support float percent

shiyan2016:
support float percent for cloneset partition to improve presicion

### DIFF
--- a/pkg/util/tools_test.go
+++ b/pkg/util/tools_test.go
@@ -285,6 +285,20 @@ func TestCalculatePartitionReplicas(t *testing.T) {
 			expectedValue: 0,
 			succeeded:     false,
 		},
+		{
+			name:          `replicas=9, partition=0.01%, expected=1, err==nil`,
+			replicas:      pointer.Int32(9),
+			partition:     &intstr.IntOrString{Type: intstr.String, StrVal: "0.01%"},
+			expectedValue: 1,
+			succeeded:     true,
+		},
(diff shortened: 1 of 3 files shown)


FillZpp:
fixes #1038

pkg/util/tools_test.go
```diff
@@ -299,3 +299,79 @@ func TestCalculatePartitionReplicas(t *testing.T) {
...
+		},
+		{
+			input:     intstr.FromString("#%"),
+			expectErr: true,
+		},
+		{
+			input:     intstr.FromString("90"),
+			expectErr: true,
+		},
+	}
```
zmberg, line 360: Please add 0.01% partition UT.

veophi:
You should also update the `GetScaledValueFromIntOrPercent ` func in CloneSet controller.

shiyan2016:
> 



> You should also update the `GetScaledValueFromIntOrPercent ` func in CloneSet controller.

```
minUpdatedReadyPodsCount, err = intstrutil.GetScaledValueFromIntOrPercent(&minUpdatedReadyPodsIntStr, int(*instance.Spec.Replicas), true) 
```
This is not partition feature, change it also?

pkg/util/tools_test.go
shiyan2016: done

furykerry (commented):
/approve

PR #1125, opened 2022-11-17
improve error hanlding of inplace update env from metadata

veophi:

Improve robustness for some codes


pkg/controller/containerrecreaterequest/crr_controller.go
```diff
@@ -60,6 +61,10 @@ var (
 	controllerKind       = appsv1alpha1.SchemeGroupVersion.WithKind("ContainerRecreateRequest")
 )
 
+func newPodReadinessControl(adp podadapter.Adapter) utilpodreadiness.Interface {
+	return utilpodreadiness.NewForAdapter(adp)
```
zmberg, line 65: Why re-define the func newPodReadinessControl, other than immediate call utilpodreadiness.NewForAdapter(adp)?

veophi: done

PR #1141, opened 2022-12-12
feat: add lifecycle preNormal for asts

wkai886:

add lifecycle preNormal for asts

### DIFF
--- a/pkg/controller/statefulset/stateful_set_control.go
+++ b/pkg/controller/statefulset/stateful_set_control.go
@@ -527,7 +527,12 @@ func (ssc *defaultStatefulSetControl) updateStatefulSet(
 				}
 			}
 
-			lifecycle.SetPodLifecycle(appspub.LifecycleStateNormal)(replicas[i])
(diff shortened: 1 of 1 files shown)


pkg/controller/statefulset/stateful_set_control.go
```diff
@@ -832,6 +837,12 @@ func (ssc *defaultStatefulSetControl) refreshPodState(set *appsv1beta1.StatefulS
 
 	var state appspub.LifecycleStateType
 	switch lifecycle.GetPodLifecycleState(pod) {
+	case appspub.LifecycleStatePreparingNormal:
```
veophi, line 840: There are two things we need to consider further:

1. This state transition is invoked after successfully scaling. Assuming that the scale is stuck,  this transition will also be stuck. Is it reasonable?

2. If a pod-x has not entered Normal state yet after creating,  should we continue to create pod-x+1?

ABNER-1:
Notice this PR has no updates recently, I will fixes this issue in PR https://github.com/openkruise/kruise/pull/1858.
It will rebase this PR changes and add more ut cases.

PR #1149, opened 2022-12-20
Make OnFailure as default restartPolicy for BroadcastJob

Shubhamurkade:
Creation of a _BroadcastJob_ without specifying _restartPolicy_ fails with the following error:

`Error from server: error when creating "broadcastjob_kruise.yaml": admission webhook "vbroadcastjob.kb.io" denied the request: spec.template.spec.restartPolicy: Invalid value: "Always": pod restartPolicy can only be Never or OnFailure`

**Root Cause:**
_restartPolicy_ is present under spec.template.spec block. This block is defined with _PodSpec_ structure and the _restartPolicy_ element in it has a default value of _Always_.

**Solution:**
For _BroadcastJob_ changed the default value to _OnFailure_ in `/Users/surkade/Documents/dev/kruise/pkg/webhook/broadcastjob/validating/broadcastjob_create_update_handler.go` file

No

Create a sample _BroadcastJob_ with the following template (Do not specify restartPolicy under _spec.template.spec_:
```
kind: BroadcastJob
metadata:
    labels:
        app: <name>
    name: <name>
spec:
    template:
        metadata:
            labels:
                app: <name>
        spec:
            containers:
            - image: <image>
              imagePullPolicy: <policy>
              name: <name>
              ports:
              - containerPort: <port>
```
The previously mentioned error can be seen

I have kept the new default for _restartPolicy_ as _OnFailure_. Please let me know if _Never_ would make more sense as a default.

### DIFF
--- a/apis/apps/defaults/v1alpha1.go
+++ b/apis/apps/defaults/v1alpha1.go
@@ -192,6 +192,11 @@ func SetDefaultsBroadcastJob(obj *v1alpha1.BroadcastJob, injectTemplateDefaults 
 	if obj.Spec.FailurePolicy.Type == "" {
 		obj.Spec.FailurePolicy.Type = v1alpha1.FailurePolicyTypeFailFast
 	}
+
+	// Default to 'OnFailure' if no restartPolicy is specified
+	if obj.Spec.Template.Spec.RestartPolicy == "" {
+		obj.Spec.Template.Spec.RestartPolicy = corev1.RestartPolicyOnFailure
+	}
 }
 
 // SetDefaults_UnitedDeployment set default values for UnitedDeployment.


veophi:
@Shubhamurkade Maybe, you should modify this function instead of event handler :)  https://github.com/openkruise/kruise/blob/1ece7fe54f65fdd5cfb38bad5d93711541445991/apis/apps/defaults/v1alpha1.go#L178.

Shubhamurkade:
> @Shubhamurkade Maybe, you should modify this function instead of event handler :)
> 
> https://github.com/openkruise/kruise/blob/1ece7fe54f65fdd5cfb38bad5d93711541445991/apis/apps/defaults/v1alpha1.go#L178
> 
> .

@veophi Thanks for the suggestion. Will do!

Shubhamurkade:
> @Shubhamurkade Maybe, you should modify this function instead of event handler :)
> 
> https://github.com/openkruise/kruise/blob/1ece7fe54f65fdd5cfb38bad5d93711541445991/apis/apps/defaults/v1alpha1.go#L178
> 
> .

Adding it into _SetDefaultsBroadcastJob_ may not work because the function is called after [Decode](https://github.com/openkruise/kruise/blob/1ece7fe54f65fdd5cfb38bad5d93711541445991/pkg/webhook/broadcastjob/mutating/broadcastjob_create_update_handler.go#L52)

So, it'll just overwrite any user-specified value for _restartPolicy_ field.

Shubhamurkade:
@veophi Please let me know if the latest changes look fine.

zmberg:
@Shubhamurkade It's good, and you should _git commit -s and push again_.

zmberg:
@Shubhamurkade You need **git commit -s** and squash the commits to one.

zmberg:
@Shubhamurkade tks, and you must resolve the checks. https://github.com/openkruise/kruise/pull/1149/checks?check_run_id=11663449687

Shubhamurkade:
> @Shubhamurkade tks, and you must resolve the checks. https://github.com/openkruise/kruise/pull/1149/checks?check_run_id=11663449687

Please check now.

PR #1152, opened 2022-12-27
move sidecarSet control to openkruise/utils repo

zmberg:


pigletfly:
Is sidecarSet control used in other repos?

PR #1158, opened 2023-01-06
when DisablePVCReuse=true, set cloneSet pvc owner to pod

zmberg:
1. Add ut and e2e for DisablePVCReuse
2. optimize some code

### DIFF
--- a/pkg/controller/cloneset/cloneset_controller.go
+++ b/pkg/controller/cloneset/cloneset_controller.go
@@ -45,6 +45,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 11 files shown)


pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -237,14 +237,15 @@ func (r *ReconcileCloneSet) doReconcile(request reconcile.Request) (res reconcil
...
 	filteredPVCs, err := r.getOwnedPVCs(instance)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
 
-	// If mcloneset doesn't want to reuse pvc, clean up
-	// the existing pvc first. Then it looks like the pod
-	// is deleted by controller, new pod can be created.
+	// If cloneSet doesn't want to reuse pvc, clean up
+	// the existing pvc first, which owned by inactive pods.
```
furykerry, line 247: which owned by inactive pods -> which are owned by inactive or deleted pods. 
> Then it looks like the pod is deleted by controller, new pod can be created. 
just delete this comment which is quite confusing

furykerry (commented):
/lgtm

PR #1164, opened 2023-01-12
fix the bug that pub calculates pods(deployment) totalReplicas incorrectly

zmberg:
When PUB protects the Deployment, the pub controller needs to take Rs into account when calculating the totalReplicas corresponding to the Pod. 
However, the current calculation method incorrectly calculates rs.spec.replicas, for example: deployment.spec.replicas=10, which contains two rs, and the pub controller will calculate totalReplicas = 20.

```yaml
apiVersion: policy.kruise.io/v1alpha1
kind: PodUnavailableBudget
metadata:
  name: test1-pub
spec:
  maxUnavailable: 25%
  selector:
    matchLabels:
      app-name: test1
status:
  currentAvailable: 10
  desiredAvailable: 15
  totalReplicas: 20
  unavailableAllowed: 0
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test1
spec:
  replicas: 10
  selector:
    matchLabels:
      app-name: test1
  template:
    metadata:
      labels:
        app-name: test1
```
As shown above, pub.status.totalReplicas should be 10.


furykerry:
plz enrich commit message and PR discription

furykerry (commented):
/lgtm

PR #1186, opened 2023-02-21
move inplace update AdditionalFuncs after patch

shiyan2016:


zmberg:
@shiyan2016 Why move the funcs? Please add some comments.

PR #1209, opened 2023-03-08
chore: simplify comparison and variable declaration

hezhizhen:

Simplify some code, mainly comparison and variable declaration.

lint and test

### DIFF
--- a/pkg/controller/broadcastjob/broadcastjob_controller.go
+++ b/pkg/controller/broadcastjob/broadcastjob_controller.go
@@ -123,10 +123,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	}
 
 	// Watch for changes to Node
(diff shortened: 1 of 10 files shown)


pkg/controller/broadcastjob/broadcastjob_controller.go
```diff
@@ -178,7 +175,8 @@ func (r *ReconcileBroadcastJob) Reconcile(_ context.Context, request reconcile.R
 	addLabelToPodTemplate(job)
 
 	if IsJobFinished(job) {
-		if isPast, leftTime := pastTTLDeadline(job); isPast {
+		isPast, leftTime := pastTTLDeadline(job)
```
zmberg, line 178: Why you need to change here?

hezhizhen: I wanted to write something like 

```go
if !isPast && leftTime>0 { return xxx }

klog.Infof(xxx)
...
return xxx
```

But I was not sure of its correctness, so I reverted it, but this modification was kept.

PR #1247, opened 2023-04-07
feat: Optimize UnitedDeployment replicas settings

### LINKED ISSUE #1220: [feature request] Optimize UnitedDeployment replicas settings
Currently, UnitedDeployment requires the sum of `subsets[x].replicas` must equal to `spec.replicas`. Actually, it is very clumsy.

We expect that：
- If `spec.replicas` is not set (`nil`),  `subsets[i]` just follows `subsets[i].replicas` (should validate all of subsets replicas are not `nil`);
- If `spec.replicas` is set (not `nil`), we should allow at most **ONE** subset replicas is `nil`, and it should equals to `spec.replicas - <sum of other subset replicas>`;

y-ykcir:
Optimize UnitedDeployment replicas settings:
add optimization for the situation when `spec.replicas` is not set (`nil`)

fixed #1220


y-ykcir:
@veophi PTAL

PR #1275, opened 2023-05-06
fix the default value of imagePullPolicy

### LINKED ISSUE #1270: [BUG] sidecarSet in-place update failed
**What happened**:
After changing the container image of sidecarSet, the container that has been injected is not restarted and it's image is not updated
The injected container is restarted and its image becomes up to date
1. declare a SidecarSet(ss) that contains only a container with an image: `nginx`, then declare a Pod that will be injected into the side car, and subsequently see that the nginx container is injected normally
2. change the image of SidecarSet from `nginx` to `nginx:1.23` (just add a tag)
3. pod's nginx container is not restarted, and its image is not updated to `nginx:1.23` 

**Environment**:
- Kruise version: v1.2.0
- Kubernetes version (use `kubectl version`): v1.22.3
- Others:

ktalg:
Fixed an issue that caused sidecarset in-place update to not work in some scenarios

fix #1270 

just see the unit test

The setting of default value of the container's imagePullPolicy in sidecarSet is [affected by the image tag](https://github.com/kubernetes/kubernetes/blob/8f3f997a98c7b44925f7654e5c66e6f46e328147/pkg/apis/core/v1/defaults.go#L73-L83C3) , which makes it possible to change the value of SidecarSetHashWithoutImage even if only the image tag is modified, thus making the in-place upgrade of the container ineffective.

### DIFF
--- a/apis/apps/defaults/v1alpha1.go
+++ b/apis/apps/defaults/v1alpha1.go
@@ -27,8 +27,25 @@ import (
 )
 
 // SetDefaults_SidecarSet set default values for SidecarSet.
-func SetDefaultsSidecarSet(obj *v1alpha1.SidecarSet) {
+func SetDefaultsSidecarSet(obj *v1alpha1.SidecarSet, old *v1alpha1.SidecarSet) {
 	setSidecarSetUpdateStrategy(&obj.Spec.UpdateStrategy)
+
+	// preset container's pull policy according to old obj
+	if old != nil {
+		setPullPolicy := func(containers []v1alpha1.SidecarContainer, oldContainers []v1alpha1.SidecarContainer) {
+			pullPolicies := make(map[string]corev1.PullPolicy)
+			for _, c := range oldContainers {
+				pullPolicies[c.Name] = c.ImagePullPolicy
+			}
+			for i, c := range containers {
+				if c.ImagePullPolicy == "" {
+					obj.Spec.Containers[i].ImagePullPolicy = pullPolicies[c.Name]
+				}
+			}
+		}
+		setPullPolicy(obj.Spec.InitContainers, obj.Spec.InitContainers)
+		setPullPolicy(obj.Spec.Containers, old.Spec.Containers)
+	}
 
 	for i := range obj.Spec.InitContainers {
 		setSidecarDefaultContainer(&obj.Spec.InitContainers[i])
--- a/pkg/webhook/sidecarset/mutating/sidecarset_create_update_handler.go
+++ b/pkg/webhook/sidecarset/mutating/sidecarset_create_update_handler.go
@@ -77,7 +77,15 @@ func (h *SidecarSetCreateHandler) Handle(ctx context.Context, req admission.Requ
 	var copy runtime.Object = obj.DeepCopy()
 	switch req.AdmissionRequest.Operation {
 	case admissionv1.Create, admissionv1.Update:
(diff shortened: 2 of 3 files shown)


zmberg:
@ktalg Thank you very much, I have tried to understand this issue. I feel that the core of this problem is that modifying the Image also changes the ImagePullPolicy field, which prevents in-place upgrades from continuing. This is indeed quite strange, but we cannot fix it in this way, because if the user has really made changes to the ImagePullPolicy field, we cannot simply reuse the old value.

If we don't use the SetDefaults_Container function, it's not very reasonable. My opinion right now is to explicitly declare the ImagePullPolicy field to avoid this issue.

ktalg:
> @ktalg Thank you very much, I have tried to understand this issue. I feel that the core of this problem is that modifying the Image also changes the ImagePullPolicy field, which prevents in-place upgrades from continuing. This is indeed quite strange, but we cannot fix it in this way, because if the user has really made changes to the ImagePullPolicy field, we cannot simply reuse the old value.
> 
> If we don't use the SetDefaults_Container function, it's not very reasonable. My opinion right now is to explicitly declare the ImagePullPolicy field to avoid this issue.

This pr does cause some logical changes, but I don't think it causes the problem you are talking about, because it is a setting of the default value of ImagePullPolicy, which only takes effect when it is empty (just like the setting of the default value of other fields, which only works on empty values).Of course, this behavior does not take effect for newly created objects

If the user does not explicitly set the ImagePullPolicy when updating (not creating) the object, then we do not treat the "not set" action of this field as "set to empty" but rather we need to set This field needs to be set to a **reasonable** value(again, just like any other field),  I think it should be the old value and not the value determined by the image tag (which raises unexpected cases and is **not reasonable**)

_(If I'm missing something, please point it out)_

zmberg:
@ktalg I'm just thinking, it seems a bit strange to me. Additionally, we have changed the default value for ImagePullPolicy in k8s, and I'm not sure if that could bring up new issues.

ktalg:
> @ktalg I'm just thinking, it seems a bit strange to me. Additionally, we have changed the default value for ImagePullPolicy in k8s, and I'm not sure if that could bring up new issues.

After confirming that this change is work, let me clarify the possible undesired side effects it can cause and the scenarios where it happens:
1. the user changes the container of the sidecarSet after the sidecar has been injected: changes the image tag (from latest to a specific version) and sets the ImagePullPolicy to empty, when the value does not change with the image tag
2. After changing the sidecarSet in the above scenario and scale up the pod, the ImagePullPolicy in the new pod does not change with the image tag

The conditions for the above to happen are so harsh that it seems we can assume that the user would not expect ImagePullPolicy to be affected by the image tag - the above side effect should not have happened

zmberg:
@ktalg You have considered things very comprehensively, but my concerns are as follows: 
1. The imagePullPolicy actually affects the SidecarSet hash, and for existing SidecarSets, there may be unexpected changes. The SidecarSet hash is a very important field, not only for new additions but also for existing ones. I haven't fully thought through the implications for existing SidecarSets, and I'm concerned about the potential risks involved.
2. Is there really such a high frequency of scenarios where sidecar image switch from "latest" to a specific tag? Would the benefits and risks of doing this be proportional? If the likelihood of this happening is small, can users fix it by rebuilding Pods or changing the Pod annotations hash?

PR #1301, opened 2023-05-29
fix: SidecarSet Expectations Leakage Bug

wangwu50:
fix sidecarSet expectations leakage bug when delete pod alfter update sidecarset


wangwu50:
fixes #1293

pkg/controller/sidecarset/sidecarset_controller.go
```diff
@@ -115,9 +112,8 @@ var _ reconcile.Reconciler = &ReconcileSidecarSet{}
 // ReconcileSidecarSet reconciles a SidecarSet object
 type ReconcileSidecarSet struct {
 	client.Client
-	scheme             *runtime.Scheme
-	updateExpectations expectations.UpdateExpectations
```
zmberg, line 119: Why did you remove this filed? My understanding of the previous bug had nothing to do with this field being defined here, right?

wangwu50: Only updateExpectations were removed because they were initialized in utils

wangwu50: A utils is added so that the sidecarset handler can access it, similar to how cloneset handles it

zmberg: you can move the function to file https://github.com/openkruise/kruise/blob/master/pkg/control/sidecarcontrol/util.go.

wangwu50: OK, I've moved it

pkg/controller/sidecarset/sidecarset_pod_event_handler.go
```diff
@@ -82,6 +102,9 @@ func (p *enqueueRequestForPod) updatePod(q workqueue.RateLimitingInterface, old,
 		if sidecarSet.Spec.UpdateStrategy.Type == appsv1alpha1.NotUpdateSidecarSetStrategyType {
 			continue
 		}
+		if newPod.DeletionTimestamp != nil {
```
zmberg, line 105: I think you should not judge the deletionTimestamp, and use the ObserveUpdated function.

pkg/controller/sidecarset/sidecarset_pod_event_handler.go
```diff
@@ -57,6 +74,9 @@ func (p *enqueueRequestForPod) addPod(q workqueue.RateLimitingInterface, obj run
 	}
 
 	for _, sidecarSet := range sidecarSets {
+		if pod.DeletionTimestamp != nil {
```
zmberg, line 77: I think it is not necessary to execute the expectation function when add pod.

wangwu50: on a restart of the controller manager, it's possible a new pod shows up in a state that is already pending deletion. Prevent the pod from being a creation observation.

pkg/controller/sidecarset/sidecarset_pod_event_handler.go
wangwu50: ObserveUpdated does not determine if the pod is deleted if the pod is pending deletion

wangwu50: If this pod is an older version, there may be problems

furykerry (commented):
/lgtm

pkg/controller/sidecarset/sidecarset_pod_event_handler.go
```diff
@@ -82,6 +99,7 @@ func (p *enqueueRequestForPod) updatePod(q workqueue.RateLimitingInterface, old,
 		if sidecarSet.Spec.UpdateStrategy.Type == appsv1alpha1.NotUpdateSidecarSetStrategyType {
 			continue
 		}
+		sidecarcontrol.UpdateExpectations.ObserveUpdated(sidecarSet.Name, sidecarcontrol.GetSidecarSetRevision(sidecarSet), newPod)
```
veophi, line 102: A little suggestion: cloud you move this Line to Line 99?

wangwu50: If I put it here, I can filter out Pods managed by sidecar of type NotUpdate, right?

PR #1303, opened 2023-06-02
Sidecar terminator ignore the exit code of the sidecar container

### LINKED ISSUE #1285: [feature request]  Sidecar terminator ignore the exit code of the sidecar container
OpenKruise job sidecar terminator controller watches and terminates sidecar containers for such job-type Pods when its main containers completed, details: [JobSidecarTerminator](https://openkruise.io/docs/user-manuals/jobsidecarterminator).

But, the above feature contains certain limitation of "Your sidecar container must respond the SIGTERM signal, and the entrypoint should exit 0 when received this signal." If the exit code of the sidecar container is non-zero, Job will judge the pod failure.

So i hope JobSidecarTerminator can solve the above problem and ensure that the job is completed normally even if the exit code of the sidecar container is non-zero.

diannaowa:

Pods are not allowed to transition out of terminal phases when kubelet report the pod status.([ref](https://github.com/kubernetes/kubernetes/blob/c07c739f04880a41b624009f4a2c5459d5cbac6c/pkg/kubelet/kubelet_pods.go#L1586))
Base on this context, the proposal as bellow：
In the sidecar terminator controller:
1，Calculate the Pod phase from the exit-code of the main container:
   the pod phase will be Succeeded,  if all main containers are succeeded .
   the pod phase will be failed ,if at least one main container is failed(the exitCode is not zero)  .
2,  Terminate it if the job pod is complete.

3, Terminate the pod and set the condition of the pod as the bellow:

```yaml
- lastProbeTime: null
    lastTransitionTime: "2023-06-06T14:06:08Z"
    status: "True"
    type: SidecarTerminated
```

4, If the job pod is completed, we will skip terminating the pod. Check whether a job is complete by:

```go
func isJobPodCompleted(pod *corev1.Pod) bool {
	mainContainers := getMain(pod)
	if !containersCompleted(pod, mainContainers) {
		return false
	}
	if containersSucceeded(pod, mainContainers) {
		return pod.Status.Phase == corev1.PodSucceeded
	} else {
		return pod.Status.Phase == corev1.PodFailed
	}
}
```

Native controller actions:

1 .Job controller update the status of the job when pod phase is updated. 
2. As we said above, kubelet only update the container status and do not update the pod phase since it is terminal phase. And the pod will reach a final state.
3. Kubelet will kill the containers that are not in terminal phase, after the pod the terminated by sidecar terminator.

fixes #1285

### DIFF
--- a/pkg/controller/sidecarterminator/sidecar_terminator_controller.go
+++ b/pkg/controller/sidecarterminator/sidecar_terminator_controller.go
@@ -18,19 +18,17 @@ package sidecarterminator
 
 import (
 	"context"
+	"encoding/json"
 	"flag"
+	"fmt"
 	"strings"
 	"time"
 
-	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
-	"github.com/openkruise/kruise/pkg/features"
-	"github.com/openkruise/kruise/pkg/util"
-	utilclient "github.com/openkruise/kruise/pkg/util/client"
-	utilfeature "github.com/openkruise/kruise/pkg/util/feature"
-	"github.com/openkruise/kruise/pkg/util/ratelimiter"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
@@ -39,6 +37,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
+	"github.com/openkruise/kruise/pkg/features"
+	"github.com/openkruise/kruise/pkg/util"
+	utilclient "github.com/openkruise/kruise/pkg/util/client"
+	utilfeature "github.com/openkruise/kruise/pkg/util/feature"
+	"github.com/openkruise/kruise/pkg/util/ratelimiter"
 )
 
 func init() {
@@ -46,7 +51,8 @@ func init() {
 }
 
 var (
-	concurrentReconciles = 3
+	concurrentReconciles                         = 3
+	SidecarTerminated    corev1.PodConditionType = "SidecarTerminated"
 )
 
 /**
@@ -131,17 +137,8 @@ func (r *ReconcileSidecarTerminator) doReconcile(pod *corev1.Pod) (reconcile.Res
 		return reconcile.Result{}, nil
 	}
 
-	if containersCompleted(pod, getSidecar(pod)) {
-		klog.V(3).Infof("SidecarTerminator -- all sidecars of pod(%v/%v) have been completed, no need to process", pod.Namespace, pod.Name)
-		return reconcile.Result{}, nil
-	}
-
-	if pod.Spec.RestartPolicy == corev1.RestartPolicyOnFailure && !containersSucceeded(pod, getMain(pod)) {
-		klog.V(3).Infof("SidecarTerminator -- pod(%v/%v) is trying to restart, no need to process", pod.Namespace, pod.Name)
-		return reconcile.Result{}, nil
-	}
-
 	sidecarNeedToExecuteKillContainer, sidecarNeedToExecuteInPlaceUpdate, err := r.groupSidecars(pod)
+
 	if err != nil {
 		return reconcile.Result{}, err
 	}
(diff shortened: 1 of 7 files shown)


pkg/controller/sidecarterminator/sidecar_terminator_controller.go
```diff
@@ -152,23 +162,137 @@ func (r *ReconcileSidecarTerminator) doReconcile(pod *corev1.Pod) (reconcile.Res
...
+			}
+			newSidecarStatus[status.Name] = newStatus
+		}
+
+	}
+
+	var err error
+	if changed {
+
+		err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
furykerry, line 222: it is not necessary to retryonconflict in controller, if the reconciliation fails, the entry will be re-enqueued. retryonconflict will introduce sleep time and will waste the controller worker time.

pkg/controller/sidecarterminator/sidecar_terminator_controller.go
```diff
@@ -152,23 +162,137 @@ func (r *ReconcileSidecarTerminator) doReconcile(pod *corev1.Pod) (reconcile.Res
...
+		}
+
+		changed = true
+
+		// The kubelet may update the state of the containers and the phase of the pod before the sidecar terminator controller
+		if status.State.Terminated != nil && status.State.Terminated.ExitCode != int32(0) {
+			klog.V(3).Infof("SidecarTerminator -- ignore the non-zero exit code of the sidecar container %s/%s, the pod phase will be updated", pod.Name, status.Name)
+			newSidecarStatus[status.Name] = *status
+		} else if status.State.Terminated == nil && status.State.Running != nil {
+			klog.V(3).Infof("SidecarTerminator -- sync the status of the sidecar container %s/%s and update the pod phase, all of the main containers are completed", pod.Name, status.Name)
```
furykerry, line 201: before adding a Terminated  state, i think sidecar terminator should wait for the sidecar container to exit with non zero code

diannaowa: In the scenario where kruise-daemon is down, this means that we cannot complete the job until  kruise-daemon is successfully started.

diannaowa: maybe we should not care the status of the sidecar containers.  And update the pod phase if no waiting sidecar container.

diannaowa: I think you are correct , it will avoid kubelet and sidecar terminator from updating container status unnecessarily.

veophi:
Can we set `Pod.Status.Phase` to `Succeeded/Failed` **BEFORE** kill sidecars?

diannaowa:
will do

pkg/controller/sidecarterminator/sidecar_terminator_controller.go
```diff
@@ -154,23 +162,84 @@ func (r *ReconcileSidecarTerminator) doReconcile(pod *corev1.Pod) (reconcile.Res
...
+	}
+	if isJobPodCompleted(pod) {
+		klog.V(3).Infof("the job pod is completed -- pod (%s/%s) ,no need to process", pod.Namespace, pod.Name)
+		return nil
+	}
+
+	// after the pod is terminated by the sidecar terminator, kubelet will kill the containers that are not in the terminal phase
+	// 1. sidecar container terminate with non-zero exit code
+	// 2. sidecar container is not in a terminal phase (still running or waiting)
+	if !containersSucceeded(pod, sidecars) {
```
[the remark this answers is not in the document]
diannaowa: done

pkg/controller/sidecarterminator/sidecar_terminator_controller.go
```diff
@@ -101,6 +109,7 @@ type ReconcileSidecarTerminator struct {
 	client.Client
 	recorder record.EventRecorder
 	scheme   *runtime.Scheme
+	clock    clock.Clock
```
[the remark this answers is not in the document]
diannaowa: done

pkg/controller/sidecarterminator/sidecar_terminator_controller.go
```diff
@@ -150,13 +150,62 @@ func (r *ReconcileSidecarTerminator) doReconcile(pod *corev1.Pod) (reconcile.Res
...
 	if err := r.executeKillContainerAction(pod, sidecarNeedToExecuteKillContainer); err != nil {
 		return reconcile.Result{}, err
 	}
 
 	return reconcile.Result{}, nil
 }
 
+// terminateJobPod terminate the job pod and skip the state of the sidecar containers
+// This method should only be called before the executeKillContainerAction
+func (r *ReconcileSidecarTerminator) terminateJobPod(pod *corev1.Pod) error {
```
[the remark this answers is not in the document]
diannaowa: Good idea.

pkg/webhook/containerrecreaterequest/mutating/crr_mutating_handler.go
```diff
@@ -123,7 +125,7 @@ func (h *ContainerRecreateRequestHandler) Handle(ctx context.Context, req admiss
 		}
 		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed to find Pod %s: %v", obj.Spec.PodName, err))
 	}
-	if !kubecontroller.IsPodActive(pod) {
+	if !kubecontroller.IsPodActive(pod) && !isTerminatedBySidecarTerminator(pod) {
```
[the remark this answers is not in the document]
diannaowa: it is necessary, we should create CRR to kill sidecar container after change the pod phase to terminal phase.    but can not create CRR for  inactive pod in normal case.

furykerry (commented):
/lgtm

PR #1314, opened 2023-06-14
add status.labelSelector field to unitedDeployment to support scale sub-resource

### LINKED ISSUE #1308: [feature request] UnitedDeployment supports scale sub-resource
add `status.selector` field to unitedDeployment to support scale sub-resource

**Why is this needed**:
support HPA

diannaowa:
Add status.labelSelector field to unitedDeployment to support scale sub-resource

fixes #1308

### DIFF
--- a/pkg/controller/uniteddeployment/uniteddeployment_controller.go
+++ b/pkg/controller/uniteddeployment/uniteddeployment_controller.go
@@ -22,12 +22,6 @@ import (
 	"fmt"
 	"reflect"
 
-	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
-	appsv1beta1 "github.com/openkruise/kruise/apis/apps/v1beta1"
-	"github.com/openkruise/kruise/pkg/controller/uniteddeployment/adapter"
-	utilclient "github.com/openkruise/kruise/pkg/util/client"
-	utildiscovery "github.com/openkruise/kruise/pkg/util/discovery"
-	"github.com/openkruise/kruise/pkg/util/ratelimiter"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


diannaowa:
This is a romantic PR id

pkg/controller/uniteddeployment/uniteddeployment_controller.go
```diff
@@ -231,6 +233,14 @@ func (r *ReconcileUnitedDeployment) Reconcile(_ context.Context, request reconci
 		return reconcile.Result{}, err
 	}
 
+	selector, err := util.ValidatedLabelSelectorAsSelector(instance.Spec.Selector)
```
furykerry, line 236: just using ValidatedLabelSelectorAsSelector is not enough to identify some label selector errors, you can call LabelSelectorAsSelector and check if there is any error. Similar code is in the cloneset validation validateCloneSetSpec

diannaowa: Do you mean validate UnitedDeploymentSpec? https://github.com/openkruise/kruise/blob/9e74506ca800877f5bf63723efa33521a3533c46/pkg/webhook/uniteddeployment/validating/uniteddeployment_validation.go#LL52C60-L52C60

furykerry: yes, it seems the uniteddeployment webhook already validate the selector, is it necessary to validate again the controller

diannaowa: LabelSelectorAsSelector has been called in the validateUnitedDeploymentSpec method to identify some label selector errors.
I think ValidatedLabelSelectorAsSelector is ok here.

diannaowa: > yes, it seems the uniteddeployment webhook already validate the selector, is it necessary to validate again the controller

Unnecessary, i think.

furykerry (commented):
/lgtm

PR #1334, opened 2023-07-17
optimize webhook patchResponse function

zmberg:
There is a risk that the current webhook patch method will cause some fields to be lost if the rollout references a low version of the api. For example, k8s 1.26 adds a new field topologySpreadConstraint.minDomains, the field will be lost after the webhook.


furykerry (commented):
/lgtm

PR #1336, opened 2023-07-18
docs: add in-place workload vertical scaling proposal

### LINKED ISSUE #1212: [feature request] In-place udpate support resources
The [In-place Pod Vertical Scaling feature](https://github.com/kubernetes/kubernetes/pull/102884) have been merged in k8s upstream, so openKruise in-place update need support resources, related workloads:
- CloneSet
- Advanced StatefulSet
- Advanced DaemonSet
- SidecarSet

LavenderQAQ:
### Describe what this PR does
Add a proposal for In-place Workload Vertical Scaling.

Fixes #1212


LavenderQAQ:
For the case of vscale failure, my current idea is to start the rollback mechanism, is it necessary to provide users with some configurable interfaces? Because for vscale, rolling back is more like starting another vscale.

LavenderQAQ:
Do I need to add more details to `expectationDiffs`? For example, which container needs vertical scaling. But this can make the structure too bloated.

docs/proposals/20230718-inplace-workload-vertical-scaling.md
```diff
@@ -0,0 +1,813 @@
...
+// CloneSetVScaleStrategy defines strategies for pods vscale.
+type CloneSetVScaleStrategy struct {
+	MaxProposed *intstr.IntOrString `json:"maxProposed,omitempty"`
+	Paused   bool  `json:"paused,omitempty"`
+}
+```
+
+##### Enhance Expectation
+
+Create a new VScaleExpectations for use by different controllers.
```
veophi, line 759: why not reuse `ResourceVersionExpectations` ?

docs/proposals/20230718-inplace-workload-vertical-scaling.md
```diff
@@ -0,0 +1,813 @@
...
+
+##### Enhance calculateDiffsWithExpectation
+
+Vscale-related parameters need to be added to expectationDiffs
+
+```go
+type expectationDiffs struct {
+	...
+	// vscale identifies whether resources have changed
+	vscale bool
```
veophi, line 780:
```go
numVScale int32
```

docs/proposals/20230718-inplace-workload-vertical-scaling.md
```diff
@@ -0,0 +1,813 @@
...
+
+### User Stories
+
+Users can start vertical scaling of all managed Pods by modifying the resource field in workload.
+
+### Implementation Details/Notes/Constraints
+
+#### Design Objectives
+
+- Use vscale as a standalone operation rather than integrated in update. For example, for cloneset, three operations will be improved to scale, vscale, and update.
```
veophi, line 740: Good idea! But we should consider how to avoid to break `maxUnavailable`  limitation in some scene, for example, users update both resources and images fields, and settings `RestartRequired` of vscale.

docs/proposals/20230718-inplace-workload-vertical-scaling.md
```diff
@@ -0,0 +1,813 @@
...
+At the same time need to modify the `calculateDiffsWithExpectation` function, when the vscale closed scale behavior and the behavior of the update.
+
+##### Flow Chart
+
+The general workload processing logic is shown in the following figure.
+![workload-resize-flowchart](../img/workload-resize-flowchart.png)
+
+### Risks and Mitigations
+
+- The main risk is the impact on the cluster when a large number of Pods adjust resources, which is entirely dependent on the user setting the maxProposed value correctly
```
veophi, line 793: What will happen if a large number of Pods adjust resources?

docs/proposals/20230718-inplace-workload-vertical-scaling.md
```diff
@@ -0,0 +1,813 @@
...
+ }
+```
+
+The corresponding changes to validation are not expanded here.
+
+### Goals
+
+- Users only need to modify template in workload to achieve vertical scaling for all managed Pods
+- Simple and clear configuration strategy, support grayscale capabilities
+- Provides the rollback function, and attempts to return to the previous state when the capacity expansion and shrinkage fails. If the rollback still fails, no further action is required
```
veophi, line 724: Maybe, we should provide more policies. For example. if a vscale proposal is rejected by kubelet (Infeasible), we should allow it to be demote to rebuild upgrade.

docs/proposals/20230718-inplace-workload-vertical-scaling.md
```diff
@@ -0,0 +1,813 @@
...
+
+#### CloneSet
+
+##### Extend Definition
+
+We want users to write in PodSpec whether or not each container needs to be restarted.
+
+```go
+// CloneSetVScaleStrategy defines strategies for pods vscale.
+type CloneSetVScaleStrategy struct {
```
veophi, line 751: Maybe, it's better to reuse `UpdateStrategy` filed for the following considers:
1. vscale also should consider `maxUnavailable`;
2. vscale also need grayscale capabilities (`partition`) ；

docs/proposals/20230718-inplace-workload-vertical-scaling.md
LavenderQAQ: I think there might be some judgment here about the vscale status, many Pods might be in the Proposed or InProgress state, and there seems to be no way to judge only resource_version.

docs/proposals/20230718-inplace-workload-vertical-scaling.md
LavenderQAQ: Yes, I should change the quantity here. This design is not perfect for the time being, and we need to consider the vertical scaling of some Pods (not all) in the future. Then I'll update it with a commit.

docs/proposals/20230718-inplace-workload-vertical-scaling.md
Hanmengnan: As a user, I may be concerned that if I want to make resource adjustments to the workload, but this operation will lead to pod reconstruction, the cost of this operation may be too high.

Hanmengnan: I may be very concerned about how to achieve dynamic scaling with minimal overhead.

LavenderQAQ: @Hanmengnan After thorough analysis, I believe that it is a better choice not to separate vertical scaling as a separate operation, but to treat it as a special upgrade operation like in-place upgrades. This can solve many issues. Regarding the risk, I think as long as the vertical scaling feature of Kubernetes Pods is stable enough, there will be no uncontrollable risks in implementing it at the workload level.

LavenderQAQ:
sorry, there may be some significant design updates, and after trying to implement a demo, I believe that treating workload vertical scaling as part of the update operation may be a better choice. We can discuss this new design in the community meeting. I have a wonderful demo here.

LavenderQAQ:
I think the current version has been upgraded to 1.28, so consider reopening this pr?

PR #1343, opened 2023-07-25
enhancement: ambiguous function name of isSidecarSetNotUpdate

### LINKED ISSUE #1342: [ENHANCEMENT] "isSidecarSetNotUpdate" ambiguous function name
The function name and its implementation have ambiguity. The function `isSidecarSetNotUpdate` should return `true` for `NotUpdate` SidecarSet and `false` otherwise. However, the current implementation is as follows:
```go
func isSidecarSetNotUpdate(s *appsv1alpha1.SidecarSet) bool {
	if s.Spec.UpdateStrategy.Type == appsv1alpha1.NotUpdateSidecarSetStrategyType {
		klog.V(3).Infof("sidecarSet spreading RollingUpdate config type, name: %s, type: %s", s.Name, s.Spec.UpdateStrategy.Type)
		return false
	}
	return true
}
```
The return results are opposite to the function name, which can make the code reading challenging.
This function is called in `sidecarset_processor.go` in a strange way:
```go
// 4. SidecarSet upgrade strategy type is NotUpdate
if !isSidecarSetNotUpdate(sidecarSet) {
}
```
I believe the ideal way to call it should be:
```go
if isSidecarSetNotUpdate(sidecarSet) {
}
```
Therefore, the implementation of the function should be corrected as follows:
```go
func isSidecarSetNotUpdate(s *appsv1alpha1.SidecarSet) bool {
	if s.Spec.UpdateStrategy.Type == appsv1alpha1.NotUpdateSidecarSetStrategyType {
		klog.V(3).Infof("sidecarSet spreading RollingUpdate config type, name: %s, type: %s", s.Name, s.Spec.UpdateStrategy.Type)
		return true
	}
	return false
}
```

Colvin-Y:
fix #1342 
This modification only fixes the ambiguity between the implementation and the name of the function isSidecarSetNotUpdate, without altering the original logic.

fix #1342

### DIFF
--- a/pkg/controller/sidecarset/sidecarset_processor.go
+++ b/pkg/controller/sidecarset/sidecarset_processor.go
@@ -131,7 +131,7 @@ func (p *Processor) UpdateSidecarSet(sidecarSet *appsv1alpha1.SidecarSet) (recon
 	}
 
 	// 4. SidecarSet upgrade strategy type is NotUpdate
-	if !isSidecarSetNotUpdate(sidecarSet) {
+	if isSidecarSetNotUpdate(sidecarSet) {
 		return reconcile.Result{}, nil
 	}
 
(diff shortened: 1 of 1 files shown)


nashtsai:
this PR is a bug fixed, I wonder why there is no associated tests?

PR #1349, opened 2023-07-28
SidecarSet add upgrade state in pod annotation

### LINKED ISSUE #1312: [feature request] SidecarSet add upgrade state in pod annotation
SidecarSet can publish Sidecar containers in batches, and currently needs to add a representation to make it easier for upper-level paas to support more clearly which Pods are in the process of being published and which are finished. 

Currently Pod have sidecarset annotations[kruise.io/sidecarset-inplace-update-state] and value is struct:
```
type SidecarSetUpgradeSpec struct {
	UpdateTimestamp              metav1.Time `json:"updateTimestamp"`
	SidecarSetHash               string      `json:"hash"`
	SidecarSetName               string      `json:"sidecarSetName"`
	SidecarList                  []string    `json:"sidecarList"`                  // sidecarSet container list
	SidecarSetControllerRevision string      `json:"controllerRevision,omitempty"` // sidecarSet controllerRevision name
}
```

So we can add field in the above struct:
```
type SidecarSetUpgradeSpec struct {
	...
         // enum: Normal, Updating
        State string `json:"state"`
}
```
```

### PR DESCRIPTION
**Pull request by xiangpingjiang**

fixes #1312

### DIFF
--- a/pkg/control/sidecarcontrol/util.go
+++ b/pkg/control/sidecarcontrol/util.go
@@ -64,6 +64,10 @@ const (
 
 	// SidecarSetUpgradable is a pod condition to indicate whether the pod's sidecarset is upgradable
 	SidecarSetUpgradable corev1.PodConditionType = "SidecarSetUpgradable"
+
+	// SidecarSetUpgradeSpec State
+	SidecarSetHashStateUpdating = "Updating"
+	SidecarSetHashStateNormal   = "Normal"
 )
 
 var (
@@ -81,6 +85,7 @@ type SidecarSetUpgradeSpec struct {
 	SidecarSetName               string      `json:"sidecarSetName"`
 	SidecarList                  []string    `json:"sidecarList"`                  // sidecarSet container list
 	SidecarSetControllerRevision string      `json:"controllerRevision,omitempty"` // sidecarSet controllerRevision name
+	State                        string      `json:"state"`                        // enum: Normal, Updating
 }
 
 // PodMatchSidecarSet determines if pod match Selector of sidecar.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pkg/control/sidecarcontrol/util.go [hunk 1]
```diff
@@ -253,11 +254,81 @@ func UpdatePodSidecarSetHash(pod *corev1.Pod, sidecarSet *appsv1alpha1.SidecarSe
 		SidecarSetName:               sidecarSet.Name,
 		SidecarList:                  sidecarList.List(),
 		SidecarSetControllerRevision: sidecarSet.Status.LatestRevision,
+		State:                        "Updating",
```
hantmac, line 257: Maybe should define the `State` using the `Enum` type or make it a const.

xiangpingjiang: done

pkg/controller/sidecarset/sidecarset_processor.go
```diff
@@ -103,6 +103,13 @@ func (p *Processor) UpdateSidecarSet(sidecarSet *appsv1alpha1.SidecarSet) (recon
 		return reconcile.Result{RequeueAfter: time.Second}, nil
 	}
 
+	// check if pod finish sidecar container upgrade
+	for _, pod := range pods {
```
hantmac, line 107: Maybe we should make `updatePodsSidecarHashState` after [line-99](https://github.com/openkruise/kruise/pull/1349/files#diff-0044cb47203edfb25c4c6e701614e275692fd84828140d452b3692b1106ae5bbR99) because some pods that already ready should be update state to `Noamal`.

pkg/controller/sidecarset/sidecarset_processor.go
```diff
@@ -189,6 +196,53 @@ func (p *Processor) updatePods(control sidecarcontrol.SidecarControl, pods []*co
 	return nil
 }
 
+func (p *Processor) updatePodSidecarHashState(control sidecarcontrol.SidecarControl, pod *corev1.Pod) error {
```
hantmac, line 199: Need some e2e tests?

Issue #1384, opened 2023-08-24
[feature request] Add force image pull


ls-2018:
```
When predownloading using imagepulljob, it appears that the latest image will not be pulled. 
(Same for tag, but different for mirror)

An image already exists. For example, redis:7
```

ls-2018:
I added a Boolean field to implement the relevant logic. Changes to existing logic are minimal

PR #1385, opened 2023-08-24
[Feature] Add force image pull

ls-2018:
- https://github.com/openkruise/kruise/issues/1384


zmberg:
I would suggest that an additional judgment is needed, e.g., local digest is not equal to the value inside the repository, then pull.

chengjoey:
we typically avoid boolean fields due to the guidance [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) to "Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options."

ls-2018:
@chengjoey Thank you for your valuable advice.
@zmberg I will modify it according to your suggestion.

pkg/daemon/imagepuller/imagepuller_worker.go
```diff
@@ -428,16 +430,24 @@ func (w *pullWorker) getImageInfo(ctx context.Context) (*runtimeimage.ImageInfo,
...
 	tag := w.tagSpec.Tag
 	startTime := metav1.Now()
 
 	klog.Infof("Worker is starting to pull image %s:%s version %v", w.name, tag, w.tagSpec.Version)
 
-	if _, e := w.getImageInfo(ctx); e == nil {
-		klog.Infof("Image %s:%s is already exists", w.name, tag)
-		newStatus.Progress = 100
-		return nil
+	if info, e := w.getImageInfo(ctx); e == nil {
```
zmberg, line 439:
```
if imagePullPolicy == appsv1alpha1.PullAlways &&  {
    if !w.shouldPull(info.ID, w.name, tag, w.secrets) {

    }
} else if info, e := w.getImageInfo(ctx); e == nil { 
   //
}
```

pkg/daemon/criruntime/imageruntime/interface.go
```diff
@@ -50,4 +50,5 @@ type ImagePullStatusReader interface {
 type ImageService interface {
 	PullImage(ctx context.Context, imageName, tag string, pullSecrets []v1.Secret, sandboxConfig *appsv1alpha1.SandboxConfig) (ImagePullStatusReader, error)
 	ListImages(ctx context.Context) ([]ImageInfo, error)
+	AuthInfos(ctx context.Context, imageName, tag string, pullSecrets []v1.Secret) []daemonutil.AuthInfo
```
zmberg, line 53: This method does not need to interact with the container runtime, so it can be defined directly into the utils to implement a public method.

pkg/daemon/imagepuller/utils.go
```diff
@@ -19,13 +19,19 @@ package imagepuller
 import (
 	"context"
 	"fmt"
+	"log"
 	"math/rand"
 	"time"
 
+	"github.com/google/go-containerregistry/pkg/authn"
```
zmberg, line 26: go import sort

ls-2018:
@zmberg Could you look at it again, please?

ls-2018:
Is there any progress in the evaluation?

iclouding:
Is there any progress on this pull request?

ls-2018:
@zmberg @veophi @furykerry Can someone evaluate it, please?

ls-2018:
@zmberg @veophi @furykerry  Can you look at this request again please? Thanks.

Issue #1386, opened 2023-08-25
[BUG] alter imagepulljob inoperative


ls-2018:
After the imagepodjob is completed, the job content is changed and the job is not working
 

## What happened
```
apiVersion: apps.kruise.io/v1alpha1
kind: ImagePullJob
metadata:
  name: job-with-always2
spec:
  image: nginx:1.9.1
  parallelism: 10
  selector:
    names:
      - kind-worker3
```

When the job is done and the selector is removed and reapplied, the job does not work

## What you expected to happen

Pull the image to the desired node

PR #1387, opened 2023-08-25
[Fix] alter imagepulljob inoperative

ls-2018:
- https://github.com/openkruise/kruise/issues/1386

### DIFF
--- a/pkg/controller/imagepulljob/imagepulljob_controller.go
+++ b/pkg/controller/imagepulljob/imagepulljob_controller.go
@@ -20,8 +20,13 @@ import (
 	"context"
 	"flag"
 	"fmt"
+
 	"sort"
 	"time"
+
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
 	daemonutil "github.com/openkruise/kruise/pkg/daemon/util"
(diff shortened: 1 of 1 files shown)


zmberg:
Would setting the completionPolicy to never be sufficient for your needs? The ability to do this needs to be discussed, and it shouldn't be possible to implement it this way.

ls-2018:
👌🏻, look forward to the discussion

zmberg:
@ls-2018 Can I talk to you in the dingding group?

ls-2018:
Of course, my name is Liu Shuo

Like k8s native job, it is not allowed to change after completion.
And our crd can be changed, which gives me the feeling that it can be changed and rerun.
Should we do the same or improve on it？

pkg/controller/imagepulljob/imagepulljob_controller.go
```diff
@@ -92,7 +97,19 @@ func add(mgr manager.Manager, r *ReconcileImagePullJob) error {
...
 	// Watch for changes to ImagePullJob
-	err = c.Watch(&source.Kind{Type: &appsv1alpha1.ImagePullJob{}}, &handler.EnqueueRequestForObject{})
+	err = c.Watch(&source.Kind{Type: &appsv1alpha1.ImagePullJob{}}, &handler.EnqueueRequestForObject{}, predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			oldIPJ := e.ObjectOld.(*appsv1alpha1.ImagePullJob)
+			newIPJ := e.ObjectNew.(*appsv1alpha1.ImagePullJob)
+
+			if !apiequality.Semantic.DeepEqual(oldIPJ.Spec, newIPJ.Spec) && oldIPJ.Status.CompletionTime != nil {
+				newIPJ.Status = appsv1alpha1.ImagePullJobStatus{}
+				r.Status().Update(context.TODO(), newIPJ)
```
veophi, line 107: Do not write object in event handler...

PR #1399, opened 2023-09-01
[Fix] Remove history field

ls-2018:


liangyuanpeng (approved):

zmberg:
@ls-2018 We need to confirm again whether this can be removed.

PR #1401, opened 2023-09-05
Fix EphemeralJob event handler for deleting object

FillZpp:
Fix EphemeralJob event handler for deleting object


furykerry (commented):
/lgtm

PR #1404, opened 2023-09-05
chore: remove refs to deprecated io/ioutil

testwill:


zmberg:
@testwill You should exec 'File is not `gofmt`-ed with `-s` (gofmt)' in file checker.go

PR #1407, opened 2023-09-08
use "leases" by default

dsxing:
LeaderElectionResourceLock  use "leases" by default

NONE

NONE

https://github.com/kubernetes/kubernetes/issues/80289


zmberg:
@dsxing If default use leases, then upgrade kruise from 1.3.0 to 1.5.0 directly, there will have two leader a period of time.

dsxing:
> @dsxing If default use leases, then upgrade kruise from 1.3.0 to 1.5.0 directly, there will have two leader a period of time.

I see it as well, Upgrading from configmaps to leases requires an intermediate step, namely configmapleases. Version 1.3 uses configmaps, while versions 1.4 and 1.5 make use of configmapleases. I am wondering if versions beyond 1.5 can directly use leases. Versions 1.4 and 1.5 serve as transitional stages in this upgrade process.

zmberg:
@dsxing Can you solve the issue, https://github.com/openkruise/kruise/pull/1407/checks?check_run_id=16603745643

dsxing:
> @dsxing Can you solve the issue, https://github.com/openkruise/kruise/pull/1407/checks?check_run_id=16603745643

Alright, it's done.

Issue #1409, opened 2023-09-09
[BUG] BroadcastJob  activeDeadlineSeconds did not take effect


ls-2018:
**What happened**:
```
pod is over 300 seconds, but still there
 ```
**What you expected to happen**:
```
pod cleared
 ```

**How to reproduce it (as minimally and precisely as possible)**:
```
echo 'kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
featureGates:
  "EphemeralContainers": true
nodes:
- role: control-plane
  image: kindest/node:v1.26.0
  extraPortMappings:
  - containerPort: 6443
    hostPort: 6443
    protocol: TCP
- role: worker
  image: kindest/node:v1.26.0
  labels:
    zone: c
' >/tmp/kind.yaml

kind create cluster --config /tmp/kind.yaml
kubectl cluster-info --context kind-kind

helm install kruise openkruise/kruise --version 1.4.0


echo 'apiVersion: apps.kruise.io/v1alpha1
kind: BroadcastJob
metadata:
  name: broadcastjob-ttl
spec:
  template:
    spec:
      containers:
        - name: pi
          image: registry.cn-hangzhou.aliyuncs.com/acejilam/tensorflow:latest-gpu     # 5.7G
      restartPolicy: OnFailure
  completionPolicy:
    type: Always
    activeDeadlineSeconds: 300
    ttlSecondsAfterFinished: 300
  paused: false
  parallelism: 3
' >/tmp/bcj.yaml
kubectl apply -f /tmp/bcj.yaml

```
**Anything else we need to know?**:
```
But it doesn't happen every time, and I've found that when the watch pod, node, or BroadcastJob have no events, the tuning logic doesn't fire
 
```
**Environment**:
- Kruise version:
- Kubernetes version (use `kubectl version`):
- Install details (e.g. helm install args):
- Others:

<img width="556" alt="image" src="https://github.com/openkruise/kruise/assets/38821215/fcbd5228-fd21-4455-8dab-075464f1e314">
<img width="674" alt="image" src="https://github.com/openkruise/kruise/assets/38821215/b81d3e45-9842-4e2c-98ca-e6b521f7378f">
<img width="1502" alt="image" src="https://github.com/openkruise/kruise/assets/38821215/480ea917-7ed9-4404-aa16-7c403ef119a5">

I changed the activeDeadlineSeconds to 30 seconds, and I did it again.
  
[...]

PR #1410, opened 2023-09-09
[FIX] BroadcastJob activeDeadlineSeconds did not take effect

ls-2018:
- https://github.com/openkruise/kruise/issues/1409

A description of the specific problem is in issue

### DIFF
--- a/pkg/controller/broadcastjob/broadcastjob_controller.go
+++ b/pkg/controller/broadcastjob/broadcastjob_controller.go
@@ -245,7 +245,15 @@ func (r *ReconcileBroadcastJob) Reconcile(_ context.Context, request reconcile.R
 	failed := int32(len(failedPods))
 	succeeded := int32(len(succeededPods))
 
-	desiredNodes, restNodesToRunPod, podsToDelete := getNodesToRunPod(nodes, job, existingNodeToPodMap)
+	isDeadLine := func(job *appsv1alpha1.BroadcastJob) bool {
+		if job.Spec.CompletionPolicy.Type == appsv1alpha1.Always &&
+			job.Spec.CompletionPolicy.ActiveDeadlineSeconds != nil {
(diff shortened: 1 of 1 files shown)


pkg/util/handler/enqueue.go
```diff
@@ -0,0 +1,99 @@
...
+
+	appsalphav1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type empty struct{}
```
zmberg, line 30: Why must this official function be redefined

ls-2018:
I simplified the code a little bit.


Reconcile all of the Reconcile function can be handled because AdvancedCronJob, BroadcastJob and Job events occur. However, in the case I submitted, I watched three kinds of resources respectively and found that no incident happened. Of course, this probability may be low, but it is not impossible

zmberg:
What I'm more interested in understanding is how this patch related to the issue? Bug?

ls-2018:
I think it was a bug. 
Pulling the large image exceeded the expected death time, but the pod could not be killed because no event was watched.

zmberg:
@ls-2018 You are right, and i think you can refer the implement, https://github.com/openkruise/rollouts/blob/master/pkg/controller/rollout/rollout_controller.go#L167

ls-2018:
I think I have found the root cause. It is not that the ActiveDeadlineSeconds have not been reconciled after expiration, but that there is no judgment on whether the pod should be removed or not.
@zmberg

ls-2018:
@zmberg  Can you look at this request again please? Thanks.

PR #1412, opened 2023-09-11
Change Kruise leader election from configmapsleases to leases

yanfeng1992:

Change Kruise leader election from configmapsleases to leases


yanfeng1992:
Duplicate with https://github.com/openkruise/kruise/pull/1407

PR #1413, opened 2023-09-12
[Fix] Index overshoot

ls-2018:
When I try to manually specify a key in configmap, the controller panics
 
image

```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
    - name: main
      image: centos:7
      command:
        - "/bin/sh"
        - "-c"
        - "sleep 1d"
    - name: sidecar
      image: centos:7
      command:
        - "/bin/sh"
        - "-c"
        - "sleep 1d"
      env:
        - name: KRUISE_CONTAINER_PRIORITY
          value: "1"
        - name: KRUISE_CONTAINER_BARRIER
          valueFrom:
            configMapKeyRef:
              name: x
              key: "p"
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: x
data:
  p: "1"
```

### DIFF
--- a/pkg/util/containerlaunchpriority/container_launch_prirotiy.go
+++ b/pkg/util/containerlaunchpriority/container_launch_prirotiy.go
@@ -44,7 +44,7 @@ func ExistsPriorities(pod *v1.Pod) bool {
 
 func GetContainerPriority(c *v1.Container) *int {
 	for _, e := range c.Env {
-		if e.Name == appspub.ContainerLaunchBarrierEnvName {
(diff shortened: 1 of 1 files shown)


zmberg:
@ls-2018 Shouldn't the user be allowed to configure KRUISE_CONTAINER_BARRIER themselves, and if they do, it should be overridden by the kruise webhook?
image

ls-2018:
@zmberg  I don't think we need to expose this to the user. The user just needs to operate KRUISE_CONTAINER_PRIORITY.

If the user sets KRUISE_CONTAINER_BARRIER, I personally feel that it is better to override the env, after all, this is a very special.

PR #1430, opened 2023-09-28
Reduce the number of updateNodePodProbes calls

ls-2018:
Currently, the update npp is called once per pod, and now the npp is updated once after the Pods are grouped by nodename. It reduces the pressure on apiserver


furykerry (commented):
/lgtm

PR #1431, opened 2023-09-28
Add PodProbe dependency

ls-2018:


zmberg:
@ls-2018 Can you also add the check, https://github.com/openkruise/kruise/blob/master/pkg/webhook/imagepulljob/validating/imagepulljob_create_update_handler.go#L51

ls-2018:
@zmberg added.

PR #1440, opened 2023-10-27
Limit active image pulling job parallelism in global level

veophi:

### DIFF
--- a/pkg/controller/imagepulljob/imagepulljob_controller.go
+++ b/pkg/controller/imagepulljob/imagepulljob_controller.go
@@ -31,7 +31,9 @@ import (
 	utildiscovery "github.com/openkruise/kruise/pkg/util/discovery"
 	"github.com/openkruise/kruise/pkg/util/expectations"
 	utilfeature "github.com/openkruise/kruise/pkg/util/feature"
+	"github.com/openkruise/kruise/pkg/util/globallimiter"
 	utilimagejob "github.com/openkruise/kruise/pkg/util/imagejob"
+	"github.com/openkruise/kruise/pkg/util/requeueduration"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


pkg/util/globallimiter/limiter.go
```diff
@@ -0,0 +1,73 @@
+package globallimiter
+
+import (
+	"sync"
+	"time"
+)
+
+type GlobalLimiter interface {
```
furykerry, line 8: plz give comment about what the func do and what the params means

pkg/controller/imagepulljob/imagepulljob_controller.go
```diff
@@ -42,34 +44,48 @@ import (
...
 // and Start it when the Manager is Started.
-func Add(mgr manager.Manager) error {
+func Add(mgr manager.Manager) (err error) {
 	if !utildiscovery.DiscoverGVK(controllerKind) || !utilfeature.DefaultFeatureGate.Enabled(features.KruiseDaemon) ||
 		!utilfeature.DefaultFeatureGate.Enabled(features.ImagePullJobGate) {
 		return nil
 	}
+
+	if utilfeature.DefaultFeatureGate.Enabled(features.ImagePullJobGlobalParallelism) {
+		globalLimiter, err = initializedGlobalLimiter(20 * time.Minute)
```
furykerry, line 84: plz replace 20* min with const params, and comment about the relationship with defaultRequeueTime

pkg/util/globallimiter/limiter.go
```diff
@@ -0,0 +1,73 @@
...
+		return true
+	}
+	if len(g.activeKeys) < g.limit {
+		g.activeKeys[key] = time.Now()
+		return true
+	}
+	return false
+}
+
+func (g *realGlobalLimiter) gcDeadKey(timeout time.Duration) {
```
furykerry, line 51: plz comment about why we need gc keys, it that means the pulling job can be counted in the limit for almost timeout time?

veophi: added comments.

PR #1441, opened 2023-11-01
Add controller-revision-hash label for ImageListPullJob

veophi:
Add controller-revision-hash label for ImageListPullJob.

To solve the problem that the ImagePullJobs created by ImageListPullJob may be deleted if webhook/people mutated them.


furykerry (commented):
/lgtm

PR #1480, opened 2023-12-29
fix(statefulset): fix maxUnavailable for rolling upgrades not taking …

Yesphet:
…into account pods that fail later in the updateIndexes

### 0. Backgroud

Advanced statefulset provides MaxUnavailable guarantee that the number of unavailable pods during the update cannot exceed this value. However, in the current implementation code, there is no consideration for unavailable pods with smaller ordinal or later in the update order. For example, for an asts with maxUnavailable= 1 and replicas=3, if the asts updates while pod-0 is in a crash, it will still trigger a rolling upgrade, and 2 pods will be unavailable because pod-2 is updated and restarted.  

It can be easily reproduced in this way：

1. create a asts 

```
kind: StatefulSet
metadata:
  name: test
spec:
  selector:
    matchLabels:
      app: busybox-test
  replicas: 3
  podManagementPolicy: Parallel
  updateStrategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 1
  template:
    metadata:
      labels:
        app: busybox-test
    spec:
      containers:
      - name: busybox
        image: busybox
        env:
        - name: MY_POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
        - name: Restart
          value: "2"
        command: ["sh","-c"]
        args:
        - |-
          [ "$MY_POD_NAME" == "test-0" ] && exit 1;tail -f /dev/null
```
2. List pods, test-0 is in crash
image

3.  Update an env of the container in asts to trigger rolling update
```
kubectl patch asts test --type='json' -p='[{"op": "replace", "path": "/spec/template/spec/containers/0/env/1/value", "value":"3"}]'
```

4. We can see that test-2 is deleted which breaking the maxUnavailable guarantee
image

I have tested k8s official statefulset with [MaxUnavailable feature gate enabled](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#maximum-unavailable-pods), it can still guarantee that the number of unavailable pods does not exceed in this scenario.

So in this PR, I separate out the update pods logics to a new function `rollingUpdateStatefulsetPods`, counts all unavailable pods before update pods.

no

refer to part 0

In k8s official statefulsets controller, if the unavailable pods count reach the MaxUnavailable limit, the rolling update won't make progress even if the first pod in the update sequence is unavailable. But in my implementation, rolling updates are still process in this case because update one unavailable pod does not increase the total number of unavailable Pods.

### DIFF
--- a/pkg/controller/statefulset/stateful_set_control.go
+++ b/pkg/controller/statefulset/stateful_set_control.go
@@ -712,37 +712,37 @@ func (ssc *defaultStatefulSetControl) updateStatefulSet(
 		return &status, nil
 	}
 
-	// TODO: separate out the below update only related logic
+	return ssc.rollingUpdateStatefulsetPods(
+		set, &status, currentRevision, updateRevision, revisions, pods, replicas, minReadySeconds,
+	)
+}
+
+func (ssc *defaultStatefulSetControl) rollingUpdateStatefulsetPods(
+	set *appsv1beta1.StatefulSet,
+	status *appsv1beta1.StatefulSetStatus,
+	currentRevision *apps.ControllerRevision,
+	updateRevision *apps.ControllerRevision,
+	revisions []*apps.ControllerRevision,
+	pods []*v1.Pod,
+	replicas []*v1.Pod,
+	minReadySeconds int32,
+) (*appsv1beta1.StatefulSetStatus, error) {
 
 	// If update expectations have not satisfied yet, skip updating pods
 	if updateSatisfied, _, updateDirtyPods := updateExpectations.SatisfiedExpectations(getStatefulSetKey(set), updateRevision.Name); !updateSatisfied {
 		klog.V(4).Infof("Not satisfied update for %v, updateDirtyPods=%v", getStatefulSetKey(set), updateDirtyPods)
-		return &status, nil
-	}
-
-	// we compute the minimum ordinal of the target sequence for a destructive update based on the strategy.
-	maxUnavailable := 1
-	if set.Spec.UpdateStrategy.RollingUpdate != nil {
-		if set.Spec.UpdateStrategy.RollingUpdate.Paused {
-			return &status, nil
-		}
-
-		maxUnavailable, err = intstrutil.GetValueFromIntOrPercent(intstrutil.ValueOrDefault(set.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable, intstrutil.FromInt(1)), int(*set.Spec.Replicas), false)
-		if err != nil {
-			return &status, err
-		}
-		// maxUnavailable should not be less than 1
-		if maxUnavailable < 1 {
-			maxUnavailable = 1
-		}
+		return status, nil
 	}
 
 	// refresh states for all pods
 	var modified bool
 	for _, pod := range pods {
+		if pod == nil {
+			continue
+		}
 		refreshed, duration, err := ssc.refreshPodState(set, pod, updateRevision.Name)
 		if err != nil {
-			return &status, err
+			return status, err
 		} else if duration > 0 {
 			durationStore.Push(getStatefulSetKey(set), duration)
 		}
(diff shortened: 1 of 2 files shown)


Yesphet:
@veophi @Fei-Guo  Could you please help to review this PR when you are free?

Yesphet:
@veophi @Fei-Guo @zmberg   It's been two weeks, and I still haven't received any feedback on this PR. I understand that everyone is busy, I just want to ensure this PR hasn't been overlooked. Your time and attention to it would be much appreciated ~

furykerry:
sorry, guys are fully occupied these days, i will take a look at this patch.

pkg/controller/statefulset/stateful_set_control.go
```diff
@@ -707,118 +707,145 @@ func (ssc *defaultStatefulSetControl) updateStatefulSet(
...
 	updateIndexes := sortPodsToUpdate(set.Spec.UpdateStrategy.RollingUpdate, updateRevision.Name, *set.Spec.Replicas, replicas)
 	klog.V(3).Infof("Prepare to update pods indexes %v for StatefulSet %s", updateIndexes, getStatefulSetKey(set))
-	minWaitTime := appsv1beta1.MaxMinReadySeconds * time.Second
 	// update pods in sequence
 	for _, target := range updateIndexes {
 
+		// the unavailable pods count exceed the maxUnavailable and the target is available, so we can't process it,
+		// wait for unhealthy Pods on update
+		if len(unavailablePods) >= maxUnavailable && !unavailablePods.Has(replicas[target].Name) {
+			klog.V(4).Infof(
```
furykerry, line 815: 1. if "getPodRevision(replicas[target]) == updateRevision.Name", we can continue the loop. 
2. we can count the pods blocked by maxUnavailable, and log the error in one place so as to avoid duplicated error logging for every blocked pods

pkg/controller/statefulset/stateful_set_control.go
```diff
@@ -707,118 +707,145 @@ func (ssc *defaultStatefulSetControl) updateStatefulSet(
...
+	updateRevision *apps.ControllerRevision,
+	revisions []*apps.ControllerRevision,
+	pods []*v1.Pod,
+	replicas []*v1.Pod,
+	minReadySeconds int32,
+) (*appsv1beta1.StatefulSetStatus, error) {
+
 	// for the OnDelete strategy we short circuit. Pods will be updated when they are manually deleted.
 	if set.Spec.UpdateStrategy.Type == apps.OnDeleteStatefulSetStrategyType {
-		return &status, nil
```
furykerry, line 712: consider move this checking out of the function, so that the function only deal with the case of rolling update.

Yesphet: moved in  [d4cecdd18060f7bcc3eeb97364ea3d77b93782d7..46e4e4111c32ab60a122f3e0fcdc5e9a74628bdd](https://github.com/openkruise/kruise/compare/d4cecdd18060f7bcc3eeb97364ea3d77b93782d7..46e4e4111c32ab60a122f3e0fcdc5e9a74628bdd)

pkg/controller/statefulset/stateful_set_control.go
Yesphet: [d4cecdd18060f7bcc3eeb97364ea3d77b93782d7..46e4e4111c32ab60a122f3e0fcdc5e9a74628bdd](https://github.com/openkruise/kruise/compare/d4cecdd18060f7bcc3eeb97364ea3d77b93782d7..46e4e4111c32ab60a122f3e0fcdc5e9a74628bdd)

1. changed
2. this log only be printed once per reconcilation when the first pod is blocked. By combining the first blocked pod with the L808 update indexes log, we can get all blocked pods if needed.

zmberg:
@Yesphet good.

Yesphet:
@furykerry @zmberg  thanks guys~

PR #1487, opened 2024-01-11
Revert "fix deleteAvailableLimit bug (#1481)"

zmberg:
This fix is incorrect.

This reverts commit f4e238fd8d45ff81b0c102cf58228daa66c0d5bc.


huaouo:
Could you please explain why should we keep to use deleteReadyLimit rather than deleteAvailableLimit despite that deleteReadyLimit is calculated from maxUnavailable and totalUnavailable

PR #1489, opened 2024-01-21
the CloneSet pod lifecycle can be transformed to Normal only if ContainerReady is true

### LINKED ISSUE #1485: [IMPROVEMENT] [CloneSet] [Lifecycle] Kruise v1.4+ works not so well with kube-scheduler <= v1.21
### Kruise CloneSet makes a BUG of the old version kube-scheduler appear more frequently

CloneSet Lifecycle adds PreparingNormal since v1.4. If a cloneset pod has no lifecycle hooks, the POD will transform from PreparingNormal to Normal very quickly. As we know, the lifecycle state is recorded by labels, this means that when the POD is created, the label will be modified quickly. 

If you use kube-scheduler (<=v1.21) for scheduling, when a pod is created, the scheduler will watch and enqueue it to the scheduling queue. After the scheduling algorithm is completed (finished finding a node), BIND will be performed concurrently. Only when BIND succeeds will the POD be considered scheduled. If someone modifies the pod before the whole process is completed, the POD will be requeued. Because Kruise modified the POD labels, the kube-scheduler will not skipPODUpdate and will retry to schedule, this will cause continuous annoying ERROR log and Warning events  (but will not actually affect POD scheduling). 
https://github.com/kubernetes/kubernetes/issues/101877
https://github.com/kubernetes/kubernetes/issues/98828

#### Maybe we can change the CloneSet Behavior?  
Of course, there is no big problem with the current behavior of CloneSet.It is a BUG of old kube-scheduler, and I have fix my scheduler to avoid this problem.But anyone using old version k8s (<=v1.21) and upgrading kruise to v1.4+ will face the same confusion as I did. 

And I think it is more reasonable that a POD will transform from PreparingNormal to Normal only after it is scheduled (NodeName not empty).I have tried this in my cluster, it seems to work well. 

Use Kruise v1.4+ and kube-scheduler less than 1.21.
Create a CloneSet, and delete POD of this CloneSet repeatedly util you get the event like this:

> Warning   FailedScheduling    POD fa210e4e-3531-4974-bf33-553a88986e5f is in the cache, so can't be assumed

It is easy to reproduce, I think it should be easier to reproduce when the cluster(kube-scheduler) is not busy.

**Environment**:
- Kruise:  v1.5:
- Kubernetes version: v1.20.8
- Install details :
- Others:

chenpeicheng9:
#### The CloneSet pod lifecycle can be transformed to Normal only if ContainerReady is true.
Sorry for being late. 
Last week After I created issue #1485, I appreciate that the Kruise team replied quickly and pointed out that it may be better "to Normal after Pod Ready"  @zmberg 
I've been thinking about this issue these days. I believe using the 'PodReady'(k8s.io/api/core/v1.PodReady, 'Ready' in yaml, same as below) condition may not be appropriate. Consider the scenario with the 'markPodNotReady' set: the Pod's lifecycle can only be transformed to 'Normal' if 'PodReady' is true. On the other hand, the 'PodReady' condition will be set to false at preparingDelete/preparingUpdate state. This bidirectional influence adds complexity and can make the lifecycle confusing. 
So I suggest that we can use the "ContainerReady" condition, it would be simple enough.
A pod's lifecycle will be transformed from PreparingNormal to Normal only if its containers are all ready.
Furthermore, it is ok that a pod's lifecycle will be transformed to Normal only if its containers are all ready.

fixes #1485

### DIFF
--- a/apis/apps/pub/lifecycle.go
+++ b/apis/apps/pub/lifecycle.go
@@ -21,7 +21,7 @@ const (
 	LifecycleTimestampKey = "lifecycle.apps.kruise.io/timestamp"
 
 	// LifecycleStatePreparingNormal means the Pod is created but unavailable.
-	// It will translate to Normal state if Lifecycle.PreNormal is hooked.
+	// It will translate to Normal state if Lifecycle.PreNormal is hooked and pod condition ContainerReady is true.
 	LifecycleStatePreparingNormal LifecycleStateType = "PreparingNormal"
 	// LifecycleStateNormal is a necessary condition for Pod to be available.
 	LifecycleStateNormal LifecycleStateType = "Normal"
--- a/pkg/controller/cloneset/sync/cloneset_update.go
+++ b/pkg/controller/cloneset/sync/cloneset_update.go
@@ -169,9 +169,12 @@ func (c *realControl) refreshPodState(cs *appsv1alpha1.CloneSet, coreControl clo
 	var state appspub.LifecycleStateType
 	switch lifecycle.GetPodLifecycleState(pod) {
 	case appspub.LifecycleStatePreparingNormal:
-		if cs.Spec.Lifecycle == nil ||
-			cs.Spec.Lifecycle.PreNormal == nil ||
-			lifecycle.IsPodAllHooked(cs.Spec.Lifecycle.PreNormal, pod) {
+		// 1.the pod lifecycle can be transformed from PreparingNormal to Normal only if ContainerReady is true.
+		// If lifecycle transformed from PreparingNormal to Normal immediately after pod created,
+		// it may trigger a BUG of old version kube-scheduler with some probability.  https://github.com/openkruise/kruise/issues/1485
+		// 2.the semantic of this hook, let's leave it as it is:
+		// PreNormalHook: only if all hooks are hooked (or no hook), the pod can be transformed to Normal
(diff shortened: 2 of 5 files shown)


zmberg:
@chenpeicheng9 Can you communicate through dingding?

chenpeicheng9:
> @chenpeicheng9 Can you communicate through dingding?

sure

furykerry:
plz squash and sign the commits

chenpeicheng9:
As discussed, for now, I have removed the update-to-normal part. And made slight modifications to the comments. @zmberg.
Thanks, @furykerry  I've squashed and signed it now.

ABNER-1:
Hi, @chenpeicheng9. 
Are you still able to make progress on this PR? 

We’ve discussed it and concluded that converting pod to normal state after the pod scheduled could be a more effective approach.

MichaelRren:
> Hi, @chenpeicheng9. Are you still able to make progress on this PR?
> 
> We’ve discussed it and concluded that converting pod to normal state after the pod scheduled could be a more effective approach.

@ABNER-1 Should the author be unavailable to continue progress on this matter, may I request that it be reassigned to me?

ABNER-1:
> > Hi, @chenpeicheng9. Are you still able to make progress on this PR?
> > We’ve discussed it and concluded that converting pod to normal state after the pod scheduled could be a more effective approach.
> 
> @ABNER-1 Should the author be unavailable to continue progress on this matter, may I request that it be reassigned to me?

Absolutely, if you’re up for it, this is a fantastic idea! You could cherry-pick the commits from @chenpeicheng9  and carry on with the work.

PR #1503, opened 2024-02-09
fix: skip revision injection when pod matched UpdateStrategy.Selector

0xgj:
This pull request aims to address the issue of revision injection for pods that match the upgrade policy. Currently, if a pod that has already been upgrade is recreated due to machine or pod failure, it will rollback to the old revision, leading to a false-positive release observation. This PR proposes a fix by implementing comments (https://github.com/openkruise/kruise/blob/5421ee7c8e9e54ec38c1c1e20f4deb4a722f47a7/apis/apps/v1alpha1/sidecarset_types.go#L226-L228).

unit test and manually.


zmberg:
@0xgj UpdateStrategy.Selector field affects the injection logic is not appropriate, in addition, there have some scenarios can not be solved, this solution is not convenient to merge in for the time being. 

If you have new ideas, welcome to the community meeting to discuss!

PR #1512, opened 2024-02-27
update pub readiness check in pub validating

Spground:

Add PodStateConsistent check condition in PUB validating.

How to check whether Pod is available is inconsistent in PUB controllers and webhooks.  

https://github.com/openkruise/kruise/blob/5421ee7c8e9e54ec38c1c1e20f4deb4a722f47a7/pkg/controller/podunavailablebudget/podunavailablebudget_controller.go#L380

https://github.com/openkruise/kruise/blob/5421ee7c8e9e54ec38c1c1e20f4deb4a722f47a7/pkg/control/pubcontrol/pub_control_utils.go#L68

Make rolling update using inplace-update, delete Pod when inplace-update is not completed, which is supposed to be successful.


Spground:
您的来信我已收到，我会尽快回复！谢谢！

furykerry (commented):
/lgtm

PR #1515, opened 2024-02-29
CloneSet scale down schedule-failed pods firstly

veophi:
In the scene:
```
replicas: 20
partition: 10

current-revision:
   runningPods: 5
   scheduleFailedPods: 5
update-revision:
   runningPods: 5
   scheduleFailedPods: 5
```

when scaling down CloneSet from 20 to 10, **we hope that 10 running Pods are reserved. Howerver, 10 current-revision pods will be reserved in current kruise version**.


furykerry:
if the partition is 50%, then it is reasonable to keep 10 running pods. However if the partition is absolute value, keeping 10 running pods will break the semantic meaning of partition. Is it ok for you to change the partition to percent form?

Spground:
> However if the partition is absolute value, keeping 10 running pods

Why not upgra partition is absolute value, 

> > if the partition is 50%, then it is reasonable to keep 10 running pods. However if the partition is absolute value, keeping 10 running pods will break the semantic meaning of partition. Is it ok for you to change the partition to percent form?
> 
> ```
> replicas: 20
> partition: 50%
> 
> current-revision:
>    runningPods: 3
>    scheduleFailedPods: 7
> update-revision:
>    runningPods: 7
>    scheduleFailedPods: 3
> ```

Will extra 2 upgraded Pods rollback to current revision if we prefer scale down failed Pod?

Issue #1468, opened 2023-12-12
[feature request] Support `progressDeadlineSeconds` in CloneSet


ronaknnathani:
**What would you like to be added**:
Add support for `progressDeadlineSeconds` in CloneSet. `.spec.progressDeadlineSeconds` is an optional field in [Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/) that specifies the number of seconds one wants to wait for their Deployment to progress before the system reports back that the Deployment has [failed progressing](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#failed-deployment) - surfaced as a condition with `type: Progressing, status: "False" and reason: ProgressDeadlineExceeded` in the status of the resource.

Reference: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#progress-deadline-seconds

**Why is this needed**:
In the case of CloneSets, if the rollout is stuck for any reason - image pull errors, insufficient permissions, insufficient capacity, unhealthy app, etc. - currently there is no way to identify that the rollout is stuck by looking at the CloneSet status. With a feature like `progressDeadlineSeconds`, one can specify a time and if the rollout hasn't made any progress for that period, the CloneSet status can be updated to reflect that. The CloneSet controller would still continue to reconcile the requests, however, this status allows continuous delivery pipelines and any other controllers that build on top of CloneSet to know that the rollout failed.

ronaknnathani:
@furykerry do we need a proposal for this or do you think we can send a PR for review?

furykerry:
yes, plz draft a proposal for this feature in https://github.com/openkruise/kruise/blob/master/docs/proposals

furykerry:
@ronaknnathani we're drafting a [proposal](https://github.com/openkruise/kruise/pull/1520) for this feature, however the definition of progressDeadlineSeconds can be different from Deployment because cloneset support partition. Can you give us some feedback about the potential use case of progressDeadlineSeconds when partition is set ?

PR #1520, opened 2024-03-08
[Proposal] Add proposal for  adding  progressDeadlineSeconds in CloneSet

hantmac:
Add proposal for support progressDeadlineSeconds in CloneSet

To fix https://github.com/openkruise/kruise/issues/1468

### DIFF
--- a/docs/proposals/20240309-cloneset-support-progressDeadlineSeconds.md
+++ b/docs/proposals/20240309-cloneset-support-progressDeadlineSeconds.md
@@ -1,0 +1,161 @@
+---
+title: CloneSet
+authors:
+- "@hantmac"
+  reviewers:
+- "@Fei-Guo"
+- "@furykerry"
+- "@FillZpp"
+  creation-date: 2024-03-10
+  last-updated: 2024-03-10
+  status: implementable
+---
+
+# Support progressDeadlineSeconds in CloneSet
+Table of Contents
+=================
+
+- [Support progressDeadlineSeconds in CloneSet](#Support progressDeadlineSeconds in CloneSet)
+- [Table of Contents](#table-of-contents)
+  - [Motivation](#motivation)
+  - [Proposal](#proposal)
+    - [1. add .spec.progressDeadlineSeconds field](#1add-.spec.progressDeadlineSeconds-field)
+    - [2. The behavior of progressDeadlineSeconds](#2the-behavior-of-progressDeadlineSeconds)
+    - [3. handle the logic](#2handle-the-logic)
+
+## Motivation
+
+`.spec.progressDeadlineSeconds` is an optional field in Deployment that specifies the number of seconds one wants to wait for their Deployment to progress before the system reports back that the Deployment has failed progressing.
+Once the deadline has been exceeded, the Deployment controller adds a DeploymentCondition with the following attributes to the Deployment's `.status.conditions`:
+```
+type: Progressing
+status: "False"
+reason: ProgressDeadlineExceeded
+```
+
+This is useful for users to control the progress of the deployment.
+So we should add support for `progressDeadlineSeconds` in CloneSet as well.
+
+## Proposal
+Firstly, add the `progressDeadlineSeconds` field to the CloneSetSpec.
+Then add the handle logic in cloneSet controller to handle the `progressDeadlineSeconds` field.
+
+### 1. add .spec.progressDeadlineSeconds field
+The default value of `progressDeadlineSeconds` is 600 seconds according to the [official document](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#progress-deadline-seconds).
+```yaml
+apiVersion: apps.kruise.io/v1alpha1
+kind: CloneSet
+metadata:
+  name: cloneset-example
+spec:
+    replicas: 3
+    progressDeadlineSeconds: 600
+    selector:
+        matchLabels:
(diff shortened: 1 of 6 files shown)


docs/proposals/20240309-cloneset-support-progressDeadlineSeconds.md
```diff
@@ -0,0 +1,94 @@
...
+        metadata:
+        labels:
+            app: nginx
+        spec:
+        containers:
+        - name: nginx
+            image: nginx:1.14.2
+ ```
+
+### 2. handle the logic
```
furykerry, line 65: cloneset support partition, so the definition of progressDeadlineSeconds may be different from deployment. If the cloneset is paused because of partition shall we still count the paused time in progress deadline ? 
1. One possible redefinition of progressDeadlineSeconds is the time to reach completion OR "the paused state". 
2. alternatively we only support progressDeadlineSeconds if partition is not set. 
I think we need more feedback from the community

hantmac: Yes, we should not count the time when cloneset is paused by the partition.  
Only supporting `progressDeadlineSeconds` if partition is not set may be easy to implement. 
More feedback is nice. Maybe we can make a vote about these two choice.

hantmac:
@furykerry the proposal is read for review 😄

docs/proposals/20240309-cloneset-support-progressDeadlineSeconds.md
```diff
@@ -0,0 +1,143 @@
...
+    ...
+}
+```
+
+When the CloneSet reaches the "paused" state, we should reset the timer to avoid the progress deadline being exceeded.
+```go
+func (c *CloneSetController) syncCloneSetStatus(cloneSet *appsv1alpha1.CloneSet, newStatus *appsv1alpha1.CloneSetStatus) error {
+    ...
+	// reset the starttime when the CloneSet reaches the "paused" state or complete state
+	if cloneSet.Status.UpdatedReadyReplicas == cloneSet.Status.Replicas {
```
furykerry, line 125: paused state also include the case, when replicas - updatedReplicas = partition

docs/proposals/20240309-cloneset-support-progressDeadlineSeconds.md
```diff
@@ -0,0 +1,143 @@
...
+After the discussion of the community, we choose the first interpretation.So we should re-set the  progressDeadlineSeconds when the CloneSet reach completion OR "the paused state".
+
+### 3. handle the logic
+In cloneset controller, we should add the logic to handle the `progressDeadlineSeconds` field. We firstly add a timer to check the progress of the CloneSet. 
+If the progress exceeds the `progressDeadlineSeconds`, we should add a CloneSetCondition to the CloneSet's `.status.conditions`: 
+```go
+// add a timer to check the progress of the CloneSet
+if cloneSet.Spec.ProgressDeadlineSeconds != nil {
+    // handle the logic
+	starttime := time.Now()
```
furykerry, line 81: shall we save the startime somewhere ? the data may be lost across reconciliation or after kruise-manager restart

docs/proposals/20240309-cloneset-support-progressDeadlineSeconds.md
```diff
@@ -0,0 +1,143 @@
...
+        spec:
+        containers:
+        - name: nginx
+            image: nginx:1.14.2
+ ```
+
+### 2. The behavior of progressDeadlineSeconds
+However, the behavior of `progressDeadlineSeconds` in CloneSet might differ from its behavior in Deployment due to the support of partition in CloneSet. If a CloneSet is paused due to partition, it's debatable whether the paused time should be included in the progress deadline.
+Here are two possible interpretations of `progressDeadlineSeconds` in the context of CloneSet:
+1. `progressDeadlineSeconds` could be redefined as the time taken for the CloneSet to reach completion or the "paused" state due to partition. In this case, the time during which the CloneSet is paused would be included in the progress deadline.
```
furykerry, line 69: > In this case, the time during which the CloneSet is paused would be included in the progress deadline.

It seems the meaning is the opposite. the time during which the CloneSet is paused would NOT be included in the progress deadline.

hantmac: Sorry, it should be  `the time during which the CloneSet is paused would NOT be included in the progress deadline.`

docs/proposals/20240309-cloneset-support-progressDeadlineSeconds.md
hantmac: I think we have no need to save the starttime.  If  kruise-manager restart, we just reset the starttime.

furykerry: deployment save time in the LastUpdateTime of the progressing condition, we can use the similar field

hantmac: Yes, updated.

furykerry (commented):
plz squash the commits 
/lgtm

hantmac:
> plz squash the commits /lgtm

Thanks. Squashed.

furykerry (commented):
/lgtm

furykerry:
@zmberg PTAL

Issue #1530, opened 2024-03-21
[BUG] when statefulset reserveOrdinals exist, scale down pvc failed


zmberg:
**What happened**:
1. Create StatefulSet reserveOrdinals=[0], replicas = 3, pvcTemplate = log
2. Currently will create three pvc=log-web-1, log-web-2, log-web-3
3. Scale down StatefulSet replicas=2, But log-web-3 also exist.
**What you expected to happen**:
Pvc log-web-3 will be deleted.
**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kruise version:
- Kubernetes version (use `kubectl version`):
- Install details (e.g. helm install args):
- Others:

furykerry:
fixed by https://github.com/openkruise/kruise/pull/1531

PR #1531, opened 2024-03-21
Fix when statefulset reserveOrdinals exist, scale down pvc failed

zmberg:
#1530

### DIFF
--- a/pkg/controller/statefulset/stateful_pod_control_test.go
+++ b/pkg/controller/statefulset/stateful_pod_control_test.go
@@ -21,14 +21,21 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
+	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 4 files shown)


pkg/controller/statefulset/stateful_pod_control_test.go
```diff
@@ -849,3 +853,402 @@ func collectEvents(source <-chan string) []string {
...
+				}
+				return metav1.OwnerReference{
+					APIVersion: "v1",
+					Kind:       "Pod",
+					Name:       podName,
+				}
+			},
+		},
+		{
+			name: "reserveOrdinals is [2,3], scaleDown=false, whenScaled=Retain, whenDeleted=Delete",
```
furykerry, line 1035: what about make reserveOrdinals cover the boundary case,e.g. "reserveOrdinals is [2,4]"

Issue #1557, opened 2024-04-01
[feature request] pod-readiness-controller modify Pod use Patch instead of Update


Spground:
**What would you like to be added**:

**Why is this needed**:

zmberg:
Anyone interested in submitting this patch? 

related code: https://github.com/openkruise/kruise/blob/master/pkg/controller/podreadiness/pod_readiness_controller.go#L132

Methods are as follows:
```
		// patch pod condition
		status := v1.PodStatus{
			Conditions: []v1.PodCondition{
				{
					Type:               appspub.KruisePodReadyConditionType,
					Status:             v1.ConditionTrue,
					LastTransitionTime: metav1.Now(),
				},
			},
		}
		by, _ := json.Marshal(status)
		patchBody := fmt.Sprintf(`{"status":%s}`, string(by))
		rcvObject := &v1.Pod{ObjectMeta: metav1.ObjectMeta{Namespace: pod.Namespace, Name: pod.Name}}
		err := r.Status().Patch(context.TODO(), rcvObject, client.RawPatch(types.StrategicMergePatchType, []byte(patchBody)))
```

BruceAko:
/assign
I am willing to submit the patch.

PR #1560, opened 2024-04-02
fix: pod-readiness-controller use Patch to modify Pod instead of Update

BruceAko:
#1557

### DIFF
--- a/pkg/controller/podreadiness/pod_readiness_controller.go
+++ b/pkg/controller/podreadiness/pod_readiness_controller.go
@@ -18,6 +18,8 @@ package podreadiness
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 	"time"
 
 	appspub "github.com/openkruise/kruise/apis/apps/pub"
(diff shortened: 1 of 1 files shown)


BruceAko:
@furykerry PTAL

pkg/controller/podreadiness/pod_readiness_controller.go
```diff
@@ -129,12 +132,23 @@ func (r *ReconcilePodReadiness) Reconcile(_ context.Context, request reconcile.R
...
 		}
 
-		pod.Status.Conditions = append(pod.Status.Conditions, v1.PodCondition{
-			Type:               appspub.KruisePodReadyConditionType,
-			Status:             v1.ConditionTrue,
-			LastTransitionTime: metav1.Now(),
-		})
-		return r.Status().Update(context.TODO(), pod)
+		// patch pod condition
+		status := v1.PodStatus{
```
zmberg, line 136: It is not necessary to use the RetryOnConflict method when using the patch method.

BruceAko: Is it ok now?

BruceAko:
@zmberg @furykerry @Fei-Guo Can this pull request be merged now?

Issue #1536, opened 2024-03-21
[feature request] When CloneSet volumeClaimTemplates changed, Pod must ReCreate update


zmberg:
**What would you like to be added**:
Question: When CloneSet volumeClaimTemplates and Image changed, Pod will in-place update, then volumeClaimTemplates not updated to Pod volumes, because the volumes field is only recalculated when the Pod is created.

**What happened**:
1. Create CloneSet with volumeClaimTemplates, as follows:
```yaml
apiVersion: apps.kruise.io/v1alpha1
kind: CloneSet
metadata:
  name: web
spec:
  updateStrategy:
    type: InPlaceIfPossible
  replicas: 2
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
        - name: nginx
          image: nginx:stable-alpine3.17
          env:
          - name: version
            value: v1
          volumeMounts:
            - name: www-data
              mountPath: /usr/share/nginx/html
  volumeClaimTemplates:
    - metadata:
        name: www-data
      spec:
        accessModes: ["ReadWriteOnce"]
        resources:
          requests:
            storage: 1Gi
```
2. Update image tag from stable-alpine3.17 to stable-alpine, and volumeClaimTemplates storage from 1Gi to 2Gi.
```yaml
apiVersion: apps.kruise.io/v1alpha1
kind: CloneSet
spec:
  template:
    spec:
      containers:
        - name: nginx
          image: nginx:stable-alpine
  volumeClaimTemplates:
    - metadata:
        name: www-data
      spec:
        accessModes: ["ReadWriteOnce"]
        resources:
          requests:
            storage: 2Gi
```
3. Update cloneSet, but pods only in-place update and pvc capacity don't change.
```bash
% kubectl get clonesets
NAME   DESIRED   UPDATED   UPDATED_READY   READY   TOTAL   AGE
web    2         2         2               2       2       2m27s

% kubectl get pods
NAME        READY   STATUS    RESTARTS      AGE
web-c48gr   1/1     Running   1 (16s ago)   2m33s
web-mjmvl   1/1     Running   1 (23s ago)   2m33s

% kubectl get pvc
NAME                 STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
www-data-web-c48gr   Bound    pvc-4100c7c4-0fb0-41b6-90e1-696cfdd970c8   1Gi        RWO            standard       2m38s
[...]

PR #1561, opened 2024-04-03
[feature] When CloneSet volumeClaimTemplates changed, Pod must ReCreate update

ABNER-1:

Implemented "When volumeClaimTemplates are modified, the pod must ReCreate update"

1. If both the volumeClaimTemplates spec and the image are modified -> recreate.
2. If only the volumeClaimTemplates spec is modified -> no changes util next template spec changes
3. next modification of the image or other configurations -> recreate update
4. To prevent unnecessary pod recreate update, the initial upgrade of CloneSets that previously utilized the volumeClaimTemplates feature will bypass the comparison of historical revisions lacking hash values after the Kruise upgrade.
    - Abandoned the implementation of calculating the hash from pod volumes and existing PVCs: some PVCs have inherent fields that do not exist in volumeClaimTemplates, leading to calculated hash values that are inconsistent with the hash values calculated based on volumeClaimTemplates.

#### Corner case in 4
  1. Before the Kruise upgrade, CloneSet with image 1 and vcTemplates size 1Mi.
  2. After the Kruise upgrade, an update was applied to use image 2 and vcTemplate size 10Mi.  However, due to the absence of hash values for historical revisions, this update will not trigger a recreation, and the actual pod will be like image 2 and vcTemplate size 1Mi. The revision will incorrectly record the hash value for the 10Mi size. Future updates in this situation need to be cautious.
  4. It is recommended not to modify the vcTemplate during the first update after the Kruise upgrade.

 detail in #1536

detail in #1536

detail case in #1536
#### Main test cases:

 1. image 1, vctemplate size of 1MB.
 2. Image 2, vcTemplate size of 1MB.
 3. Image 3, vcTemplate size of 1MB. (in function testUpdate)
 5. Image 4, vcTemplate size of 10MB => recreate.
 6. Image 2, vcTemplate size of 10MB => (expected in-place upgrade)
    --- Although the history revision hash is consistent, the vct hash is not, update history revision id and hash value
 7. Image 4, vcTemplate size of 10MB => in-place upgrade --- Update history revision.
 8. Image 4, vcTemplate size of 100MB => (temporarily unchanged) --- Do not generate a new revision.
 9. Image 5, vcTemplate size of 100MB => recreate.

### DIFF
--- a/pkg/controller/cloneset/cloneset_controller.go
+++ b/pkg/controller/cloneset/cloneset_controller.go
@@ -39,6 +39,8 @@ import (
 	imagejobutilfunc "github.com/openkruise/kruise/pkg/util/imagejob/utilfunction"
 	"github.com/openkruise/kruise/pkg/util/ratelimiter"
 	"github.com/openkruise/kruise/pkg/util/refmanager"
+	"github.com/openkruise/kruise/pkg/util/volumeclaimtemplate"
+
 	apps "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -50,7 +52,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/record"
-	"k8s.io/klog/v2"
+	klog "k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/controller/history"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -427,14 +429,32 @@ func (r *ReconcileCloneSet) getActiveRevisions(cs *appsv1alpha1.CloneSet, revisi
 		return nil, nil, collisionCount, err
 	}
 
+	// Here, we do not consider the case where the volumeClaimTemplates Hash was not added before the upgrade feature.
+	// 1. It is not possible to actually extract the previous volumeClaimTemplates Hash.
+	// 2. the hash must exist in all revisions that have been updated after the feature is enabled.
+	VCTHashEqual := func(revision *apps.ControllerRevision, needle *apps.ControllerRevision) bool {
+		needleVolumeClaimHash, _ := volumeclaimtemplate.GetVCTemplatesHash(needle)
+		tmpHash, exist := volumeclaimtemplate.GetVCTemplatesHash(revision)
+		// vcTemplate hash may not exist in old revision, use hash of new revision instead
+		if !exist || tmpHash != needleVolumeClaimHash {
+			return false
+		}
+		return true
+	}
+
+	// When there is a change in the PVC only, no new revision will be generated.
 	// find any equivalent revisions
 	equalRevisions := history.FindEqualRevisions(revisions, updateRevision)
 	equalCount := len(equalRevisions)
-
 	if equalCount > 0 && history.EqualRevision(revisions[revisionCount-1], equalRevisions[equalCount-1]) {
 		// if the equivalent revision is immediately prior the update revision has not changed
 		updateRevision = revisions[revisionCount-1]
 	} else if equalCount > 0 {
+		lastEqualRevision := equalRevisions[equalCount-1]
+		if !VCTHashEqual(lastEqualRevision, updateRevision) {
+			klog.Infof("revision %v vct hash will be updated from %v to %v", lastEqualRevision.Name, lastEqualRevision.Annotations[volumeclaimtemplate.HashAnnotation], updateRevision.Annotations[volumeclaimtemplate.HashAnnotation])
+			lastEqualRevision.Annotations[volumeclaimtemplate.HashAnnotation] = updateRevision.Annotations[volumeclaimtemplate.HashAnnotation]
+		}
 		// if the equivalent revision is not immediately prior we will roll back by incrementing the
 		// Revision of the equivalent revision
 		updateRevision, err = r.controllerHistory.UpdateControllerRevision(equalRevisions[equalCount-1], updateRevision.Revision)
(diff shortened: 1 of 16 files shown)


pkg/util/volumeclaimtemplate/volume_templates_hash.go
```diff
@@ -0,0 +1,136 @@
...
+	return "", false
+}
+
+func PatchVCTemplateHash(revision *apps.ControllerRevision, vcTemplates []v1.PersistentVolumeClaim) {
+	if len(revision.Annotations) == 0 {
+		revision.Annotations = make(map[string]string)
+	}
+	if len(vcTemplates) != 0 {
+		// get hash of vct
+		vcTemplateHash := NewVolumeClaimTemplatesHasher().GetExpectHash(vcTemplates)
```
zmberg, line 120: It's not necessary to initialize an object for each call, is it better to have a global variable?

pkg/util/volumeclaimtemplate/volume_templates_hash.go
```diff
@@ -0,0 +1,136 @@
...
+	} else {
+		revision.Annotations[VolumeClaimTemplateHashAnnotation] = ""
+	}
+}
+
+func CanVCTemplateInplaceUpdate(oldRevision, newRevision *apps.ControllerRevision) bool {
+	newVCTemplatesHash, newExist := GetVCTemplatesHash(newRevision)
+	oldVCTemplatesHash, oldExist := GetVCTemplatesHash(oldRevision)
+	// oldExist = false, pass this case to avoid unnecessary recreate when first open this feature
+	// newExist = false, actually this case not exist
```
zmberg, line 131: If this is not the case, isn't it unnecessary to judge newExist?

pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -427,14 +428,31 @@ func (r *ReconcileCloneSet) getActiveRevisions(cs *appsv1alpha1.CloneSet, revisi
...
+			return false
+		}
+		return true
+	}
+
 	// find any equivalent revisions
 	equalRevisions := history.FindEqualRevisions(revisions, updateRevision)
 	equalCount := len(equalRevisions)
-
 	if equalCount > 0 && history.EqualRevision(revisions[revisionCount-1], equalRevisions[equalCount-1]) {
```
zmberg, line 447: Let's add a comment that states that only PVC changes are not newly generated revisions.

pkg/controller/cloneset/cloneset_controller.go
```diff
@@ -427,14 +428,32 @@ func (r *ReconcileCloneSet) getActiveRevisions(cs *appsv1alpha1.CloneSet, revisi
 		return nil, nil, collisionCount, err
 	}
 
+	// Here, we do not consider the case where the volumeClaimTemplates Hash was not added before the upgrade feature.
+	// 1. It is not possible to actually extract the previous volumeClaimTemplates Hash.
+	// 2. the hash must exist in all revisions that have been updated after the feature is enabled.
+	EqualVCTHashEqual := func(revision *apps.ControllerRevision, needle *apps.ControllerRevision) bool {
```
furykerry, line 434: EqualVCTHashEqual  -> VCTHashEqual

pkg/util/volumeclaimtemplate/volume_templates_hash.go
```diff
@@ -0,0 +1,111 @@
...
+
+type PVCGetter = func(string) *v1.PersistentVolumeClaim
+
+type volumeClaimTemplatesHasher struct{}
+
+func NewVolumeClaimTemplatesHasher() VolumeClaimTemplatesHasher {
+	return &volumeClaimTemplatesHasher{}
+}
+
+func (h *volumeClaimTemplatesHasher) GetExpectHash(templates []v1.PersistentVolumeClaim) uint64 {
```
furykerry, line 54: is this func ever used outside the package, if not, plz make it private func

pkg/util/volumeclaimtemplate/volume_templates_hash_test.go
```diff
@@ -0,0 +1,387 @@
...
+		templates []v1.PersistentVolumeClaim
+	}
+	tests := []struct {
+		name string
+		args args
+		want uint64
+	}{
+
+		{
+			name: "single-pvc",
```
furykerry, line 23: add the test case of no pvc, that is , no pvc template defined in cloneset

pkg/util/inplaceupdate/inplace_update_defaults.go
```diff
@@ -234,6 +236,11 @@ func defaultCalculateInPlaceUpdateSpec(oldRevision, newRevision *apps.Controller
 		return nil
 	}
 
+	canInPlace := volumeclaimtemplate.CanVCTemplateInplaceUpdate(oldRevision, newRevision)
```
furykerry, line 239: plz add an ut for CanUpdateInPlace, so that to cover the changed code

pkg/util/volumeclaimtemplate/volume_templates_hash.go
ABNER-1: re-add this judgement again.

One corner case may exist:
1. cloneset A with image 1 vct size 1Gi
2. kriuse update to new version, next revision will patch hash
3. 1st update cloneset A with image2 vct size 1Gi -> in-place update and revision hash is not nil
4. 1st update over
5. kruise rollback, next revision will not patch hash
6. 2rd update cloneset A with image3 vct size 1Gi, partition 50% -> in-place update and revision hash is nil
7. kruise re-update
8. 2rd update, with partition 0 

In step8 of this case, new revision hash does not exist and old revision exists.
It might be necessary to include a case where the new revision is non-existent as well.

ABNER-1:
I will add a feature gate about this feature, and set default value to false.

ABNER-1:
squash commits and re test

furykerry (commented):
/lgtm

HURUIZHE:
I would like to ask a question, since the VolumeClaimTemplate is not stored in the ControllerRevision of the controller, how can I roll it back?

PR #1566, opened 2024-04-06
fix: WorkloadSpread feature judgment

ls-2018:


zmberg:
@ls-2018 Creating workloadSpread fails when featureGate is false, refer code: https://github.com/openkruise/kruise/blob/master/pkg/webhook/workloadspread/validating/workloadspread_create_update_handler.go#L46

ls-2018:
image

plz  ignore the file directory

zmberg:
@ls-2018 It shouldn't be allowed to be created, and when it is created it just gives an error telling him that he should open the featrueGate to do so.

PR #1579, opened 2024-04-16
Fixed an issue where pub webhooks unexpectedly return error when PUB is NOT FOUND

### LINKED ISSUE #1567: [BUG] pub webhooks unexpectedly return error when PUB is NOT FOUND
**What happened**:

PUB may unexpectedly interrupt Pod gc issued by KCM, which can lead Pod leak if KCM gc did not retry or retry many hours later.

PUB webhooks never interrupt Pod gc.

Delete workload let's say Sts or CloneSet, then Pod will be deleted by KCM gc later. Sometimes, Pod to delete will be leaking there for a lone time.

The root cause is we return error when PUB CR is deleted in RetryOnConflict. Related codes is here, https://github.com/openkruise/kruise/blob/8bb89648de6c4944239c03a9a4a22f18e9616193/pkg/control/pubcontrol/pub_control_utils.go#L105

The solution is simple, just check error type as we can , ignore it if it is NotFound error.

**Environment**:
- Kruise version:
- Others:

opencmit2:
…is NOT FOUND

Fixed an issue #1567  where pub webhooks unexpectedly return error when PUB is NOT FOUND

fixes #1567

### DIFF
--- a/pkg/control/pubcontrol/pub_control_utils.go
+++ b/pkg/control/pubcontrol/pub_control_utils.go
@@ -102,7 +102,7 @@ func PodUnavailableBudgetValidatePod(pod *corev1.Pod, operation policyv1alpha1.P
 		if refresh {
 			pubClone, err = kubeClient.GetGenericClient().KruiseClient.PolicyV1alpha1().
 				PodUnavailableBudgets(pub.Namespace).Get(context.TODO(), pub.Name, metav1.GetOptions{})
-			if err != nil {
+			if err != nil && !errors.IsNotFound(err) {
 				klog.ErrorS(err, "Failed to get podUnavailableBudget form etcd", "pub", klog.KObj(pub))
 				return err
 			}
--- a/pkg/control/pubcontrol/pub_control_utils.go
+++ b/pkg/control/pubcontrol/pub_control_utils.go
@@ -104,8 +104,8 @@ func PodUnavailableBudgetValidatePod(pod *corev1.Pod, operation policyv1alpha1.P
 				PodUnavailableBudgets(pub.Namespace).Get(context.TODO(), pub.Name, metav1.GetOptions{})
 			if err != nil {
 				if errors.IsNotFound(err) {
-                                    return nil
-			        }
+					return nil
+				}
 				klog.ErrorS(err, "Failed to get podUnavailableBudget form etcd", "pub", klog.KObj(pub))
 				return err
 			}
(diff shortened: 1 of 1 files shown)


pkg/control/pubcontrol/pub_control_utils.go
```diff
@@ -102,7 +102,7 @@ func PodUnavailableBudgetValidatePod(pod *corev1.Pod, operation policyv1alpha1.P
 		if refresh {
 			pubClone, err = kubeClient.GetGenericClient().KruiseClient.PolicyV1alpha1().
 				PodUnavailableBudgets(pub.Namespace).Get(context.TODO(), pub.Name, metav1.GetOptions{})
-			if err != nil {
+			if err != nil && !errors.IsNotFound(err) {
```
furykerry, line 105: if errors.IsNotFound(err), code will goes to the else clause, can the code just return so as to avoid unnecessary and potentially problematic code?

opencmit2: Do you mean to return nil directly?

Spground: We also should check error returned in pub status updating,  just ingore error if NotFound error retured.  https://github.com/openkruise/kruise/blob/142458151be55a538c78c80c6c791cfe9c1fdc10/pkg/control/pubcontrol/pub_control_utils.go#L165

opencmit2: @Spground  
Adding this feels a bit redundant since PUB exits without being discovered and does not continue running kclient.Status().Update

If you need to add it, is the following suitable?
```go
                err = kclient.Status().Update(context.TODO(), pubClone)
                costOfUpdate += time.Since(start)
                if err == nil {
                        if err = util.GlobalCache.Add(pubClone); err != nil {
                                klog.ErrorS(err, "Failed to add cache for podUnavailableBudget", "pub", klog.KObj(pub))
                        }
                        return nil
                } else {
                        if errors.IsNotFound(err) {
                                return nil
                        }
                }
```

Spground: > @Spground Adding this feels a bit redundant since PUB exits without being discovered and does not continue running kclient.Status().Update
> 
> If you need to add it, is the following suitable?
> 
> ```go
>                 err = kclient.Status().Update(context.TODO(), pubClone)
>                 costOfUpdate += time.Since(start)
>                 if err == nil {
>                         if err = util.GlobalCache.Add(pubClone); err != nil {
>                                 klog.ErrorS(err, "Failed to add cache for podUnavailableBudget", "pub", klog.KObj(pub))
>                         }
>                         return nil
>                 } else {
>                         if errors.IsNotFound(err) {
>                                 return nil
>                         }
>                 }
> ```

It's not exactly.  PUB can be deleted anytime, e.g,  before you UPDATE Status and after GET pub.

opencmit2: @Spground  ok,IsNotFound has been added to both locations. Do I still need get informerCached to make IsNotFound judgment?

furykerry (commented):
/lgtm

Issue #1627, opened 2024-05-24
[feature request] Make CloneSet and StatefulSet Support the generateName setting


iambocai:
**What would you like to be added**:

To make CloneSet and StatefulSet support the GenerateName setting.

**Why is this needed**:

Currently, CloneSet rigidly uses `{workloadName}-{randomID}` when generating `pod.Name`. There is a risk here as workload names can be up to 253 bytes long, while pod names are limited to 63 bytes. Therefore, if the user-defined workload name exceeds 57 bytes, the resulting pod name will exceed the limit and become unusable. The general solution to this problem is to specify `GenerateName`, but our current implementation of CloneSet does not support this feature. This leaves users facing such issues with no option but to shorten the workload name, which compromises readability and is not a good solution from a management standpoint. Therefore, I hope that support for the `GenerateName` setting can be incorporated.

https://github.com/openkruise/kruise/blob/a836e90578067d0c44372bc3be116cc3d06babed/pkg/controller/cloneset/core/cloneset_core.go#L98

The requirements for StatefulSet would be similar.

Radeity:
Hi @furykerry , I can work on this issue, can you assign me?

Radeity:
Hi @iambocai , I've tested this issue, actually, the limit is not pod name, but the length of `label name` & `label value`. 

For this issue, you can set feature gate [`CloneSetShortHash`](https://openkruise.io/zh/docs/installation#%E5%8F%AF%E9%80%89-feature-gate)  to `true`, and then you can use longer cs name. 

**Related code logic:**
https://github.com/openkruise/kruise/blob/89d9558dd511a96c7ed1a01ca7fdfe2eec5c3324/pkg/controller/cloneset/utils/cloneset_utils.go#L71-L79

Thus, it seems like no need to make any changes. cc @furykerry

furykerry:
@Radeity although short cloneset hash can be helpful to break the name length limit, it is still beneficial to support generateName field which is a more common way to handle this problem

Radeity:
Hi @furykerry , thanks for your reply. If I understand correctly, in this feat, we should support to use CloneSet/StatefulSet's `GenerateName` as the prefix of their children's pod name, and keep suffix of pod name unchanged (like availableID in CloneSet, ordinal of StatefulSet)?

furykerry:
yes, just replace the prefix with generateName if available

PR #1635, opened 2024-05-31
Make CloneSet and StatefulSet Support the generateName setting

Radeity:
- Support `generateName` setting for CloneSet and StatefulSet.
- Use `generateName` as prefix of the name of owned pod/pvc, rather using `name` like before. And if user doesn't declare `generateName`, `generateName` will be set the same value of `name` in webhook.

- This PR closes https://github.com/openkruise/kruise/issues/1627
- Manually test locally (kind cluster).
- Modify some UT, set different `name` and `generateName` to verify pod/pvc using `generateName` can work well during doReconcile (like cloneset_controller_test:TestReconcile, stateful_set_utils_test.go:newStatefulSet).

- TODO: update doc

### DIFF
--- a/pkg/controller/cloneset/core/cloneset_core.go
+++ b/pkg/controller/cloneset/core/cloneset_core.go
@@ -95,7 +95,7 @@ func (c *commonControl) newVersionedPods(cs *appsv1alpha1.CloneSet, revision str
 		}
 		clonesetutils.WriteRevisionHash(pod, revision)
 
-		pod.Name = fmt.Sprintf("%s-%s", cs.Name, id)
+		pod.Name = fmt.Sprintf("%s-%s", cs.GenerateName, id)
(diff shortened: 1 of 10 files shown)


pkg/controller/cloneset/core/cloneset_core.go
```diff
@@ -95,7 +95,7 @@ func (c *commonControl) newVersionedPods(cs *appsv1alpha1.CloneSet, revision str
 		}
 		clonesetutils.WriteRevisionHash(pod, revision)
 
-		pod.Name = fmt.Sprintf("%s-%s", cs.Name, id)
+		pod.Name = fmt.Sprintf("%s-%s", cs.GenerateName, id)
```
furykerry, line 98: shall we leave pod name empty and let the apiserver assign unique names using generateName?

Issue #1626, opened 2024-05-21
[BUG] CRR is blocking, if recreate a pod which is already during recreate


ColdsteelRail:
**What happened**:
Apply two CRRs to recreate the same pod, there is probability to happen: there is only one CRR reaches ```Completed```, but the other one stay in ```Pending``` unless event ( label the CRR) to trigger reconcile

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
0. prepare pod
```
$ k get pod --selector="app"=sample
NAME       READY   STATUS    RESTARTS   AGE
sample-0   1/1     Running   0          3s
sample-1   1/1     Running   0          2s
```
1. apply two CRRs to recreate pod same named ```sample-0```
```
$ k apply -f crrs.yaml
```
```yaml
$ cat crrs.yaml
 ---
apiVersion: apps.kruise.io/v1alpha1
kind: ContainerRecreateRequest
metadata:
  namespace: default
  name: crr # crr to recreate sample-0
spec:
  podName: sample-0
  containers:      
  - name: nginx
 ---
apiVersion: apps.kruise.io/v1alpha1
kind: ContainerRecreateRequest
metadata:
  namespace: default
  name: crr2 # crr2 also to recreate sample-0
spec:
  podName: sample-0
  containers:       
  - name: nginx
```

2. wait for seconds, crr is completed, but crr2 is in ```Pending``` permanently
```
$ k get crr
NAME   PHASE       POD        NODE           AGE
crr    Pending     sample-0   kind-worker2   36s.  # stay in Pending unless label to trigger reconcile
crr2   Completed   sample-0   kind-worker2   38s

$ k get pod --selector="app"=sample
NAME       READY   STATUS    RESTARTS      AGE
sample-0   1/1     Running   1 (37s ago)   85s
sample-1   1/1     Running   0             84s
```
**Anything else we need to know?**:

**Environment**:
- Kruise version: ```1.6.3```
- Kubernetes version (use `kubectl version`): ```1.27```
- Install details (e.g. helm install args): ```helm install kruise openkruise/kruise```
- Others:

BH4AWS:
I am reproducing the issue and working on a solution. Please give me more time.

chengjoey:
@BH4AWS are you still working for this? feel free to close my pr

BH4AWS:
@chengjoey  yes, i am still working. Please give me a little more time.

PR #1653, opened 2024-06-28
fix crr is stuck when multi crrs specify same pod together

chengjoey:
fix crr is stuck when multi crrs specify same pod together #1626

when multi crrs specify same pod at the same time, only one will be enqueue
https://github.com/openkruise/kruise/blob/8ae13b1b818f4b46f9317a0ce8c806bcfe4fdc00/pkg/daemon/containerrecreate/crr_daemon_controller.go#L160-L162

The original logic may be designed this way intentionally, but would it be better to use `${namespace}/${crdName}` as the key?


chengjoey:
Also, is it possible to modify the webhook to directly reject requests when there is a running crr processing the same pod?

zmberg:
> Also, is it possible to modify the webhook to directly reject requests when there is a running crr processing the same pod?

@chengjoey  agree.

PR #1655, opened 2024-07-03
add support for external certs

### LINKED ISSUE #509: [feature request] cert-manager support

The webhooks should support a helm value for creating cert-manager certificate / ca requests so that cert-manager secures the webhooks rather then generating the certs in the container.

**Why is this needed**:

This centralizes all the certificate management code in one place that can be much more easily monitored and secured. It is difficult to manage lots of different webhooks in the cluster each with their own certificate infrastructure.

**Examples**
There are examples of how this works here: https://cert-manager.io/docs/concepts/ca-injector/

Kuromesi:
Add new feature to read and validate certs stored in secret in order to support external certs generated by cert manager and etc.

Fixes #509.


pkg/webhook/util/writer/external.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	return &externalCertWriter{ExternalCertWriterOptions: &ops}, nil
+}
+
+// EnsureCert read and validate certs from k8s secret.
+func (s *externalCertWriter) EnsureCert(dnsName string) (*generator.Artifacts, bool, error) {
+	// Create or refresh the certs based on clientConfig
+	s.dnsName = dnsName
+	certs, err := s.read()
+	if err != nil {
```
furykerry, line 84: is it possible the ca read from the secret is the same as before ? it necessary to update webhookconfiguration again ?

pkg/webhook/util/writer/external.go
```diff
@@ -0,0 +1,127 @@
...
+func NewExternalCertWriter(ops ExternalCertWriterOptions) (CertWriter, error) {
+	err := ops.validate()
+	if err != nil {
+		return nil, err
+	}
+	return &externalCertWriter{ExternalCertWriterOptions: &ops}, nil
+}
+
+// EnsureCert read and validate certs from k8s secret.
+func (s *externalCertWriter) EnsureCert(dnsName string) (*generator.Artifacts, bool, error) {
```
furykerry, line 80: plz add ut for this func

pkg/webhook/util/writer/external.go
Kuromesi: add logic to check if the certs are same as before, however, the `changed` bool value is not used in webhook controller, and the webhook controller will use `reflect.DeepEqual()` to see if the webhook configuration need to be updated.

pkg/webhook/util/configuration/configuration.go
```diff
@@ -58,65 +59,102 @@ func Ensure(kubeClient clientset.Interface, handlers map[string]types.HandlerGet
...
+			if err != nil {
+				return err
+			}
+			if _, ok := handlers[path]; !ok {
+				klog.Warningf("Ignore webhook for %s in configuration", path)
+				continue
+			}
+			if !bytes.Equal(wh.ClientConfig.CABundle, caBundle) {
+				return fmt.Errorf("caBundle of ValidatingWebhookConfiguration %s does not match the external caBundle", mutatingWebhookConfigurationName)
 			}
```
furykerry, line 91: just return at the end of the if clause, and remove the else to make indentation smaller

pkg/webhook/util/crd/crd.go
```diff
@@ -49,41 +52,61 @@ func Ensure(client apiextensionsclientset.Interface, lister apiextensionslisters
...
+				return fmt.Errorf("bad conversion configuration of CRD %s", crd.Name)
+			}
 
-	for _, crd := range crdList {
-		if len(crd.Spec.Versions) == 0 || crd.Spec.Conversion == nil || crd.Spec.Conversion.Strategy != apiextensionsv1.WebhookConverter {
-			continue
+			if !bytes.Equal(crd.Spec.Conversion.Webhook.ClientConfig.CABundle, caBundle) {
+				return fmt.Errorf("caBundle of CRD %s does not match external caBundle", crd.Name)
+			}
+		}
```
furykerry, line 71: just return in the end of if clause and remove the else to make the indentation smaller

PR #1677, opened 2024-07-30
read certs from secrets to support external certs

Kuromesi:
A commit is submitted previously to support certs generated externally, kruise will only read cabunle generated and updated by cert manager in webhook to do ready check. However, it is crucial to get both ca certs and private key to fire up a webhook server, since, the old logic fails to work.

This commit make changes listed behind:
- update external writer to read certs and keys from a secret in kruise namespace without any validation checks (since kruise can do nothing when the certs are not valid)
- when ensure webhook and crd configuration, check whether the cabundle is equal to the one kruise read from secret.

NONE

When users want to use this feature, they should create certificate like below to bind the certs into a secret in kruise namespace, so kruise can read from secrets.

```yaml
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: kruise-webhook
  namespace: kruise-system
spec:
  secretName: kruise-webhook-certs
  dnsNames:
  - kruise-webhook-service.kruise-system.svc
  - localhost
  issuerRef:
    name: selfsigned-kruise
    kind: Issuer
```


pkg/webhook/util/writer/external.go
```diff
@@ -62,12 +78,15 @@ func NewExternalCertWriter(ops ExternalCertWriterOptions) (CertWriter, error) {
 	return &externalCertWriter{ExternalCertWriterOptions: &ops}, nil
 }
 
+// EnsureCert read and validate certs from k8s secret.
 func (s *externalCertWriter) EnsureCert(dnsName string) (*generator.Artifacts, bool, error) {
-	// Read certs from mutating webhook configuration generated by external
+	// Read certs from secrets generated by external
+	s.dnsName = dnsName
```
furykerry, line 84: is dnsName ever used?

PR #1678, opened 2024-07-31
Fix nil panic in cloneset validating webhook

Spground:

Fix potential nil panic in CloneSet validating webhook when Pod's controller owner ref is nil.


furykerry (commented):
/lgtm
/approve

PR #1688, opened 2024-08-02
fix kind install bug

### LINKED ISSUE #1686: [BUG] local kind test tools are invalid
**What happened**:

I want to use command `make create-cluster` to set up a Kind cluster locally for testing purposes. By default, the Makefile builds Kind results in an error since there is no code in the `tools/src/kind` directory. To resolve this, I have to manually download the Kind executable by running the command `go install sigs.k8s.io/kind@v0.23.0`, and then copy it into the `tools/bin` directory to proceed with further operations.

when I type `make create-cluster` command, the Makefile should download the Kind executable to `tools/bin` automatically instead of building the ghost code.

1. clone a clean kruise repository
2. type `make create-cluster`

**Environment**:
- Kruise version: 1.6.3
- Kubernetes version (use `kubectl version`): Any
- Install details (e.g. helm install args): Nothing
- Others:

ABNER-1:
auto install kind when `make create-cluster`

fixes #1686

### DIFF
--- a/tools/src/kind/pin.go
+++ b/tools/src/kind/pin.go
@@ -16,3 +16,5 @@
 // +build pin
 
 package ignore
+
+import _ "sigs.k8s.io/kind"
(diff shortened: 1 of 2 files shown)


furykerry (commented):
/approve

PR #1714, opened 2024-08-28
Auto resize pvc when only expand volumeClaimTemplates in asts

### LINKED ISSUE #1699: [feature request]When only resize VolumeClaimTemplates, asts should also resize pvcs

The StatefulSet should resize Persistent Volume Claims (PVCs) whenever only the VolumeClaimTemplates are resized.

**Why is this needed**:

This feature is crucial for users of stateful applications in production environments.
When expanding storage is required, users should be able to do so without the need for configuration or image alterations, as these changes can introduce greater risks.

ABNER-1:
Auto resize pvc when only expand volumeClaimTemplates in asts

fixes #1699

### DIFF
--- a/pkg/controller/statefulset/stateful_set_control.go
+++ b/pkg/controller/statefulset/stateful_set_control.go
@@ -619,6 +619,8 @@ func (ssc *defaultStatefulSetControl) rollingUpdateStatefulsetPods(
 				klog.V(4).ErrorS(err, "StatefulSet check owned pvcs unready",
 					"statefulSet", klog.KObj(set), "pod", klog.KObj(replicas[target]))
 			}
+			klog.V(4).InfoS("owned pvcs unready, added in unavailable Pods",
+				"statefulSet", klog.KObj(set), "pod", klog.KObj(replicas[target]))
 			unavailablePods.Insert(replicas[target].Name)
 		}
 	}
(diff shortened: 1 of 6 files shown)


pkg/controller/statefulset/stateful_set_control.go
```diff
@@ -646,9 +653,7 @@ func (ssc *defaultStatefulSetControl) rollingUpdateStatefulsetPods(
 		if utilfeature.DefaultFeatureGate.Enabled(features.StatefulSetAutoResizePVCGate) &&
 			set.Spec.VolumeClaimUpdateStrategy.Type == appsv1beta1.OnPodRollingUpdateVolumeClaimUpdateStrategyType {
 			// resize pvc if necessary and wait for resize completed
-			if match, err := ssc.podControl.IsClaimsCompatible(set, replicas[target]); err != nil {
-				return status, err
-			} else if !match {
+			if !pvcMatched {
```
furykerry, line 656: in L 646,  shall we also log the volumes to be resizing? otherwise we can not tell whether the rolling is blocked by unavailable pods or resizing volumes.

ABNER-1: Now resizing volumes lead to pods becoming unavailable. To address this, I have added logs at lines 621 and 622. 

In the future, when we add parallelism in the volumeUpdateStrategy, it will enable users to control the maximum number of volumes being updated simultaneously, thereby separately accounting for these two causes of unavailability during updates.

furykerry (approved):

PR #1720, opened 2024-09-02
Adaptive schedule strategy for UnitedDeployment

### LINKED ISSUE #1673: [feature request] UnitedDeployment support reschedule pod to other subset if current subset lacks resources
Workloadspread currently support reschedule pod to other subset if current subset lacks resources, but UnitedDeployment does not. UnitedDeployment should match Workloadspread as far as possbile.

https://openkruise.io/docs/user-manuals/workloadspread/#schedule-strategy

Add schedule-strategy in UnitedDeployment, and support Fixed(current behavior) and Adaptive(missing). When schedule-strategy is adaptive, and the pod in one subset if unschedulable, UnitedDeployment should try next the subset.

**Why is this needed**:
Enable UnitedDeployment to support use cases such as deployment pods to fixed ECS nodes as many as possible, and if ECS nodes is not enough for new Pods, deploy pods to virtual nodes as serverless pods.

AiRanthem:
Added an adaptive scheduling strategy to UnitedDeployment. During scaling up, if a subset causes some Pods to be unschedulable for certain reasons, the unschedulable Pods will be rescheduled to other partitions. During scaling down, if elastic allocation is used (i.e., the subset is configured with min/max), each partition will retain the ready Pods as much as possible without exceeding the maximum capacity, rather than strictly scaling down in reverse order of the Subset list.

fixes #1673

Use the yaml below to create a UD with subset-b unschedulable.

```yaml
kind: UnitedDeployment
metadata:
  name: sample-ud
spec:
  replicas: 5
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: sample
  template:
    deploymentTemplate:
      metadata:
        labels:
          app: sample
      spec:
        selector:
          matchLabels:
            app: sample
        template:
          metadata:
            labels:
              app: sample
          spec:
            terminationGracePeriodSeconds: 0
            containers:
              - name: nginx
                image: curlimages/curl:8.8.0
                command: ["/bin/sleep", "infinity"]
  topology:
    scheduleStrategy:
      type: Adaptive
      adaptive:
        rescheduleCriticalSeconds: 10
        unschedulableLastSeconds: 20

    subsets:
      - name: subset-a
        maxReplicas: 2
        nodeSelectorTerm:
          matchExpressions:
            - key: kubernetes.io/hostname
              operator: In
              values:
                - ci-testing-worker
      - name: subset-b
        maxReplicas: 2
        nodeSelectorTerm:
          matchExpressions:
            - key: kubernetes.io/hostname
              operator: In
              values:
                - not-exist
      - name: subset-c
        nodeSelectorTerm:
          matchExpressions:
            - key: kubernetes.io/hostname
              operator: In
              values:
                - ci-testing-worker3
```
1. when created, two pods in subset-b will stay pending
2. after 10s, the two pending pods will be rescheduled to subset-c
3. scale up immediately, new pods will be created in subset-c instead of subset-b (even not full)
4. wait 20s, when subset-b is recovered, scale up again, 2 pods will be scheduled into subset-b again (and still pending)
5. whenever you scale down:
  subset-c -> subset-b -> subset-a

1. adapter.go: GetReplicaDetails returns pods in the subset
2. xxx_adapter.go: return pods implementation ⬆️
3. allocator.go: about safeReplica
4. pod_condition_utils.go: extract PodUnscheduledTimeout function from workloadwpread
5. reschedule.go: PodUnscheduledTimeout function extracted
6. subset.go: add some field to Subset object to carry related information
7. subset_control.go: store subset pods to Subset object
8. uniteddeployment_controller.go
   1. add requeue feature to check failed pods
   2. subset unschedulable status management
9. uniteddeployment_types.go: API change
10. uniteddeployment_update.go: sync unschedulable to CR


pkg/controller/uniteddeployment/allocator_test.go
```diff
@@ -307,6 +308,92 @@ func TestCapacityAllocator(t *testing.T) {
...
+				"subset-2": {
+					Status: SubsetStatus{},
+				},
+			}
+	}
+	cases := []struct {
+		totalReplicas, minReplicas, maxReplicas, pendingPods int32
+		subset1Replicas, subset2Replicas                     int32
+	}{
+		{
```
furykerry, line 354: consider add a human readable name for each cases

pkg/util/expectations/resource_version_expectation.go
```diff
@@ -36,6 +36,13 @@ func NewResourceVersionExpectation() ResourceVersionExpectation {
 	return &realResourceVersionExpectation{objectVersions: make(map[types.UID]*objectCacheVersions, 100)}
 }
 
+func NewReallyNewerResourceVersionExpectation() ResourceVersionExpectation {
```
furykerry, line 39: plz add ut for the new expectation implementation

AiRanthem: implementation removed

pkg/controller/uniteddeployment/uniteddeployment_update.go
```diff
@@ -79,6 +80,17 @@ func (r *ReconcileUnitedDeployment) manageSubsets(ud *appsv1alpha1.UnitedDeploym
 	if updateErr == nil {
 		SetUnitedDeploymentCondition(newStatus, NewUnitedDeploymentCondition(appsv1alpha1.SubsetUpdated, corev1.ConditionTrue, "", ""))
 	} else {
+		// If using an Adaptive scheduling strategy, when the subset is scaled out leading to the creation of new Pods,
+		// future potential scheduling failures need to be checked for rescheduling.
+		var newPodCreated = false
```
furykerry, line 85: is it really necessary to trigger the requeue here given that manageUnschedulableStatusForExistingSubset had already push some duration time.

pkg/controller/uniteddeployment/uniteddeployment_update.go
```diff
@@ -132,6 +144,11 @@ func (r *ReconcileUnitedDeployment) manageSubsetProvision(ud *appsv1alpha1.Unite
 			return nil
 		})
 		if createdErr == nil {
+			// When a new subset is created, regardless of whether it contains newly created Pods,
```
furykerry, line 147: is it really necessary to trigger the requeue here given that manageUnschedulableStatusForExistingSubset had already push some duration time.

pkg/controller/uniteddeployment/uniteddeployment_controller.go
```diff
@@ -348,10 +449,14 @@ func (r *ReconcileUnitedDeployment) classifySubsetBySubsetName(ud *appsv1alpha1.
 	return mapping
 }
 
-func (r *ReconcileUnitedDeployment) updateStatus(instance *appsv1alpha1.UnitedDeployment, newStatus, oldStatus *appsv1alpha1.UnitedDeploymentStatus, nameToSubset *map[string]*Subset, nextReplicas, nextPartition *map[string]int32, currentRevision, updatedRevision *appsv1.ControllerRevision, collisionCount int32, control ControlInterface) (reconcile.Result, error) {
+func (r *ReconcileUnitedDeployment) updateStatus(instance *appsv1alpha1.UnitedDeployment, newStatus, oldStatus *appsv1alpha1.UnitedDeploymentStatus, nameToSubset *map[string]*Subset, nextReplicas, nextPartition *map[string]int32, currentRevision, updatedRevision *appsv1.ControllerRevision, collisionCount int32, control ControlInterface) error {
```
furykerry, line 452: there is too many func params in the updateStatus, consider move the calculateStatus to Reconcile, and make updateStatus accept only the instance, new and old status.

pkg/controller/uniteddeployment/uniteddeployment_update.go
AiRanthem: manageUnscheddulableStatusForExistingSubset pushes some duration time only for existing subsets, new subsets is not processed

pkg/controller/uniteddeployment/uniteddeployment_update.go
AiRanthem: 1. if a duration time is pushed in manageUnschedulableStatusForExistingSubset, this will overwrite it
2. manageUnschedulableStatusForExistingSubset only pushes some duration only when an unschedulable subset is dectected or not recovered, which means for normal subsets, where scaling up happens most often, no duration will be pushed.

PR #1733, opened 2024-09-09
Feature/revision index

shiyan2016:
add revision fieldindex to optimize list revisions operation

### DIFF
--- a/pkg/controller/sidecarset/sidecarset_controller_test.go
+++ b/pkg/controller/sidecarset/sidecarset_controller_test.go
@@ -9,6 +9,7 @@ import (
 	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
 	"github.com/openkruise/kruise/pkg/control/sidecarcontrol"
 
+	"github.com/openkruise/kruise/pkg/util/fieldindex"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -17,6 +18,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/kubernetes/pkg/apis/apps"
 	utilpointer "k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
@@ -137,6 +139,7 @@ func init() {
 	utilruntime.Must(appsv1.AddToScheme(scheme))
 	utilruntime.Must(appsv1alpha1.AddToScheme(scheme))
 	utilruntime.Must(corev1.AddToScheme(scheme))
+	utilruntime.Must(apps.AddToScheme(scheme))
 }
 
 func getLatestPod(client client.Client, pod *corev1.Pod) (*corev1.Pod, error) {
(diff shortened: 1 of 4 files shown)


pkg/util/history/controller_history.go
```diff
@@ -44,16 +47,14 @@ type realHistory struct {
 func (rh *realHistory) ListControllerRevisions(parent metav1.Object, selector labels.Selector) ([]*apps.ControllerRevision, error) {
 	// List all revisions in the namespace that match the selector
 	revisions := apps.ControllerRevisionList{}
-	err := rh.List(context.TODO(), &revisions, &client.ListOptions{Namespace: parent.GetNamespace(), LabelSelector: selector})
-	if err != nil {
-		return nil, err
+	opts := &client.ListOptions{
```
furykerry, line 50: consider add DisableDeepCopy option

furykerry:
plz fix the ut

pkg/util/history/controller_history_test.go
```diff
@@ -69,7 +71,13 @@ func TestRevisionHistory(t *testing.T) {
 		t.Fatalf("Failed to new controller revision: %v", err)
 	}
 
-	fakeClient := fake.NewClientBuilder().Build()
+	fakeClient := fake.NewClientBuilder().WithIndex(&apps.ControllerRevision{}, fieldindex.IndexNameForOwnerRefUID, func(obj client.Object) []string {
+		var owners []string
```
ABNER-1, line 75: There are so many errors in other uts and can be fixed by using client with index.
I think we should wrap this code with a customized function and replace `fake.NewClientBuilder()`.

pkg/util/history/controller_history.go
ABNER-1: When use DisableDeepCopy, kruise will panic in e2e because some update codes in cloneset VCTHashEqual handler. 
We should fix it. @shiyan2016 


pkg/controller/cloneset/cloneset_controller.go:453
```go
lastEqualRevision := equalRevisions[equalCount-1]
if !VCTHashEqual(lastEqualRevision, updateRevision) {
	klog.InfoS("Revision vct hash will be updated", "revisionName", lastEqualRevision.Name, "lastRevisionVCTHash", lastEqualRevision.Annotations[volumeclaimtemplate.HashAnnotation], "updateRevisionVCTHash", updateRevision.Annotations[volumeclaimtemplate.HashAnnotation])
	lastEqualRevision.Annotations[volumeclaimtemplate.HashAnnotation] = updateRevision.Annotations[volumeclaimtemplate.HashAnnotation]
}
// if the equivalent revision is not immediately prior we will roll back by incrementing the
// Revision of the equivalent revision
updateRevision, err = r.controllerHistory.UpdateControllerRevision(equalRevisions[equalCount-1], updateRevision.Revision)
```

pkg/controller/sidecarset/sidecarset_controller_test.go
```diff
@@ -137,6 +139,7 @@ func init() {
 	utilruntime.Must(appsv1.AddToScheme(scheme))
 	utilruntime.Must(appsv1alpha1.AddToScheme(scheme))
 	utilruntime.Must(corev1.AddToScheme(scheme))
+	utilruntime.Must(apps.AddToScheme(scheme))
```
furykerry, line 142: there is no need to add unversioned scheme here, L139 already do the job

ABNER-1 (changes requested):

pkg/controller/sidecarset/sidecarset_controller_test.go
```diff
@@ -219,6 +229,13 @@ func testUpdateWhenSidecarSetPaused(t *testing.T, sidecarSetInput *appsv1alpha1.
 
 	fakeClient := fake.NewClientBuilder().WithScheme(scheme).
 		WithObjects(sidecarSetInput, podInput).
+		WithIndex(&apps.ControllerRevision{}, fieldindex.IndexNameForOwnerRefUID, func(obj client.Object) []string {
```
furykerry, line 232: use appsv1.ControllerRevision{}

Issue #1735, opened 2024-09-10
[BUG] Sidecarset sourceContainerNameFrom 并不支持name,namespace


magicsong:
**What happened**:
https://github.com/pigletfly/kruise/blob/6a62320848f8e10e037c747eeef6b99552af7fa9/pkg/control/sidecarcontrol/util.go#L434
**What you expected to happen**:
这里hardcode枚举了，仅仅支持labels和annoation，sidecar无法获取自己的name,namepsace
**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kruise version: 最新代码
- Kubernetes version (use `kubectl version`):
- Install details (e.g. helm install args):
- Others:

magicsong:
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/pods/helpers.go 更新一下代码是不是就可以了

ABNER-1:
看起来是的，你愿意同步一下这段代码贡献一下吗？
如果可以的话可以将这几个case 加到对应的 ut 中。

furykerry:
it is strange to use pod name and namespace as part of the container name , are there any reason of this usage? @magicsong

magicsong:
> it is strange to use pod name and namespace as part of the container name , are there any reason of this usage? @magicsong

"sourceContainerNameFrom should be a mechanism similar to the downward API, used to mount pod metadata to environment variables, rather than its literal name meaning."

furykerry:
> "sourceContainerNameFrom should be a mechanism similar to the downward API, used to mount pod metadata to environment variables, rather than its literal name meaning."
so what field do you need the complete downward api ? it make no sense to make sourceContainerNameFrom refer to the pod name. Do you mean other fields? TransferEnv can refer to the environment name of another container, this environment can use downward api.

magicsong:
> > "sourceContainerNameFrom should be a mechanism similar to the downward API, used to mount pod metadata to environment variables, rather than its literal name meaning."
> 
> so what field do you need the complete downward api ? it make no sense to make sourceContainerNameFrom refer to the pod name. Do you mean other fields? TransferEnv can refer to the environment name of another container, this environment can use downward api.

The sidecar needs to have the ability to know its own name and namespace without interfering with the main container. Not all main containers have the environment variables PodName and PodNamespace, and the names are fixed.

Additionally, the OpenKruise documentation mentions support for [metadata.name], but it is not actually supported.

[...]

PR #1739, opened 2024-09-12
fix: support more fieldpath

magicsong:
simply use kube code instead of copying it to support more field path in "transferEnv" of sidecarset

fixes: #1735


furykerry:
per discussion in the issue https://github.com/openkruise/kruise/issues/1735, the patch is not necessary

PR #1740, opened 2024-09-12
fix a bug about workloadspread webhook validation on patches

### LINKED ISSUE #1738: [BUG] Can not create a WorkloadSpread with subset patches if the target CloneSet, which includes a volume template, already exists.
**What happened**:

When creating a WorkloadSpread, the request will be intercepted by the Webhook and result in failure if the following conditions are met:
1. A subset of the WorkloadSpread contains a patch.
2. The WorkloadSpread's `targetRef` is an existing CloneSet.
3. This CloneSet has a volume template, and a container mounts that volume.

Created successfully.

1. Create a CloneSet named test-cs with a volume template
2. Create a WorkloadSpread whose targetRef is test-cs, and patches anything in a subset
3. An error related to volume will occur like:
```text
Error from server: error when creating "workloadspread.yaml": admission webhook "vworkloadspread.kb.io" denied the request: [spec.subsets[0].patch.containers[0].volumeMounts[0].name: Not found: "vol-1"]
```

This bug is caused by the validation of patches in the Webhook. During this validation, the Webhook reads the template of the existing target workload, applies the subset's patch to it as a real corev1.Pod object, and checks if the patched Pod is valid. In this bug, the patched Pod's `volumes` field does not contain the volumes related to the CloneSet's volume template, leading the Webhook to mistakenly believe that the patched Pod would mount a non-existent volume. As a result, it incorrectly deems the patch invalid, ultimately rejecting the creation of the WorkloadSpread.

**Environment**:
- Kruise version: any
- Kubernetes version (use `kubectl version`): any
- Others:

AiRanthem:

patches volume claim templates into pods before ValidatePodSpec in workloadspread patch validation

fixes #1738

1. create a CloneSet with volume claim template
2. create a workloadspread references it


pkg/webhook/workloadspread/validating/workloadspread_validation_test.go
```diff
@@ -943,3 +949,129 @@ func TestValidateWorkloadSpreadConflict(t *testing.T) {
 		})
 	}
 }
+
+func Test_validateWorkloadSpreadSubsets(t *testing.T) {
+	clonesetYaml := `
```
furykerry, line 954: consider constructing the workload with struct literal instead of yaml

pkg/webhook/workloadspread/validating/workloadspread_validation_test.go
```diff
@@ -943,3 +949,129 @@ func TestValidateWorkloadSpreadConflict(t *testing.T) {
...
+				{
+					Name: "test",
+					Patch: runtime.RawExtension{
+						Raw: patch,
+					},
+				},
+			},
+		},
+	}
+	errList := field.ErrorList{}
```
furykerry, line 1068: plz add cases that we expect validateWorkloadSpreadSubsets will fail, and organize all cases using sub cases, you can refer to TestValidateWorkloadSpreadCreate

pkg/webhook/workloadspread/validating/workloadspread_validation_test.go
```diff
@@ -943,3 +947,170 @@ func TestValidateWorkloadSpreadConflict(t *testing.T) {
...
+			workload: badSts,
+			testFunc: func(errList field.ErrorList) {
+				if len(errList) != 1 {
+					t.Fatalf("expected 1 error, got %d, errList = %+v", len(errList), errList)
+				}
+			},
+		},
+	}
+	for _, tc := range testCases {
+		fmt.Printf("test case: \"%s\"\n", tc.name)
```
furykerry, line 1111: use t.run() to execute the subcases

furykerry (commented):
/lgtm

PR #1771, opened 2024-10-02
bugfix(cloneset-update): correct the wrong unavailableCount logic.

zk-123:
In my opinion, the correct approach should be` !IsPodAvailable()`, rather than `IsPodAvailable`.


zk-123:
@FillZpp

Issue #1639, opened 2024-06-03
[BUG] Override namespace config after deploying


Radeity:
**What happened**:
After `make deploy`, `config/manager/kustomization.yaml` will be overwritten and we will miss namespace config. This will cause two problems:
- New file in git working area. (config/manager/kustomization.yaml)
- If we don't recover the change in this file, and run `make deploy` again, wrong namespace `system` is created, rather `kruise-system`.

**What you expected to happen**: 
- Do not overwrite this file. Maybe remove line 90?
https://github.com/openkruise/kruise/blob/0e69ed4bec56d7956457802225a45b962e7869be/Makefile#L87-L91

**How to reproduce it (as minimally and precisely as possible)**:
- Run `make deploy` twice.

**Anything else we need to know?**:

**Environment**:
- Kruise version:
- Kubernetes version (use `kubectl version`):
- Install details (e.g. helm install args):
- Others:

PR #1772, opened 2024-10-02
fix: override namespace config after deploying

hantmac:
Fix https://github.com/openkruise/kruise/issues/1639


furykerry (approved):
/lgtm
/approve

PR #1810, opened 2024-10-30
Fix the issue where the webhook port does not support IPv6 listening …

lidajia:
…configuration, and use IPv4 compatibility mode to listen to the webhook port

Fix the issue where the webhook port does not support IPv6 listening configuration, and use IPv4 compatibility mode to listen to the webhook port

support ipv6


ABNER-1:
Hello, @lidajia.
Please fix DCO checks by following [step](https://github.com/openkruise/kruise/pull/1810/checks?check_run_id=32267789712).

Issue #1832, opened 2024-11-19
[BUG] A smaller MaxUnavailable will block the sidecarSet from updating Pods.


zmberg:
**What happened**:
```
apiVersion: apps.kruise.io/v1alpha1
kind: SidecarSet
metadata:
  name: test-sidecarset
spec:
  containers:
  - command:
    - sleep
    - 999d
    image: mirrors-ssl.aliyuncs.com/centos:6.8
    imagePullPolicy: IfNotPresent
    name: sidecar1
    podInjectPolicy: BeforeAppContainer
    resources: {}
    shareVolumePolicy:
      type: disabled
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    upgradeStrategy:
      upgradeType: ColdUpgrade
  injectionStrategy: {}
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: nginx
  updateStrategy:
    maxUnavailable: 5%
    partition: 0
    type: RollingUpdate
status:
  collisionCount: 0
  latestRevision: test-sidecarset-7658c89f8d
  matchedPods: 5
  observedGeneration: 3
  readyPods: 5
  updatedPods: 0
```
**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kruise version:
- Kubernetes version (use `kubectl version`):
- Install details (e.g. helm install args):
- Others:

PR #1834, opened 2024-11-20
fix: a smaller maxUnavaila will block the sidecarSet from updating pods

zmberg:

https://github.com/openkruise/kruise/issues/1832


furykerry (approved):
/lgtm
/approve

PR #1873, opened 2025-01-02
feature: range reserveOrdinals for AdvancedStatefulSet

### LINKED ISSUE #1846: [feature request] asts reserveOrdinals support range type

`spec.reserveOrdinals` of asts support to specify ID list to reserve now. I suggest to support range type to avoid filling in too much number, which will make the yaml hard to check and add new IDs when ID is getting more.

For example, when users want to reserve list [1,2,3,4,5,6,7], if supporting range type, it will be [1 -> 7], much easier.

AiRanthem:

support range value like `1-5` to assign a lot of reserveOrdinals in AdvancedStatefulSet, which looks like:

```yaml
kind: StatefulSet
spec:
  # ...
  replicas: 4
  reserveOrdinals:
  - 1
  - 2-5 # <- SEE THIS
```

the example is the same as:

```yaml
kind: StatefulSet
spec:
  # ...
  replicas: 4
  reserveOrdinals:
  - 1
  - 2
  - 3
  - 4
  - 5
```

so you don't have to write a very long reserveOrdinals list.

fixes #1846 

use the example above


pkg/webhook/statefulset/validating/statefulset_validation.go
```diff
@@ -44,15 +43,16 @@ func validatePodManagementPolicy(spec *appsv1beta1.StatefulSetSpec, fldPath *fie
...
 	var allErrs field.ErrorList
-
-	if spec.ReserveOrdinals != nil {
-		orders := sets.NewInt(spec.ReserveOrdinals...)
-		if orders.Len() != len(spec.ReserveOrdinals) {
-			allErrs = append(allErrs, field.Invalid(fldPath.Root(), spec.ReserveOrdinals, "reserveOrdinals contains duplicated items"))
-		}
-		for _, i := range spec.ReserveOrdinals {
-			if i < 0 {
-				allErrs = append(allErrs, field.Invalid(fldPath.Root(), spec.ReserveOrdinals, fmt.Sprintf("reserveOrdinals contains %d which must be order >= 0", i)))
```
furykerry, line 55: plz keep the duplication checking

apis/apps/v1beta1/statefulset_types.go
```diff
@@ -299,6 +305,37 @@ type StatefulSetSpec struct {
 	Ordinals *StatefulSetOrdinals `json:"ordinals,omitempty"`
 }
 
+type ReserveOrdinal []intstr.IntOrString
+
+func (r ReserveOrdinal) GetIntSet() sets.Set[int] {
```
furykerry, line 310: consider moving the function to the utility package, the api package should not contain func other than simple deepcopys

pkg/webhook/statefulset/validating/statefulset_validation.go
```diff
@@ -44,15 +43,16 @@ func validatePodManagementPolicy(spec *appsv1beta1.StatefulSetSpec, fldPath *fie
...
+	if len(spec.ReserveOrdinals) > 0 {
+		matcher := regexp.MustCompile(`^\d+-\d+$`)
+		for i, elem := range spec.ReserveOrdinals {
+			if elem.Type == intstr.String && !matcher.MatchString(elem.StrVal) {
+				allErrs = append(allErrs, field.Invalid(fldPath.Root(), spec.ReserveOrdinals,
```
furykerry, line 50: we should also checking the start and ending string is a valid number, avoid the case like "03-01"

apis/apps/v1beta1/statefulset_types.go
```diff
@@ -299,6 +305,37 @@ type StatefulSetSpec struct {
 	Ordinals *StatefulSetOrdinals `json:"ordinals,omitempty"`
 }
 
+type ReserveOrdinal []intstr.IntOrString
```
zmberg, line 308: Can you tell me if there will be any duplication or order issues with this method? For example:
1. 1, 2, 2-5, 5
2. 2-5, 1, 2 

Will it work properly?

pkg/webhook/statefulset/validating/statefulset_validation.go
zmberg: I also think it should be more strictly judged, e.g. refer to GetIntSet

apis/apps/v1beta1/statefulset_types.go
AiRanthem: ut added

furykerry (commented):
/lgtm

PR #1877, opened 2025-01-09
Feature/Bugfix: This commit includes the following two changes:

AiRanthem:
1. Bugfix: WorkloadSpread cannot patch priorityClassName
2. Feature: Add context logging to WorkloadSpread


pkg/util/workloadspread/workloadspread.go
```diff
@@ -640,14 +645,21 @@ func injectWorkloadSpreadIntoPod(ws *appsv1alpha1.WorkloadSpread, pod *corev1.Po
...
 			return false, err
 		}
 		newPod := &corev1.Pod{}
 		if err = json.Unmarshal(modified, newPod); err != nil {
-			klog.ErrorS(err, "failed to unmarshal to Pod", "pod", modified)
+			logger.Error(err, "failed to unmarshal to Pod", "pod", modified)
 			return false, err
 		}
+		if newPod.Spec.PriorityClassName != pod.Spec.PriorityClassName {
+			// Workloadspread webhook is called after builtin k8s pod webhook, which means the priority
```
furykerry, line 657: consider rephase the comment as follows:

Workloadspread webhook is called after builtin admission plugin, which means the priority is already set before priorityClassName being patched. Mismatched priorityClassName and priority value will be rejected by apiserver. we have to clear priority to avoid the problem, and the builtin admission plugin will be reinvoked after kruise webhook to setting the correct priority value.

furykerry (commented):
consider separate logging and priority mutation logic into two commits

furykerry (approved):
/lgtm
/approve

PR #1881, opened 2025-01-14
CloneSet supports to reuse PVC during upgrade

### LINKED ISSUE #1876: [feature request] support cloneset update policy not to delete pvc during recreate update
support cloneset update policy not to delete pvc during recreate update

**Why is this needed**:
sometimes we need to keep existing pvc during update

FillZpp:
CloneSet supports to reuse PVC during upgrade.

Newly added:
* `CloneSetPVCReuseDuringUpdate` feature-gate that enables CloneSet to reuse PVC during update.
* `apps.kruise.io/keep-pvc-for-deletion` label that can either be used by the above feature, or user can directly set it in pod or pod template, so that CloneSet will keep the PVC to reuse when it deletes the pod.

fixes #1876

### DIFF
--- a/pkg/util/specifieddelete/specified_delete.go
+++ b/pkg/util/specifieddelete/specified_delete.go
@@ -18,13 +18,14 @@ package specifieddelete
 
 import (
 	"context"
-	"fmt"
 
-	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	appsv1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
+	"github.com/openkruise/kruise/pkg/util"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/util/specifieddelete/specified_delete.go
```diff
@@ -18,29 +18,43 @@ package specifieddelete
...
 )
 
 func IsSpecifiedDelete(obj metav1.Object) bool {
 	_, ok := obj.GetLabels()[appsv1alpha1.SpecifiedDeleteKey]
 	return ok
 }
 
-func PatchPodSpecifiedDelete(c client.Client, pod *v1.Pod, value string) (bool, error) {
+func ShouldKeepPVC(obj metav1.Object) bool {
+	return obj.GetLabels()[appsv1alpha1.KeepPVCForDeletionKey] == "true"
```
furykerry, line 37: is it necessary to introduce an extra label ? can we just checking the following logic?
!cs.Spec.ScaleStrategy.DisablePVCReuse && utilfeature.DefaultFeatureGate.Enabled(features.CloneSetPVCReuseDuringUpdate)

ABNER-1:
@FillZpp I think this PR still needs some modifications related to RecreatePodWhenChangeVCTInCloneSetGate. I will submit another PR for adaptation later.

Could you please rebase master?

Issue #1916, opened 2025-02-10
[BUG] About all zero monitoring and insecure protocols


hcwnbs:
When I was using this software, I found some listening ports with all-zero values during the scanning process, such as 9876, 8080, 8090, and 8000. Moreover, the insecure TLS 1.0 and 1.1 protocols, as well as insecure cipher suites (for example, using RSA as the key exchange algorithm and cipher suites containing CBC symmetric cipher algorithms in the TLS protocol) are being used for port 9876. May I ask if there is any plan to address these issues?

![Image](https://github.com/user-attachments/assets/d4d8ac5c-1da6-4e90-8386-9113e101c8ae)

furykerry:
it is related to the controller-runtime webhook setting, a possible fix can be refer to the [issue](https://github.com/kubernetes-sigs/controller-runtime/issues/2020)

RohanMishra315:
Hey @furykerry Could I look into this ?

RohanMishra315:
Got the idea how to solve this, but having problem where should the changes be made , @furykerry could you help me a bit .

furykerry:
the webhook server initialization code is [here](https://github.com/openkruise/kruise/blob/master/main.go#L202)

furykerry:
@hcwnbs actually there are two issues related
 
1. the use of tls 1.0 and 1.1
2. kruise webhook listen on all zero address 

#1 is already fixed in kruise 1.8 since we have upgraded the golang version to 1.22, in which , tls 1.2 is now the default minVersion for both server and client.

hcwnbs:
> [@hcwnbs](https://github.com/hcwnbs) What version of kruise are you using?

I directly used the image, and the image version is 1.5.4.

Crazy-Hopper:
Pardon, don't know how these bots work. Is this really "won't fix" ? Cause I still need a way to select cipher suites in order to fulfil PCI DSS requirements.

furykerry:
it is still opened, but the previous assigned developer has closed the patch, and we have to find someone else to fix the issue

Crazy-Hopper:
Hi! As per my understanding, please correct me if I'm wrong, this issue wasn't really resolved. If you agree, could you please reopen it ?

PR #1932, opened 2025-02-20
fixed the bug

RohanMishra315:
#1916

Modify the webhook options to explicitly set TLS configuration:
1.  MinVersion: Ensures only TLS 1.2 and above are allowed.
2. CipherSuites: Removes insecure ciphers like CBC-based ones and RSA key exchange.

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"crypto/tls"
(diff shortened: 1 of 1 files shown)


RohanMishra315:
@furykerry made the pr , pls look at it and suggest the necessary changes

PR #1949, opened 2025-03-06
JobSidecarTerminator support ignore exit code capability via env

zmberg:


pkg/controller/sidecarterminator/sidecar_terminator_controller.go
```diff
@@ -136,22 +134,25 @@ func (r *ReconcileSidecarTerminator) doReconcile(pod *corev1.Pod) (reconcile.Res
...
+	if err = r.executeInPlaceUpdateAction(pod, inplaceUpdateSidecarNames); err != nil {
 		return reconcile.Result{}, err
 	}
-
+	if err = r.executeKillContainerAction(pod, normalSidecarNames); err != nil {
+		return reconcile.Result{}, err
+	}
+	if ignoreExitCodeSidecarNames.Len() == 0 || !containersCompleted(pod, normalSidecarNames) || !containersCompleted(pod, inplaceUpdateSidecarNames) {
+		return reconcile.Result{}, nil
+	}
```
furykerry, line 151: plz leave a comment to explain why ignore exit code is not the default behavior and what is the potential risk of enable ingore exit code.

pkg/controller/sidecarterminator/sidecar_terminator_controller.go
```diff
@@ -136,22 +134,25 @@ func (r *ReconcileSidecarTerminator) doReconcile(pod *corev1.Pod) (reconcile.Res
 	if !isInterestingPod(pod) {
 		return reconcile.Result{}, nil
 	}
-
-	sidecarNeedToExecuteKillContainer, sidecarNeedToExecuteInPlaceUpdate, err := r.groupSidecars(pod)
-
+	vk, err := IsPodRunningOnVirtualKubelet(pod, r.Client)
```
furykerry, line 137: is it necessary to check whether node is vk? for non-vk situation, we can also call inplace-update to terminate sidecar if the kruise-daemon is not installed.

zmberg: @furykerry How do we know if there is a kruise-daemon installed or not, it's always just by whether it's a VK node or not.

PR #1976, opened 2025-03-28
Fix: add some logs for WorkloadSpread managing existing pods

AiRanthem:
In the E2E testing of WorkloadSpread, there occasionally appears an issue where it fails to properly handle Pods that existed before the WorkloadSpread resource was created. However, troubleshooting is difficult due to the lack of logs.


furykerry (approved):
/lgtm
/approve

PR #1986, opened 2025-04-03
fix: set the ResourceIsPod of DefaultPodValidationOptions to false

### LINKED ISSUE #1985: [BUG] CloneSet webhook denied the create request of host network container with containerPort but without hostPort
**What happened**:

We are using a cloneset with hostnetwork, but in the podTemplate, we only specified the containerPort without specifying the hostPort (we expected it to be equal to containerPort when it's empty), this used to work correctly in the older versions of openkruise. But after we upgrade the openkruise from 1.6 to 1.7, the CloneSet webhook denied the create request, below error occurs: 

```
2025-04-02T21:17:05.295+0800    ERROR   { "error": "admission webhook \"vcloneset.kb.io\" denied the request: [spec.template.spec.containers[0].ports[0].hostPort: Invalid value: 0: must match `containerPort` when `hostNetwork` is true, spec.template.spec.containers[0].ports[1].hostPort: Invalid value: 0: must match `containerPort` when `hostNetwork` is true, spec.template.spec.containers[0].ports[2].hostPort: Invalid value: 0: must match `containerPort` when `hostNetwork` is true]"}
```

I read the cloneset webhook code https://github.com/openkruise/kruise/blob/483dc2fdd54dbd188b2eeae903504053ce80787d/pkg/webhook/cloneset/validating/validation.go#L78 , it validate the PodTemplate with `webhookutil.DefaultPodValidationOptions` whose ResourceIsPod changed from False to True in openkruise-1.7, and in the [validatePodHostNetworkDeps](https://github.com/kubernetes/kubernetes/blob/3d0594d4d92653fcefea0b91363f473ae457c6b8/pkg/apis/core/validation/validation.go#L3866), it will validate the containerPort with hostPort based on ResourceIsPod. Based on the semantics, `ResourceIsPod` should be False in CloneSet webhook. Other CRs' webhooks might also have this issue. If this is confirmed to be a problem, I can submit a PR to fix it.

this create request should not be denied

just create a cloneset with hostnetwork=true, and specify containerPort without hostPort

**Environment**:
- Kruise version:
- Others:

Yesphet:
fix #1985

### DIFF
--- a/pkg/webhook/util/pod_validation.go
+++ b/pkg/webhook/util/pod_validation.go
@@ -26,6 +26,6 @@ var (
 		AllowInvalidLabelValueInSelector:                  true,
 		AllowHostIPsField:                                 true,
 		AllowInvalidTopologySpreadConstraintLabelSelector: true,
-		ResourceIsPod:                                     true,
+		ResourceIsPod:                                     false,
 	}
 )
--- a/test/e2e/apps/cloneset.go
+++ b/test/e2e/apps/cloneset.go
@@ -63,6 +63,31 @@ var _ = SIGDescribe("CloneSet", func() {
 		randStr = rand.String(10)
 	})
 
+	framework.KruiseDescribe("CloneSet Creating", func() {
+
+		framework.ConformanceIt("creates with hostNetwork pod, specified containerPort without specifying hostPort", func() {
+			cs := tester.NewCloneSet("hostnetwork-port-"+randStr, 1, appsv1alpha1.CloneSetUpdateStrategy{})
+			cs.Spec.Template.Spec.HostNetwork = true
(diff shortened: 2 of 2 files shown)


ABNER-1:
@Yesphet  Please fix the DCO check error by following the instructions here: https://github.com/openkruise/kruise/pull/1986/checks?check_run_id=39890896135

Yesphet:
> @Yesphet Please fix the DCO check error by following the instructions here: https://github.com/openkruise/kruise/pull/1986/checks?check_run_id=39890896135

@ABNER-1 fix

ABNER-1:
Perhaps we could add e2e test cases to solidify this scenario for future reference?

Yesphet:
@ABNER-1 ok, i will add tests for it

furykerry (approved):
/lgtm
/approve

PR #1996, opened 2025-04-16
Fix: WorkloadSpread e2e tests that run in parallel frequently encounter failures.

AiRanthem:
Fixed an issue where, during parallel execution of WorkloadSpread E2E tests, the fake-zone label of a node would be deleted after one test completed, causing other ongoing tests to fail.


AiRanthem:
@furykerry PTAL

furykerry (approved):
/lgtm
/approve

PR #2003, opened 2025-04-22
feat: podProbeMarker support http prober

### LINKED ISSUE #1306: PodProbeMarker Support http/tcp capacities
[PodProbeMarker](https://openkruise.io/docs/user-manuals/podprobemarker) currently only support exec model, so need support tcp/http model, for example:
```yaml
kind: PodProbeMarker
metadata:
  name: game-server-probe
  namespace: ns
spec:
  selector:
    matchLabels:
      app: game-server
  probes:
  - name: Idle
    containerName: game-server
    probe:
      tcpSocket:
        port: 80
      httpGet:
        path: /healthz
        port: 80
        scheme: HTTP
      initialDelaySeconds: 30
      periodSeconds: 10
      timeoutSeconds: 5
    markerPolicy:
    ...
```
Related codes: https://github.com/openkruise/kruise/tree/master/pkg/daemon/podprobe

kagaya85:
support HTTP probe for podProbeMarker

fixes #1306 

by e2e tests

None

### DIFF
--- a/pkg/daemon/podprobe/prober.go
+++ b/pkg/daemon/podprobe/prober.go
@@ -128,27 +128,24 @@ func newRequestForHTTPGetAction(httpGet *v1.HTTPGetAction, podIP string, userAge
 		return nil, fmt.Errorf("invalid port number: %v", port)
 	}
 	path := httpGet.Path
-	url := formatURL(scheme, host, port, path)
+	url, err := formatURL(scheme, host, port, path)
+	if err != nil {
+		return nil, err
+	}
 	headers := v1HeaderToHTTPHeader(httpGet.HTTPHeaders)
 
 	return newProbeRequest(url, headers, userAgentFragment)
 }
 
 // formatURL formats a URL from args.  For testability.
-func formatURL(scheme string, host string, port int, path string) *url.URL {
+func formatURL(scheme string, host string, port int, path string) (*url.URL, error) {
 	u, err := url.Parse(path)
-	// Something is busted with the path, but it's too late to reject it. Pass it along as is.
-	//
-	// This construction of a URL may be wrong in some cases, but it preserves
-	// legacy prober behavior.
 	if err != nil {
-		u = &url.URL{
-			Path: path,
-		}
+		return nil, err
 	}
 	u.Scheme = scheme
 	u.Host = net.JoinHostPort(host, strconv.Itoa(port))
-	return u
+	return u, nil
 }
 
 func userAgent(purpose string) string {
--- a/pkg/daemon/podprobe/prober.go
+++ b/pkg/daemon/podprobe/prober.go
@@ -21,13 +21,19 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"strconv"
 	"time"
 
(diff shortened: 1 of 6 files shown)


pkg/webhook/podprobemarker/validating/probe_create_update_handler.go
```diff
@@ -245,6 +245,25 @@ func validateTCPSocketAction(tcp *corev1.TCPSocketAction, fldPath *field.Path) f
 	return ValidatePortNumOrName(tcp.Port, fldPath.Child("port"))
 }
 
+var supportedHTTPSchemes = sets.New(corev1.URISchemeHTTP, corev1.URISchemeHTTPS)
+
+func validateHTTPGetAction(http *corev1.HTTPGetAction, fldPath *field.Path) field.ErrorList {
+	allErrors := field.ErrorList{}
+	if len(http.Path) == 0 {
+		allErrors = append(allErrors, field.Required(fldPath.Child("path"), ""))
```
furykerry, line 253: shall we validate the format of path e.g. url.Parse

pkg/daemon/podprobe/prober.go
```diff
@@ -71,25 +80,118 @@ func (pb *prober) runProbe(p *appsv1alpha1.ContainerProbeSpec, probeKey probeKey
 		timeSecond = 1
 	}
 	timeout := time.Duration(timeSecond) * time.Second
-	// current only support exec
-	// todo: http
```
furykerry, line 75: maybe we should change the TODO to grpc

kagaya85: done

pkg/webhook/podprobemarker/validating/probe_create_update_handler.go
furykerry: url.Parse support parse relative path(a path, without a host), and that is exactly what formatURL in probe.go do. I think we can  add more strict checking in Kruise, since we don't have some legacy cases to support.

pkg/daemon/podprobe/prober.go
```diff
@@ -86,10 +107,91 @@ func (pb *prober) runProbe(p *appsv1alpha1.ContainerProbeSpec, probeKey probeKey
...
+func formatURL(scheme string, host string, port int, path string) *url.URL {
+	u, err := url.Parse(path)
+	// Something is busted with the path, but it's too late to reject it. Pass it along as is.
+	//
+	// This construction of a URL may be wrong in some cases, but it preserves
+	// legacy prober behavior.
+	if err != nil {
+		u = &url.URL{
+			Path: path,
+		}
```
furykerry, line 147: can we just return the err and do not proceed with the wrong path? I think we can add more strict checking in Kruise, since we don't have some legacy cases to support as in kubernetes code

pkg/webhook/podprobemarker/validating/probe_create_update_handler.go
kagaya85: Actually, okay, I will finish it.

pkg/daemon/podprobe/prober.go
kagaya85: done

PR #2043, opened 2025-05-19
allow adding/removing VCTs for StatefulSet with OnPodRollingUpdate

ninavdl:

Currently, when `OnPodRollingUpdate` is enabled, as per https://openkruise.io/docs/user-manuals/advancedstatefulset#validation-rules-onpodrollingupdate-mode it is not supported to add/remove volumeClaimTemplates to an existing Advanced StatefulSet. 

This PR adds a feature flag `RecreatePodWhenChangeVCTInStatefulSetGate` (similar to the already existing `RecreatePodWhenChangeVCTInCloneSetGate`), with which VCTs can be added to an existing `StatefulSet` with `OnPodRollingUpdate`. Adding/removing a VCT then causes the Pods to be recreated.

NONE

* Create `StatefulSet` with `OnPodRollingUpdate`
* After reconciliation, add or remove a VCT to the `StatefulSet` and mount it in a Pod template
* Observe that the Pod is recreated with the new VCT.


pkg/util/inplaceupdate/inplace_update_defaults.go
```diff
@@ -269,13 +268,10 @@ func defaultCalculateInPlaceUpdateSpec(oldRevision, newRevision *apps.Controller
...
 	}
 
-	// RecreatePodWhenChangeVCTInCloneSetGate enabled
-	if utilfeature.DefaultFeatureGate.Enabled(features.RecreatePodWhenChangeVCTInCloneSetGate) {
-		if !opts.IgnoreVolumeClaimTemplatesHashDiff {
-			canInPlace := volumeclaimtemplate.CanVCTemplateInplaceUpdate(oldRevision, newRevision)
-			if !canInPlace {
-				return nil
-			}
+	if opts.RecreatePodWhenChangedVolumeClaimTemplate && !opts.IgnoreVolumeClaimTemplatesHashDiff {
```
ABNER-1, line 279: Why is it necessary to make this change? It doesn't seem essential.

ABNER-1: 1. The hash is only recorded for the CloneSet.
2. When adding or removing VCTs in a StatefulSet, users may need to add or remove corresponding volume mounts. This implies that the Pod cannot be updated in place.

pkg/webhook/statefulset/validating/statefulset_validation.go
```diff
@@ -389,18 +392,23 @@ func ValidateVolumeClaimTemplateUpdate(c client.Client, sts, oldSts *appsv1beta1
 	var err error
 	for _, template := range sts.Spec.VolumeClaimTemplates {
 		templateIdStr := fmt.Sprintf("volumeClaimTemplates[%v]", template.Name)
+
 		oldTemplate, exist := name2Template[template.Name]
```
ABNER-1, line 396: Perhaps the code below would be simpler.
```
		if !exist {
                        if utilfeature.DefaultFeatureGate.Enabled(features.RecreatePodWhenChangeVCTInStatefulSetGate) {
                            continue
                        }
			return field.ErrorList{field.Forbidden(field.NewPath("spec", templateIdStr, "name"), "volumeClaimTemplate name can not be modified")}
		}

		// ...
```

test/e2e/apps/statefulset.go
```diff
@@ -918,6 +918,110 @@ var _ = SIGDescribe("AppStatefulSetStorage", func() {
 			sst.WaitForStatusPVCReadyReplicas(ss, 1)
 		})
 	})
+
+	ginkgo.Describe("Recreate Pod when modifying volumeClaimTemplate with OnPodRollingUpdate", func() {
```
ABNER-1: Please add cases about sts with`InPlaceUpdateReady`.

ABNER-1: Please add a case about rename volume name in vcts and volumeMounts.

ABNER-1 (commented):
Hi, @ninavdl .
Great job! Your contribution is highly appreciated. 
Here are some suggestions to make it even better.

ABNER-1:
Please add a TODO item near `UpdatePodClaimForRetentionPolicy` to address the removal of PVCs when users delete VCTs.

ABNER-1:
> Please add a TODO item near `UpdatePodClaimForRetentionPolicy` to address the removal of PVCs when users delete VCTs.

Perhaps we can add a label to help users identify these PVCs. 
If we do not recommend deletion during the update process due to the possibility of rollback, this label can allow users to manually clean up PVCs that are no longer referenced by any Pod after a certain period.

pkg/util/inplaceupdate/inplace_update_defaults.go
ninavdl: `opts.RecreatePodWhenChangedVolumeClaimTemplate` is true if the feature gate `RecreatePodWhenChangeVCTInCloneSetGate` or `RecreatePodWhenChangeVCTInStatefulSetGate` is enabled

pkg/webhook/statefulset/validating/statefulset_validation.go
ninavdl: done

test/e2e/apps/statefulset.go
ninavdl: I added a case for renaming VCTs.
What do you mean by sts withInPlaceUpdateReady?

pkg/webhook/statefulset/validating/statefulset_validation.go
```diff
@@ -377,7 +379,8 @@ func ValidateVolumeClaimTemplateUpdate(c client.Client, sts, oldSts *appsv1beta1
 		sts.Spec.VolumeClaimUpdateStrategy.Type == appsv1beta1.OnPVCDeleteVolumeClaimUpdateStrategyType {
 		return nil
 	}
-	if len(sts.Spec.VolumeClaimTemplates) != len(oldSts.Spec.VolumeClaimTemplates) {
+	if len(sts.Spec.VolumeClaimTemplates) != len(oldSts.Spec.VolumeClaimTemplates) &&
+		!utilfeature.DefaultFeatureGate.Enabled(features.RecreatePodWhenChangeVCTInStatefulSetGate) {
```
furykerry, line 383: maybe we can check against the feature gate features.RecreatePodWhenChangeVCTInStatefulSetGate and return nil early in L379

furykerry (commented):
@ninavdl this patch only ensure the pod is recreated if VCT of statefulset changes, but it does not address the problem of changing existing PVC. currently PVC is only deleted when statefulset scale down or statefulset itself get deleted. So we may have to add another PersistentVolumeClaimRetentionPolicyType to make the patch really work

test/e2e/apps/statefulset.go
ABNER-1: sry. 
By default, the StatefulSets created by `NewStatefulSet` use the Recreate update strategy. 
We should add test cases to cover all scenarios (adding, removing, and renaming) for StatefulSets with `InPlaceIfPossiblePodUpdateStrategyType` update strategy.

ABNER-1: You may refer to 
https://github.com/openkruise/kruise/blob/e85fa5b031878a92e03421fded8e4d5268c6321f/test/e2e/apps/statefulset.go#L85-L91

ninavdl:
> @ninavdl this patch only ensure the pod is recreated if VCT of statefulset changes, but it does not address the problem of changing existing PVC. currently PVC is only deleted when statefulset scale down or statefulset itself get deleted. So we may have to add another PersistentVolumeClaimRetentionPolicyType to make the patch really work

Sorry for the late reply. I'm not sure if this is in the scope of this PR?

pkg/webhook/statefulset/validating/statefulset_validation.go
ninavdl: if the feature gate is enabled we still want to check that the SC allows volume expansion

test/e2e/apps/statefulset.go
ninavdl: done

ninavdl:
@furykerry @ABNER-1 could you please review again? :)

PR #2053, opened 2025-05-22
Feature: pub should not protect pod vertical scale -  fix #2049

### LINKED ISSUE #2049: [BUG] pub allowed decreased when do pod-inplace-update
**What happened**:
I enabled the pod resource vertical update feature on Kubernetes 1.28, allowing in-place modifications to pod resources. For OpenKruise, I set the feature gates to PodUnavailableBudgetUpdateGate=true, PodUnavailableBudgetDeleteGate=true, and InPlaceWorkloadVerticalScaling=true. At this point, I created a PUB object with its TargetRef pointing to an ASTS object that has multiple replicas, and I set the upgrade strategy of the ASTS object to onDelete. I then modified the resources/limits of the corresponding pods, and I observed that the allowed count of the PUB object decreased by 1 until the in-place update operation was completed.

In the scenario of vertical scaling, this behavior does not cause any interruption to the application process, so I believe the PUB allowed count should not decrease in this situation. Additionally, even if pub maxUnavailable=0, it should not block this scaling operation, as this operation does not cause any disruption to the application.

Create a pub object, set targetRef to an ASTS. ASTS upgradeStrategy shoule be onDelete, then update ASTS pod resources.

I would like to know whether this is by design or a bug. If it is a bug, I am willing to contribute code to fix it.
I'll modify func`IsPodUnavailableChanged` in file `pkg/control/pubcontrol`  to check if podSpecWithoutResourcesHash changed.

**Environment**:
- Kruise version: 1.8.0
- Kubernetes version (use `kubectl version`): 1.28
- Install details (e.g. helm install args): featureGate = PodUnavailableBudgetUpdateGate=true, PodUnavailableBudgetDeleteGate=true, and InPlaceWorkloadVerticalScaling=true
- Others:

Colvin-Y:
fix #2049

### DIFF
--- a/apis/policy/v1alpha1/podunavailablebudget_types.go
+++ b/apis/policy/v1alpha1/podunavailablebudget_types.go
@@ -27,14 +27,22 @@ import (
 type PubOperation string
 
 const (
-	// PubProtectOperationAnnotation indicates the pub protected Operation[DELETE,UPDATE,EVICT]
-	// if annotations[kruise.io/pub-protect-operations]=EVICT indicates the pub only protect evict pod
-	// if the annotations do not exist, the default DELETE,EVICT,UPDATE are protected
+	// PubProtectOperationAnnotation indicates the pub protected Operation[DELETE,UPDATE,EVICT,RESIZE].
+	// if annotations[kruise.io/pub-protect-operations]=EVICT indicates the pub only protect evict pod.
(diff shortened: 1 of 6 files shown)


abhi0324:
Hi @Colvin-Y,
To ensure the DCO check passes, please sign off your commit.

PR #2061, opened 2025-05-26
feature: add RESIZE action for pub protection.

### LINKED ISSUE #2049: [BUG] pub allowed decreased when do pod-inplace-update

Colvin-Y:
1. Add feature-gate `InPlacePodVerticalScaling` to reflect whether kubernetes can do pod in-place vertical scaling
2. Add `kruise.io/pub-protect-operations: RESIZE` for `podUnavailableBudget` Object

fix #2049

### DIFF
--- a/apis/policy/v1alpha1/podunavailablebudget_types.go
+++ b/apis/policy/v1alpha1/podunavailablebudget_types.go
@@ -27,14 +27,22 @@ import (
 type PubOperation string
 
 const (
-	// PubProtectOperationAnnotation indicates the pub protected Operation[DELETE,UPDATE,EVICT]
-	// if annotations[kruise.io/pub-protect-operations]=EVICT indicates the pub only protect evict pod
-	// if the annotations do not exist, the default DELETE,EVICT,UPDATE are protected
+	// PubProtectOperationAnnotation indicates the pub protected Operation[DELETE,UPDATE,EVICT,RESIZE].
+	// if annotations[kruise.io/pub-protect-operations]=EVICT indicates the pub only protect evict pod.
+	// if the annotations do not exist, the default DELETE,EVICT,UPDATE,RESIZE are protected.
+	// RESIZE is a special UPDATE operation that represents changes to container resources. When the
+	// pod feature gate InPlacePodVerticalScaling is not enabled, it will degrade to the UPDATE strategy.
+	// When the pod feature gate InPlacePodVerticalScaling is enabled, it will protect against cases
+	// where in-place resizing is not possible. Specifically, these scenarios include:
+	// - 1. Changes to resources when restartPolicy == restartContainer.
+	// - 2. Changes to static pod resources.
+	// - 3. Changes to pod QoS.
 	PubProtectOperationAnnotation = "kruise.io/pub-protect-operations"
 	// pod webhook operation
 	PubUpdateOperation PubOperation = "UPDATE"
 	PubDeleteOperation PubOperation = "DELETE"
 	PubEvictOperation  PubOperation = "EVICT"
+	PubResizeOperation PubOperation = "RESIZE"
 	// PubProtectTotalReplicasAnnotation is the target replicas.
 	// By default, PUB will get the target replicas through workload.spec.replicas. but there are some scenarios that may workload doesn't
 	// implement scale subresources or Pod doesn't have workload management. In this scenario, you can set pub.kruise.io/protect-total-replicas
--- a/apis/policy/v1alpha1/podunavailablebudget_types.go
+++ b/apis/policy/v1alpha1/podunavailablebudget_types.go
@@ -31,7 +31,8 @@ const (
 	// if annotations[kruise.io/pub-protect-operations]=EVICT indicates the pub only protect evict pod.
 	// if the annotations do not exist, the default DELETE,EVICT,UPDATE are protected.
 	// RESIZE: Pod vertical scaling action. If it's enabled, all resize action will be protected. RESIZE
-	// is an extension of UPDATE, so if RESIZE is enabled, UPDATE will also be protected.
+	// is an extension of UPDATE, if RESIZE is disabled and UPDATE is enabled, any UPDATE operation will
+	// be protected only as it will definitely cause container restarts.
 	// UPDATE: Kruise will carefully differentiate whether this update will cause interruptions. When
 	// the FeatureGate InPlacePodVerticalScaling is enabled, pod inplace vertical scaling will be
 	// considered non-disruption only when allowedResources(cpu、memory) changes、restartPolicy
--- a/apis/policy/v1alpha1/podunavailablebudget_types.go
+++ b/apis/policy/v1alpha1/podunavailablebudget_types.go
@@ -27,16 +27,16 @@ import (
 type PubOperation string
 
 const (
-	// PubProtectOperationAnnotation indicates the pub protected Operation[DELETE,UPDATE,EVICT,RESIZE].
+	// PubProtectOperationAnnotation indicates the pub protected Operation[DELETE,UPDATE,EVICT].
 	// if annotations[kruise.io/pub-protect-operations]=EVICT indicates the pub only protect evict pod.
-	// if the annotations do not exist, the default DELETE,EVICT,UPDATE,RESIZE are protected.
-	// RESIZE is a special UPDATE operation that represents changes to container resources. When the
-	// pod feature gate InPlacePodVerticalScaling is not enabled, it will degrade to the UPDATE strategy.
-	// When the pod feature gate InPlacePodVerticalScaling is enabled, it will protect against cases
-	// where in-place resizing is not possible. Specifically, these scenarios include:
-	// - 1. Changes to resources when restartPolicy == restartContainer.
-	// - 2. Changes to static pod resources.
-	// - 3. Changes to pod QoS.
+	// if the annotations do not exist, the default DELETE,EVICT,UPDATE are protected.
+	// RESIZE: Pod vertical scaling action. If it's enabled, all resize action will be protected. RESIZE
+	// is an extension of UPDATE, so if RESIZE is enabled, UPDATE will also be protected.
+	// UPDATE: Kruise will carefully differentiate whether this update will cause interruptions. When
+	// the FeatureGate InPlacePodVerticalScaling is enabled, pod inplace vertical scaling will be
+	// considered non-disruption only when allowedResources(cpu、memory) changes、restartPolicy
+	// is not restartContainer、is not static pod and QoS not changed. But if featureGate
+	// InPlacePodVerticalScaling is disabled, all resize action will be considered as disruption.
 	PubProtectOperationAnnotation = "kruise.io/pub-protect-operations"
 	// pod webhook operation
 	PubUpdateOperation PubOperation = "UPDATE"
--- a/pkg/control/pubcontrol/pub_control.go
+++ b/pkg/control/pubcontrol/pub_control.go
@@ -18,6 +18,9 @@ package pubcontrol
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
 	"reflect"
 	"strconv"
 	"strings"
@@ -25,17 +28,22 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
+	v1qos "k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
 	kubecontroller "k8s.io/kubernetes/pkg/controller"
+	"k8s.io/kubernetes/pkg/kubelet/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	appspub "github.com/openkruise/kruise/apis/apps/pub"
 	policyv1alpha1 "github.com/openkruise/kruise/apis/policy/v1alpha1"
 	"github.com/openkruise/kruise/pkg/control/sidecarcontrol"
+	"github.com/openkruise/kruise/pkg/features"
 	"github.com/openkruise/kruise/pkg/util"
 	utilclient "github.com/openkruise/kruise/pkg/util/client"
 	"github.com/openkruise/kruise/pkg/util/controllerfinder"
+	"github.com/openkruise/kruise/pkg/util/feature"
 	"github.com/openkruise/kruise/pkg/util/inplaceupdate"
 )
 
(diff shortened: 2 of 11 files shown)


Colvin-Y:
@kruise-bot /retest

pkg/control/pubcontrol/pub_control.go
```diff
@@ -69,6 +82,139 @@ func (c *commonControl) IsPodUnavailableChanged(oldPod, newPod *corev1.Pod) bool
...
+	return isNeedPubProtection(pub, policyv1alpha1.PubResizeOperation)
+}
+
+func (c *commonControl) canResizeInplace(oldPod, newPod *corev1.Pod) bool {
+	if !c.isNeedProtectResizeAction(newPod) {
+		return false
+	}
+
+	// now check if only container resources changed
+	oldPodSpecWithoutResourcesHash, err := podSpecWithoutResourcesHash(oldPod)
```
furykerry, line 106: since currently only resize of cpu and mem is supported, we need to check against the changes of other resources

pkg/control/pubcontrol/pub_control.go
```diff
@@ -56,8 +64,13 @@ func (c *commonControl) IsPodReady(pod *corev1.Pod) bool {
 }
 
 func (c *commonControl) IsPodUnavailableChanged(oldPod, newPod *corev1.Pod) bool {
-	// If pod.spec changed, pod will be in unavailable condition
+	// If pod.spec changed, pod may be in unavailable condition
 	if !reflect.DeepEqual(oldPod.Spec, newPod.Spec) {
+		if c.canResizeInplace(oldPod, newPod) {
+			klog.V(3).InfoS("Pod resize inplace", "pod", klog.KObj(newPod))
```
furykerry, line 70: if the pod can resize inplace, it still need change UnavailableLabel, so can not return false directly

pkg/control/pubcontrol/pub_control.go
```diff
@@ -69,6 +82,139 @@ func (c *commonControl) IsPodUnavailableChanged(oldPod, newPod *corev1.Pod) bool
...
+	if oldOk && !newOk || !oldOk && newOk {
+		return true
+	}
+
+	// both exist
+	return !oldResource.Equal(newResource)
+}
+
+func podSpecWithoutResourcesHash(podIn *corev1.Pod) (string, error) {
+	pod := podIn.DeepCopy()
```
furykerry, line 199: considering the InPlaceUpdateEnvFromMetadata feature, we need also checking the annotation bind with env using downward api

pkg/control/pubcontrol/pub_control.go
```diff
@@ -69,6 +82,139 @@ func (c *commonControl) IsPodUnavailableChanged(oldPod, newPod *corev1.Pod) bool
...
+	if err != nil || pub == nil {
+		return false
+	}
+
+	return isNeedPubProtection(pub, policyv1alpha1.PubResizeOperation)
+}
+
+func (c *commonControl) canResizeInplace(oldPod, newPod *corev1.Pod) bool {
+	if !c.isNeedProtectResizeAction(newPod) {
+		return false
```
furykerry, line 102: the logic seems be reversed. if one does need protect resize action, we should account the resizing as disruption even if the resize may not case container starts. if one has no need to protect resize action, we can safely treat inplace-resizing as no disruption along as the pod qos remain the same.

pkg/control/pubcontrol/pub_control.go
Colvin-Y: ack, and I'll add case to cover this

pkg/control/pubcontrol/pub_control.go
Colvin-Y: > considering the InPlaceUpdateEnvFromMetadata feature

But we cannot get featureGate info in kruise-daemon.

pkg/control/pubcontrol/pub_control.go
Colvin-Y: > 逻辑似乎是相反的。如果确实需要保护调整大小操作，我们应该将调整大小视为中断，即使调整大小可能不会导致容器启动。如果不需要保护调整大小操作，我们可以安全地将就地调整大小视为无中断，因为 pod qos 保持不变。

The current design specifies that if targeted protection for the RESIZE behavior is needed and the feature gate InPlacePodVerticalScaling is enabled, then the in-place resizing will be differentiated finely. Otherwise, it will degrade to the UPDATE strategy. Which option is relatively more reasonable? For a detailed description of the specific strategy, please refer to apis/policy/v1alpha1/podunavailablebudget_types.go#L30.

pkg/control/pubcontrol/pub_control.go
Colvin-Y: If users implement in-place scaling for scalar resources, should we consider adding the corresponding configuration option as an advanced setting, i.e., as an environment variable?

pkg/control/pubcontrol/pub_control.go
furykerry: maybe we can have a discussion in the community meeting this week?

Colvin-Y: 经过讨论后我认为合适的解法可能如下：

furykerry:
> 2.2. If the FeatureGate InPlacePodVerticalScaling is enabled: it will finely differentiate whether it is a pod-inplace vertical scaling behavior.

yes, if the resize operation is not listed, and it is pod-inplace vertical scaling, then we do not consider it as disruption

Colvin-Y:
@kruise-bot  /retest

Colvin-Y:
could not tigger retest, ready to review.

apis/policy/v1alpha1/podunavailablebudget_types.go
```diff
@@ -27,14 +27,22 @@ import (
...
 
 const (
-	// PubProtectOperationAnnotation indicates the pub protected Operation[DELETE,UPDATE,EVICT]
-	// if annotations[kruise.io/pub-protect-operations]=EVICT indicates the pub only protect evict pod
-	// if the annotations do not exist, the default DELETE,EVICT,UPDATE are protected
+	// PubProtectOperationAnnotation indicates the pub protected Operation[DELETE,UPDATE,EVICT].
+	// if annotations[kruise.io/pub-protect-operations]=EVICT indicates the pub only protect evict pod.
+	// if the annotations do not exist, the default DELETE,EVICT,UPDATE are protected.
+	// RESIZE: Pod vertical scaling action. If it's enabled, all resize action will be protected. RESIZE
+	// is an extension of UPDATE, so if RESIZE is enabled, UPDATE will also be protected.
```
furykerry, line 34: > if RESIZE is enabled, UPDATE will also be protected.
this comment seems misleading. Maybe a better comment is :

if RESIZE is disabled and UPDATE is enabled, any UPDATE operation will be protected only as it will definitely cause container restarts.

pkg/control/pubcontrol/pub_control.go
```diff
@@ -56,19 +64,203 @@ func (c *commonControl) IsPodReady(pod *corev1.Pod) bool {
...
+		!feature.DefaultFeatureGate.Enabled(features.InPlacePodVerticalScaling) {
+		return true
+	}
+
+	return false
+}
+
+// return true if this action won't cause unavailability
+func (c *commonControl) canResizeInplace(oldPod, newPod *corev1.Pod) bool {
+	if c.isNeedProtectResizeAction(newPod) {
```
furykerry, line 109: maybe we should not call isNeedProtectResizeAction at all, since isNeedPubProtection will be called in PodUnavailableBudgetValidatePod.

apis/policy/v1alpha1/podunavailablebudget_types.go
Colvin-Y: ack, thx

pkg/control/pubcontrol/pub_control.go
Colvin-Y: ACK
> maybe we should not call isNeedProtectResizeAction at all

Pre-checking is aimed at reducing unnecessary performance overhead.

> change the operation from PubUpdateOperation to PubResizeOperation

It was initially my intention to treat RESIZE as a special case of UPDATE, but now it seems your suggestion is better. I will modify it.

Colvin-Y:
@kruise-bot  /retest

pkg/control/pubcontrol/pub_control_utils.go
```diff
@@ -251,9 +253,29 @@ func IsReferenceEqual(ref1, ref2 *policyv1alpha1.TargetReference) bool {
 
 func isNeedPubProtection(pub *policyv1alpha1.PodUnavailableBudget, operation policyv1alpha1.PubOperation) bool {
 	operationValue, ok := pub.Annotations[policyv1alpha1.PubProtectOperationAnnotation]
+	enableInPlacePodVerticalScaling := feature.DefaultFeatureGate.Enabled(features.InPlacePodVerticalScaling)
 	if !ok || operationValue == "" {
+		// when operation == RESIZE
+		// 1. if featureGate InPlacePodVerticalScaling is enabled, then do not protect resize
+		// 2. if featureGate InPlacePodVerticalScaling is disabled, then RESIZE is regarded as UPDATE, so return true
+		if operation == policyv1alpha1.PubResizeOperation {
+			return !enableInPlacePodVerticalScaling
```
furykerry: can we move the checking against features.InPlacePodVerticalScaling into CanResizeInplace func? it seems more appropriately semantically there.

pkg/control/pubcontrol/pub_control_utils.go
```diff
@@ -251,9 +253,29 @@ func IsReferenceEqual(ref1, ref2 *policyv1alpha1.TargetReference) bool {
...
-	operations := sets.NewString(strings.Split(operationValue, ",")...)
+
+	operations := func() sets.Set[string] {
+		operations := sets.New(strings.Split(operationValue, ",")...)
+		// if resize set in pub protect operation, then protect update and resize
+		if operations.Has(string(policyv1alpha1.PubResizeOperation)) {
+			operations.Insert(string(policyv1alpha1.PubUpdateOperation))
+		}
+		// if disable featureGate InPlacePodVerticalScaling, update will contain resize
+		if !enableInPlacePodVerticalScaling && operations.Has(string(policyv1alpha1.PubUpdateOperation)) {
```
furykerry: if disable featureGate InPlacePodVerticalScaling, resize operation will be treated as update operation, so just checking whether update operation should be protected is enough, it seems not necessary to insert extra resize operation in the allowed list.

pkg/control/pubcontrol/pub_control_utils.go
```diff
@@ -251,9 +253,29 @@ func IsReferenceEqual(ref1, ref2 *policyv1alpha1.TargetReference) bool {
...
+			return !enableInPlacePodVerticalScaling
+		}
 		return true
 	}
-	operations := sets.NewString(strings.Split(operationValue, ",")...)
+
+	operations := func() sets.Set[string] {
+		operations := sets.New(strings.Split(operationValue, ",")...)
+		// if resize set in pub protect operation, then protect update and resize
+		if operations.Has(string(policyv1alpha1.PubResizeOperation)) {
```
furykerry: why allowing resize implies allowing update?

pkg/control/pubcontrol/pub_control_utils.go
Colvin-Y: From a code structure perspective, the `isNeedPubProtection` function is called after confirming the `ACTION` type, while the `CanResizeInplace` function is solely used to determine changes in the Pod and does not concern itself with PUB configuration. The status of this function is equivalent to that of `IsPodUnavailableChanged`. Therefore, I believe it is reasonable to place it in `isNeedPubProtection`.

pkg/control/pubcontrol/pub_control_utils.go
Colvin-Y: If input param `operaion` is `RESIZE` but protected operation array is `[UPDATE]`, we also need return true.

pkg/control/pubcontrol/pub_control_utils.go
Colvin-Y: RESIZE is a special type of UPDATE operation, and we believe it should semantically include UPDATE. Additionally, in practical scenarios, it seems there are no cases where RESIZE needs protection while UPDATE does not. If we must make a distinction, perhaps we should modify UPDATE to NON_RESIZE_UPDATE, but this seems unnecessary.

furykerry: previous UPDATE only protect against the update of container image which will case inplace-update of container, so to ensure the compatibility, i think limit the scope of UPDATE operation to only  inplace-update is more safer. In addition, if we consider RESIZE as a special UPDATE, the implicit relationship can be hard to notice for users. As for the practical scenarios, the users may choose to only protect RESIZE but not UPDATE do exists, since RESIZE may be caused by VPA component, while UPDATE is generally operated by workload controller which already consider maxUnavaiable.

Colvin-Y: Got it. In other words, UPDATE will include RESIZE when InPlacePodVerticalScaling is not enabled, but RESIZE should not be associated with UPDATE under any circumstances.

pkg/control/pubcontrol/pub_control_utils.go
```diff
@@ -251,9 +253,25 @@ func IsReferenceEqual(ref1, ref2 *policyv1alpha1.TargetReference) bool {
 
 func isNeedPubProtection(pub *policyv1alpha1.PodUnavailableBudget, operation policyv1alpha1.PubOperation) bool {
 	operationValue, ok := pub.Annotations[policyv1alpha1.PubProtectOperationAnnotation]
+	enableInPlacePodVerticalScaling := feature.DefaultFeatureGate.Enabled(features.InPlacePodVerticalScaling)
 	if !ok || operationValue == "" {
+		// when operation == RESIZE
+		// 1. if featureGate InPlacePodVerticalScaling is enabled, then do not protect resize
+		// 2. if featureGate InPlacePodVerticalScaling is disabled, then RESIZE is regarded as UPDATE, so return true
+		if operation == policyv1alpha1.PubResizeOperation {
```
furykerry: maybe we can rewrite the logic as assigning the default value "DELETE,UPDATE,EVICT" to operationValue, so that we can avoid duplicate the logic here and L270

Colvin-Y: thx for your suggestion. The function has been refactored, and the logic now looks clearer and more concise.

PR #2067, opened 2025-05-28
upgrade k8s deps to 1.32

chengjoey:
1. upgrade k8s deps to 1.32
2. use patch resize for inplace update resources in 1.32+
3.  add 1.32 e2e test

feat: #2054


ABNER-1:
Hi, @chengjoey.  Please rebase onto the latest master branch, as GitHub Actions is now configured to ignore ubuntu-20.04.

chengjoey:
/hold

pkg/daemon/kuberuntime/kuberuntime_container.go:L99
HashWithoutResources broken by this [pr](https://github.com/kubernetes/kubernetes/pull/124220)

chengjoey:
/hold cancel

@ABNER-1 @LavenderQAQ PLAT

It seems that I have broken the inplace updateResource logic, is there any background information I need to know?

ABNER-1:
> /hold cancel
> 
> @ABNER-1 @LavenderQAQ PLAT
> 
> It seems that I have broken the inplace updateResource logic, is there any background information I need to know?
Hi @chengjoey 
Wait a moment while I sort out the current errors and the changes in the PR.

ABNER-1:
@chengjoey  https://github.com/kubernetes/kubernetes/pull/128266
The resize sub-resource was introduced in version 1.32, and modifications via the update interface are prohibited. We need to adapt to this change and control it through a feature gate, which should automatically detect the Kubernetes version in the charts to toggle the feature on or off.

ABNER-1:
https://github.com/chengjoey/kruise/blob/a7721b01af645455236f10e8a15f0717dc79f16a/pkg/util/inplaceupdate/inplace_update_defaults.go#L449-L456

https://github.com/chengjoey/kruise/blob/a7721b01af645455236f10e8a15f0717dc79f16a/pkg/util/inplaceupdate/inplace_update_defaults.go#L533C1-L540C5

In earlier versions of Kruise, the hash value calculated by kubeletcontainer.HashContainer was compared with the hash value recorded in CRI. However, this function was modified in version 1.31. To accommodate the hash value of older Kubernetes versions, I think we should be compatible with this logic.
Similar modifications are as follows:

```go
// hashContainer copy from kubelet v1.31-
func hashContainer(container *v1.Container) uint64 {
	hash := fnv.New32a()
	// Omit nil or empty field when calculating hash value
	// Please see https://github.com/kubernetes/kubernetes/issues/53644
	containerJSON, _ := json.Marshal(container)
	hashutil.DeepHashObject(hash, containerJSON)
	return uint64(hash.Sum32())
}
```


```go
                     case plainHash:
			if expectedHash := kubeletcontainer.HashContainer(containerSpec); containerMeta.Hashes.PlainHash != expectedHash {
				klog.InfoS("Find container in runtime-container-meta for Pod has different plain hash(hash only image) with spec",
					"containerName", containerSpec.Name, "namespace", pod.Namespace, "podName", pod.Name,
					"metaHash", containerMeta.Hashes.PlainHash, "expectedHash", expectedHash)
			}
			// To adapt to Kubernetes environments running versions below 1.31,
			// the hash value reported by the daemon is still calculated by the lower-version kubelet.
			//i.e., using all fields for the computation.
			if expectedHash := hashContainer(containerSpec); containerMeta.Hashes.PlainHash != expectedHash {
				klog.InfoS("Find container in runtime-container-meta for Pod has different plain hash(hash all field) with spec",
					"containerName", containerSpec.Name, "namespace", pod.Namespace, "podName", pod.Name,
					"metaHash", containerMeta.Hashes.PlainHash, "expectedHash", expectedHash)
				return false
			}
```

chengjoey:
@ABNER-1 Now all the CI have passed (except codecov), PTAL

furykerry:
@chengjoey can you attend the community meeting this week, we can discuss about this upgrade and possible risks associating with upgrading

.github/workflows/e2e-1.18.yaml
```diff
@@ -11,7 +11,7 @@ permissions: read-all
 
 env:
   # Common versions
-  GO_VERSION: '1.22'
+  GO_VERSION: '1.23'
   KIND_VERSION: 'v0.14.0'
   KIND_IMAGE: 'kindest/node:v1.18.20'
```
furykerry: we can delete e2e for 1.18 and 1.20, it is not possible for us to cover the tests for every kubernetes version

Makefile
```diff
@@ -160,7 +160,7 @@ ENVTEST ?= $(TESTBIN)/setup-envtest
 envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(TESTBIN)
 ifeq (, $(shell ls $(TESTBIN)/setup-envtest 2>/dev/null))
-	GOBIN=$(TESTBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@c7e1dc9b5302d649d5531e19168dd7ea0013736d
+	GOBIN=$(TESTBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
```
furykerry: plz pin the tool to specific version to avoid un-expected upgrade issue

pkg/util/inplaceupdate/inplace_update_defaults.go
```diff
@@ -537,20 +540,24 @@ func checkAllContainersHashConsistent(pod *v1.Pod, runtimeContainerMetaSet *apps
...
-			containerSpecCopy := containerSpec.DeepCopy()
-			containerSpecCopy.Resources = v1.ResourceRequirements{}
-			if expectedHash := kubeletcontainer.HashContainer(containerSpecCopy); containerMeta.Hashes.PlainHashWithoutResources != expectedHash {
-				klog.InfoS("Find container in runtime-container-meta for Pod has different plain hash with spec(except resources)",
-					"containerName", containerSpecCopy.Name, "namespace", pod.Namespace, "podName", pod.Name,
-					"metaHash", containerMeta.Hashes.PlainHashWithoutResources, "expectedHash", expectedHash)
-				return false
+			ver := client.GetCurrentServerVersion()
+			if semver.New(fmt.Sprintf("%s.%s.0", ver.Major, ver.Minor)).Compare(*semver.New("1.31.0")) >= 0 {
+				if expectedHash := kubeletcontainer.HashContainer(containerSpec); containerMeta.Hashes.PlainHash != expectedHash {
```
ABNER-1: I don't think this change can solve all the problems:
1. When the apisever is 1.32 but kubelet is still 1.30, the pod hash generated will still follow the calculation logic of version 1.30.
2. Even if both apiserver and kubelet are upgraded to 1.32, there may still be some containers that were created by the old version of kubelet. In this case, the pod hash will remain the same as the old version, which can cause the in-place upgrade to get stuck.

pkg/util/inplaceupdate/inplace_update.go
```diff
@@ -227,10 +230,24 @@ func (c *realControl) finishGracePeriod(pod *v1.Pod, opts *UpdateOptions) (time.
...
 			}
 
-			if clone, err = opts.PatchSpecToPod(clone, &spec, &updateState); err != nil {
+			var expectedResources map[string]*v1.ResourceRequirements
+			clone, expectedResources, err = opts.PatchSpecToPod(clone, &spec, &updateState)
+			if err != nil {
 				return err
 			}
 			appspub.RemoveInPlaceUpdateGrace(clone)
+			if utilfeature.DefaultFeatureGate.Enabled(features.InPlaceUpdateResourceByResize) && len(expectedResources) != 0 {
```
ABNER-1: Hi, @chengjoey.
I've had a change of mind. I now believe that using client.GetCurrentServerVersion() to switch between the two branches is a more effective approach than introducing the InPlaceUpdateResourceByResize feature gate.

ABNER-1 (changes requested):

cmd/daemon/main.go
```diff
@@ -62,14 +62,14 @@ func main() {
...
 	rand.Seed(time.Now().UnixNano())
-	features.SetDefaultFeatureGates()
-	ctrl.SetLogger(klogr.New())
-
 	cfg := config.GetConfigOrDie()
 	cfg.UserAgent = "kruise-daemon"
 	if err := client.NewRegistry(cfg); err != nil {
 		klog.Fatalf("Failed to init clientset registry: %v", err)
 	}
+	features.SetDefaultFeatureGates()
```
furykerry: plz comment in the code why set logger should be placed here, explain possible dependency explicitly in case others breaks the dependencies

pkg/util/inplaceupdate/inplace_update_defaults.go
```diff
@@ -537,20 +540,24 @@ func checkAllContainersHashConsistent(pod *v1.Pod, runtimeContainerMetaSet *apps
...
-		case plainHashWithoutResources:
-			containerSpecCopy := containerSpec.DeepCopy()
-			containerSpecCopy.Resources = v1.ResourceRequirements{}
-			if expectedHash := kubeletcontainer.HashContainer(containerSpecCopy); containerMeta.Hashes.PlainHashWithoutResources != expectedHash {
-				klog.InfoS("Find container in runtime-container-meta for Pod has different plain hash with spec(except resources)",
-					"containerName", containerSpecCopy.Name, "namespace", pod.Namespace, "podName", pod.Name,
-					"metaHash", containerMeta.Hashes.PlainHashWithoutResources, "expectedHash", expectedHash)
-				return false
+			ver := client.GetCurrentServerVersion()
+			if semver.New(fmt.Sprintf("%s.%s.0", ver.Major, ver.Minor)).Compare(*semver.New("1.31.0")) >= 0 {
```
furykerry: we should try to match the hash using both the new and old hashing algorithm, so as to handle the old pod created before k8s upgrade.

pkg/util/inplaceupdate/inplace_update_defaults.go
```diff
@@ -566,6 +573,15 @@ func checkAllContainersHashConsistent(pod *v1.Pod, runtimeContainerMetaSet *apps
 	return true
 }
 
+func hashContainer(container *v1.Container) uint64 {
```
furykerry, line 567: plz comment the func is used to mimic legacy hash method in previous kubelet

pkg/util/inplaceupdate/inplace_update_vertical_test.go
```diff
@@ -269,3 +269,39 @@ func TestIsContainerUpdateCompleted(t *testing.T) {
...
 }
+
+func TestGenerateResourcePatch(t *testing.T) {
+	v := NativeVerticalUpdate{}
+	pod := &v1.Pod{
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{
+				{
+					Name: "test-container",
+					Resources: v1.ResourceRequirements{
```
furykerry, line 281: plz add more actual field such image to the spec, so that we can ensure that the resource patch will not include fields other than the resources

pkg/util/inplaceupdate/inplace_update_vertical_test.go
```diff
@@ -269,3 +269,39 @@ func TestIsContainerUpdateCompleted(t *testing.T) {
...
+						Limits:   v1.ResourceList{v1.ResourceCPU: resource.MustParse("100m")},
+						Requests: v1.ResourceList{v1.ResourceMemory: resource.MustParse("512Mi")},
```
furykerry, line 283: plz also add cases that include non-resizable resources

pkg/util/inplaceupdate/inplace_update_vertical.go
```diff
@@ -168,6 +171,44 @@ func (v *NativeVerticalUpdate) UpdateResource(pod *v1.Pod, expectedResources map
 	return
 }
 
+func (v *NativeVerticalUpdate) GenerateResourcePatch(pod *v1.Pod, expectedResources map[string]*v1.ResourceRequirements) []byte {
+	var patchContainers []string
```
furykerry, line 175: plz add a case for a patch without any container

pkg/util/inplaceupdate/inplace_update.go
```diff
@@ -266,9 +283,23 @@ func (c *realControl) updateNextBatch(pod *v1.Pod, opts *UpdateOptions) (bool, e
...
+			if patchResources != nil {
+				if adp, ok := c.podAdapter.(podadapter.AdapterWithPatch); ok {
+					cloneCopy := clone.DeepCopy()
+					_, err = adp.PatchPodResource(cloneCopy, client.RawPatch(types.StrategicMergePatchType, patchResources))
+					if err != nil {
+						return err
+					}
+				}
+			}
+		}
```
furykerry, line 301: is it possible that pod update is not necessary for a resize only operation.

Makefile
```diff
@@ -30,7 +30,7 @@ all: build
 ##@ Development
 
 go_check:
-	@scripts/check_go_version "1.22"
+	@scripts/check_go_version "1.23"
```
furykerry, line 33: we should also update the controlloer-tools version to v0.17.3 in L122

go.mod
```diff
@@ -10,169 +10,174 @@ require (
...
-	golang.org/x/tools v0.21.1-0.20240508182429-e35e4ccd0d2d // indirect
-	google.golang.org/appengine v1.6.8 // indirect
-	google.golang.org/genproto/googleapis/api v0.0.0-20240227224415-6ceb2ff114de // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20240401170217-c3f982113cda // indirect
-	google.golang.org/protobuf v1.33.0 // indirect
+	golang.org/x/tools v0.26.0 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20240826202546-f6391c0de4c7 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20240826202546-f6391c0de4c7 // indirect
+	google.golang.org/protobuf v1.35.1 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
```
furykerry, line 134: can you run go mod tidy once again?  it seems that the tidy run may remove extra unused dependencies such as 
1. github.com/sirupsen/logrus v1.9.3
2.  github.com/opencontainers/runc v1.2.1

chengjoey:
hi @ABNER-1 @furykerry , thanks for reviewing

I will make the requested changes, and then the community meeting is this Thursday evening, right? I will try to attend if I have time.

furykerry:
> hi @ABNER-1 @furykerry , thanks for reviewing
> 
> I will make the requested changes, and then the community meeting is this Thursday evening, right? I will try to attend if I have time.

yes, the community meeting is tonight

pkg/util/inplaceupdate/inplace_update_defaults.go
chengjoey:
```
case plainHash:
			isConsistentInNewVersion := kubeletcontainer.HashContainer(containerSpec) == containerMeta.Hashes.PlainHash
			isConsistentInOldVersion := hashContainer(containerSpec) == containerMeta.Hashes.PlainHash
			if !isConsistentInNewVersion && !isConsistentInOldVersion {
				klog.InfoS("Find container in runtime-container-meta for Pod has different plain hash with spec",
					"containerName", containerSpec.Name, "namespace", pod.Namespace, "podName", pod.Name,
					"metaHash", containerMeta.Hashes.PlainHash, "expectedHashInNewVersion", kubeletcontainer.HashContainer(containerSpec), "expectedHashInOldVersion", hashContainer(containerSpec))
				return false
			}
```

How do you think it is changed like this, if both isConsistentInNewVersion and isConsistentInOldVersion do not match, then return false, otherwise we consider it consistent.

pkg/util/inplaceupdate/inplace_update.go
chengjoey: Yes, using client.GetCurrentServerVersion() might be better, as this should not be an option for the user.

Makefile
chengjoey: https://github.com/kubernetes-sigs/controller-runtime/blob/32e5e9e948a572779280969aaaf7db92f76d8252/tools/setup-envtest/README.md?plain=1#L7-L12

we upgrade go to 1.23, so use latest it's ok

.github/workflows/e2e-1.18.yaml
chengjoey: done

cmd/daemon/main.go
chengjoey: new featuregate is deleted, so i have reverted this change

pkg/util/inplaceupdate/inplace_update_defaults.go
chengjoey: commented in https://github.com/openkruise/kruise/pull/2067#discussion_r2168230609

pkg/util/inplaceupdate/inplace_update_defaults.go
chengjoey: done

pkg/util/inplaceupdate/inplace_update_vertical_test.go
chengjoey: all test cases added in TestGenerateResourcePatch

pkg/util/inplaceupdate/inplace_update.go
chengjoey: I think this is feasible, if it only needs to update resources and does not include image, env, etc., it can just need patch resize.

Currently, I have added TODO, which can be done as an optimization item, as there is no blocking feature, and the current PR already has a lot of changes.

Makefile
chengjoey: done

go.mod
chengjoey: done

chengjoey:
image

also fix flaky unit-test



https://github.com/openkruise/kruise/blob/fc611f0505d14bfe049040de7e91d0c217be46bb/pkg/daemon/imagepuller/imagepuller_controller.go#L98-L102


@ABNER-1 there seems go is unnecessary, `workerLimitedPool.Start()` is enough, what do you think

ABNER-1:
> image
> also fix flaky unit-test
> 
> https://github.com/openkruise/kruise/blob/fc611f0505d14bfe049040de7e91d0c217be46bb/pkg/daemon/imagepuller/imagepuller_controller.go#L98-L102
> 
> @ABNER-1 there seems go is unnecessary, `workerLimitedPool.Start()` is enough, what do you think

plz remove it. Thanks!

Makefile
chengjoey: done, set tag as v0.0.0-20250620151452-b9a9ca01fd37

chengjoey:
> plz remove it. Thanks!

done

Makefile
```diff
@@ -160,7 +160,7 @@ ENVTEST ?= $(TESTBIN)/setup-envtest
 envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(TESTBIN)
 ifeq (, $(shell ls $(TESTBIN)/setup-envtest 2>/dev/null))
-	GOBIN=$(TESTBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@c7e1dc9b5302d649d5531e19168dd7ea0013736d
+	GOBIN=$(TESTBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20250620151452-b9a9ca01fd37
```
furykerry: this version requires golang 1.24, can we just use release-0.20 which only requires golang 1.23

chengjoey: done
because I am using version 1.24 locally, so I didn't meet this issue.
now i change tag to v0.0.0-20250517180713-32e5e9e948a5 (release-0.20), it's ok for 1.23

furykerry: i cannot find tag for v0.0.0-20250517180713-32e5e9e948a5 in github, is it possible to just use the branch name (release-0.20)

chengjoey:
```
go install sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.20.0
go: downloading sigs.k8s.io/controller-runtime v0.20.0
go: sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.20.0: module sigs.k8s.io/controller-runtime@v0.20.0 found, but does not contain package sigs.k8s.io/controller-runtime/tools/setup-envtest
```

in my test ,i can't use 0.20.0 tag for setup-envtest

since we use 1.23+, maybe we should use `latest`, otherwise we should only use commitid
image

Issue #2071, opened 2025-05-31
[BUG] Never completionPolicy ImagePullJob timeout setting should consider backoffLimit


MajLuu:
**What happened**:
In our production env, we consider that never completionPolicy ImagePullJob timeout setting should consider backoffLimit, not only timeoutSeconds setting
**What you expected to happen**:
never completionPolicy ImagePullJob timeout = max(defaultActiveDeadlineSecondsForNever, timeoutSeconds*backoffLimit)
**How to reproduce it (as minimally and precisely as possible)**:
None
**Anything else we need to know?**:
If you think this scenario is a bug, I have personally made the relevant PR. The modification has already been applied to our production environment.
**Environment**:
- Kruise version: latest
- Kubernetes version (use `kubectl version`): None
- Install details (e.g. helm install args): None
- Others: None

PR #2072, opened 2025-05-31
Fix: Never completionPolicy ImagePullJob timeout setting should consider backoffLimit

MajLuu:
issue: [2071](https://github.com/openkruise/kruise/issues/2071)

### DIFF
--- a/pkg/controller/imagepulljob/imagepulljob_utils.go
+++ b/pkg/controller/imagepulljob/imagepulljob_utils.go
@@ -111,13 +111,14 @@ func getTTLSecondsForNever() *int32 {
 }
 
 func getActiveDeadlineSecondsForNever(job *appsv1alpha1.ImagePullJob) *int64 {
-	if job.Spec.PullPolicy != nil && job.Spec.PullPolicy.TimeoutSeconds != nil &&
-		int64(*job.Spec.PullPolicy.TimeoutSeconds) > defaultActiveDeadlineSecondsForNever {
-
-		ret := int64(*job.Spec.PullPolicy.TimeoutSeconds)
-		return utilpointer.Int64(ret)
-	}
 	var ret = defaultActiveDeadlineSecondsForNever
+	if job.Spec.PullPolicy != nil && job.Spec.PullPolicy.TimeoutSeconds != nil {
+		var backoff = 1
+		if job.Spec.PullPolicy.BackoffLimit != nil {
+			backoff = int(*job.Spec.PullPolicy.BackoffLimit)
+		}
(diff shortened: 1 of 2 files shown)


ABNER-1:
/lgtm  
@zmberg, could you please review this?

MajLuu:
@zmberg @FillZpp @veophi , could you please review this?

zmberg:
@MajLuu Currently timeout is pretty clear and we use it heavily internally. Is there an actual scenario on your end that the current logic can't accommodate?

MajLuu:
> @MajLuu Currently timeout is pretty clear and we use it heavily internally. Is there an actual scenario on your end that the current logic can't accommodate?

In our machine learning cluster, we set the pull timeout for a single image to 1 hour, which can be retried 3 times; at the same time, we started many never completionPolicy image pull jobs (10+). It is estimated that the size of our machine learning images (including cuda, pytorch, etc.) is more than 20GB. Most nodes can pull images down after running normally for a period of time, but the newly added nodes in the cluster suddenly received 10+ image pull jobs, which resulted in 1 hour of failure to download the images. At this time, setting the timeout to 3*1h can solve this scenario. In this scenario, we do not think it is reasonable to change the timeout to 3h * 1.

furykerry:
@MajLuu we've added command line argument for controlling concurrent imagepulling workers, maybe setting the worker limits can help increasing the effective image pulling efficiency for newly added nodes. plz check the patch for more detail: https://github.com/openkruise/kruise/commit/318165b7ea75226b71c704c332c92de9fa9dc274

MajLuu:
> @MajLuu we've added command line argument for controlling concurrent imagepulling workers, maybe setting the worker limits can help increasing the effective image pulling efficiency for newly added nodes. plz check the patch for more detail: [318165b](https://github.com/openkruise/kruise/commit/318165b7ea75226b71c704c332c92de9fa9dc274)

OK, we have also implemented the waiting mechanism on the internal 1.4 version, and configured it to pull a maximum of 3 images. At the same time, Containerd has implemented a mechanism to limit the number of concurrent pulling images. However, there is still a problem of excessive disk IO, so this codes was modified. In our machine learning clusters, users can wait for the image to be downloaded; however, there is a strict container startup timeout limit in our flink clusters, and the container image must be guaranteed to exist on the node (there are a large number of pods in the flink nodes, which leads to a large IO utilization rate of the node). The image pull job created is never pullPolicy. If the image cannot be downloaded normally within the specified time, it can only be continued after 24 hours. The problem mentioned in the issue was discovered when troubleshooting the problem of inaccurate download timeouts. If the community considers this modification is unnecessary, please close the issue. Thank you~

PR #2083, opened 2025-06-12
add restart count base in annotation

### LINKED ISSUE #910: [feature request] add inPlaceUpdate restart count to annotation

ABNER-1:
Signed-off-by: Abner-1 <yuanyuxing.yyx@alibaba-inc.com>

Carry on with the work on PR #1097

fixes https://github.com/openkruise/kruise/issues/910

### DIFF
--- a/pkg/controller/daemonset/daemonset_controller.go
+++ b/pkg/controller/daemonset/daemonset_controller.go
@@ -1076,6 +1076,10 @@ func (dsc *ReconcileDaemonSet) refreshUpdateStates(ctx context.Context, ds *apps
 		if res.RefreshErr != nil {
 			klog.ErrorS(res.RefreshErr, "DaemonSet failed to update pod condition for inplace", "daemonSet", klog.KObj(ds), "pod", klog.KObj(pod))
 			return res.RefreshErr
+		} else if opts.CheckPodUpdateCompleted(pod) == nil {
+			//patch restart_cnt to annotation
+			_, _ = dsc.inplaceControl.RefreshRestartCountBaseToPod(pod)
+			inplaceupdate.CalcInplaceUpdateDuration(pod)
 		}
 		if res.DelayDuration != 0 {
 			durationStore.Push(dsKey, res.DelayDuration)
--- a/pkg/util/inplaceupdate/inplace_utils.go
+++ b/pkg/util/inplaceupdate/inplace_utils.go
@@ -1,0 +1,74 @@
+package inplaceupdate
+
+import (
+	"errors"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	appspub "github.com/openkruise/kruise/apis/apps/pub"
+	"github.com/openkruise/kruise/pkg/util"
+	"github.com/openkruise/kruise/pkg/util/metrics/update"
+	"github.com/openkruise/kruise/pkg/util/podadapter"
+)
+
+const AnnotationRestartCountInfoKey = "inplace.apps.kruise.io/restart-count-info"
+
+type RestartCountInfo = map[string]int32
+
+func (c *realControl) RefreshRestartCountBaseToPod(pod *v1.Pod) (gotPod *v1.Pod, err error) {
+	if len(pod.Status.ContainerStatuses) == 0 {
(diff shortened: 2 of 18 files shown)


pkg/util/inplaceupdate/inplace_utils.go
```diff
@@ -0,0 +1,86 @@
...
+	base := &RestartCountInfo{
+		RestartCountBase: make(map[string]int32),
+	}
+	for _, cs := range pod.Status.ContainerStatuses {
+		base.RestartCountBase[cs.Name] = cs.RestartCount
+	}
+
+	info := util.DumpJSON(base)
+	podCopy := pod.DeepCopy()
+	if adp, ok := c.podAdapter.(podadapter.AdapterWithPatch); ok {
```
furykerry, line 38: it seems the only adapter that does not support patch is AdapterInformer, which is not used anywhere in the code other than ut, is it possible to remove AdapterInformer and make patch part of common adapter interface. In this way, it is not necessary to do the checking

pkg/util/inplaceupdate/inplace_utils.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	base := &RestartCountInfo{
+		RestartCountBase: make(map[string]int32),
+	}
+	for _, cs := range pod.Status.ContainerStatuses {
+		base.RestartCountBase[cs.Name] = cs.RestartCount
+	}
+
+	info := util.DumpJSON(base)
+	podCopy := pod.DeepCopy()
```
furykerry: it is not necessary to deepcopy the pod, just initiate an empty pod with the given pod name/namespace

pkg/util/inplaceupdate/inplace_utils.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	state := ParseInPlaceUpdateState(pod)
+	if state == nil {
+		return
+	}
+	inplaceType := update.GetInplaceUpdateType(state.UpdateImages, state.UpdateEnvFromMetadata, state.UpdateResources)
+	fn(inplaceType, int(time.Since(state.UpdateTimestamp.Time).Seconds()))
+}
+
+func ParseInPlaceUpdateState(pod *v1.Pod) *appspub.InPlaceUpdateState {
```
furykerry: it is not necessary to expose this func

pkg/controller/daemonset/daemonset_controller.go
```diff
@@ -1070,6 +1070,10 @@ func (dsc *ReconcileDaemonSet) refreshUpdateStates(ctx context.Context, ds *apps
 		if res.RefreshErr != nil {
 			klog.ErrorS(res.RefreshErr, "DaemonSet failed to update pod condition for inplace", "daemonSet", klog.KObj(ds), "pod", klog.KObj(pod))
 			return res.RefreshErr
+		} else {
+			//patch restart_cnt to annotation
+			_, _ = dsc.inplaceControl.RefreshRestartCountBaseToPod(pod)
```
furykerry, line 1081: is it possible that the inplace update is still progressing here ?

pkg/util/inplaceupdate/inplace_utils.go
```diff
@@ -0,0 +1,86 @@
...
+	appspub "github.com/openkruise/kruise/apis/apps/pub"
+	"github.com/openkruise/kruise/pkg/util"
+	"github.com/openkruise/kruise/pkg/util/metrics/update"
+	"github.com/openkruise/kruise/pkg/util/podadapter"
+)
+
+const AnnotationRestartCountInfoKey = "inplace.apps.kruise.io/restart-count-info"
+
+type RestartCountInfo struct {
+	RestartCountBase map[string]int32 `json:"base"`
```
furykerry: is it possible to use the map directly in the "restart-count-info" without the extra `base` field?

pkg/util/inplaceupdate/inplace_utils.go
ABNER-1: I think the parsing of annotation inplace update state should be wrapped into a function to unify this part of the logic. 

So I'd like to put it in the inplaceupdate package and replace all the other occurrences with this function later.

ABNER-1: I will move it to appspub package.

PR #2087, opened 2025-06-14
fix(cloneset): transition pod lifecycle from PreparingNormal to Normal after scheduling when PreNormal hook is not specified

### LINKED ISSUE #1485: [IMPROVEMENT] [CloneSet] [Lifecycle] Kruise v1.4+ works not so well with kube-scheduler <= v1.21

MichaelRren:

Carry on with the work on PR #1489 

Special thanks to @chenpeicheng9 for laying the groundwork on #1489.
 
This PR refines initial implementation by converting pod to normal state after the pod scheduled since PreNormal Hook is not maintained.

fixes #1485

### DIFF
--- a/pkg/controller/cloneset/sync/cloneset_update.go
+++ b/pkg/controller/cloneset/sync/cloneset_update.go
@@ -170,9 +170,19 @@ func (c *realControl) refreshPodState(cs *appsv1alpha1.CloneSet, coreControl clo
 	var state appspub.LifecycleStateType
 	switch lifecycle.GetPodLifecycleState(pod) {
 	case appspub.LifecycleStatePreparingNormal:
-		if cs.Spec.Lifecycle == nil ||
-			cs.Spec.Lifecycle.PreNormal == nil ||
-			lifecycle.IsPodAllHooked(cs.Spec.Lifecycle.PreNormal, pod) {
+		// 1.the pod lifecycle can be transformed from PreparingNormal to Normal only if Pod is scheduled.
+		// If lifecycle transformed from PreparingNormal to Normal immediately after pod created,
+		// it may trigger a BUG of old version kube-scheduler with some probability.  https://github.com/openkruise/kruise/issues/1485
+		// 2.the semantic of this hook, let's leave it as it is:
+		// PreNormalHook: only if all hooks are hooked (or no hook), the pod can be transformed to Normal
+		var shouldNormal bool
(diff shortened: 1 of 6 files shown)


furykerry:
@MichaelRren can you sign your commit and push again so as to pass the DCO test.

MichaelRren:
> @MichaelRren can you sign your commit and push again so as to pass the DCO test.

@furykerry Sure, plz check it, thanks.

furykerry (approved):
/lgtm
/approve

Issue #2074, opened 2025-06-02
[feature request] Enhance Test Coverage for Kruise


ABNER-1:
**What would you like to be added**:
This task can be broken down into several subtasks, which can be categorized by directory. In other words, if you are interested, you can improve the test coverage starting with the smallest granularity of directories.

**Why is this needed**:
Improving test coverage helps to uncover edge case defects in the main codebase and contributes to the maturation of Kruise.

Prakhar-Shankar:
Hey @ABNER-1, are you working on this issue or can I also help with this?

ABNER-1:
> Hey [@ABNER-1](https://github.com/ABNER-1), are you working on this issue or can I also help with this?

Certainly! You can contribute to any one or multiple directories, and simply link this issue when creating your PR.

PR #2100, opened 2025-07-08
Add unit tests for pkg/daemon.go

GautamBytes:
This PR adds unit tests for the errSignaler utility in pkg/daemon.

In Support of #2074 

These tests cover the core error handling logic, including thread-safety, to ensure the daemon's lifecycle management is reliable.

**Note:** I initially planned to cover the entire package, but the NewDaemon function's use of a global client makes it difficult to mock. Rather than get blocked, this PR adds valuable, targeted tests for the self-contained errSignaler component as a good first step. We can tackle a refactor for broader test coverage in a follow-up.

### DIFF
--- a/pkg/daemon/daemon_test.go
+++ b/pkg/daemon/daemon_test.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2025 The Kruise Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


GautamBytes:
@furykerry  can you rerun the workflow ? the 2 failing workflows seems unrelated to my changes!

PR #2104, opened 2025-07-09
Add unit tests for pkg/util/secret

GautamBytes:
## What this PR does

This PR introduces a unit test suite for the utility functions in `pkg/util/secret/parse.go`, which are responsible for extracting image pull credentials from Kubernetes Secrets.

in support of issue #2074 

## Why is this change necessary?

This part of the codebase handles critical authentication logic but lacked unit tests. Adding tests ensures that the credential lookup, including its fallback behavior, is reliable and can be safely refactored in the future. This change increases test coverage and overall project stability.

## How to test

All new tests pass successfully. This can be verified by running:

```bash
go test ./pkg/util/secret -v

### DIFF
--- a/pkg/util/secret/parse_test.go
+++ b/pkg/util/secret/parse_test.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"flag"
 	"io"
+	"os"
 	"testing"
 
 	daemonutil "github.com/openkruise/kruise/pkg/daemon/util"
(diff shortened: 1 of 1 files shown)


furykerry:
newly added ut fails, plz check it out

GautamBytes:
> newly added ut fails, plz check it out

Sure will fix it by EOD. Btw i have added uts for kruise-tools repo also. Workflows for each ut are passing , would you check them once also?

GautamBytes:
closing this since test file for this is added already during its review time!

PR #2108, opened 2025-07-10
[WIP] feat: introduce progressDeadlineSeconds for CloneSet

### LINKED ISSUE #2006: [feature request] support progressDeadlineSeconds for cloneset
support progressDeadlineSeconds for CloneSet, and add ProgressDeadlineExceeded condition in CloneSet status once the CloneSet failed to make desirable replicas to updated and ready. 

one can refer to the [design docs](https://github.com/openkruise/kruise/blob/master/docs/proposals/20240309-cloneset-support-progressDeadlineSeconds.md)

**Why is this needed**:
help users to understand the progress of CloneSet rollouts.

MichaelRren:
Introduce spec progressDeadlineSeconds 

fix #2006


apis/apps/v1alpha1/cloneset_types.go
```diff
@@ -221,6 +245,8 @@ type CloneSetCondition struct {
 	Type CloneSetConditionType `json:"type"`
 	// Status of the condition, one of True, False, Unknown.
 	Status v1.ConditionStatus `json:"status"`
+	// Last time the condition is updated.
+	LastUpdateTime metav1.Time `json:"lastUpdateTime,omitempty"`
```
furykerry, line 253: what is the difference with LastTransitionTime ? it seems that we can just use  LastTransitionTime for the same use case

MichaelRren: There is a little difference between two fields. LastUpdateTime tracks when the condition or resource was last updated, while LastTransitionTime tracks when the condition last changed its status. For example, during a CloneSetProgressing condition is CloneSetProgressUpdated, if a scale-up event occurs, the condition may be updated (e.g., message or other fields), so the LastUpdateTime changes to now, but the LastTransitionTime remains unchanged.

pkg/controller/cloneset/cloneset_status.go
```diff
@@ -113,4 +120,131 @@ func (r *realStatusUpdater) calculateStatus(cs *appsv1alpha1.CloneSet, newStatus
...
+func getRequeueSecondsFromCondition(condition *appsv1alpha1.CloneSetCondition, pds int32, now time.Time) time.Duration {
+	if condition == nil {
+		return time.Duration(-1)
+	}
+
+	after := condition.LastUpdateTime.Time.Add(time.Duration(pds) * time.Second).Sub(now)
+	if after < time.Second {
+		return time.Duration(0)
+	}
+	return after + time.Second
```
furykerry, line 197: why add 1 second here ?

pkg/controller/cloneset/utils/cloneset_utils.go
```diff
@@ -251,3 +252,123 @@ func DoItSlowly(count int, initialBatchSize int, fn func() error) (int, error) {
...
+	if currentCond != nil && currentCond.Status == condition.Status && currentCond.Reason == condition.Reason {
+		return
+	}
+
+	if currentCond != nil && currentCond.Status == condition.Status {
+		condition.LastTransitionTime = currentCond.LastTransitionTime
+	}
+
+	newConditions := filterOutCondition(status.Conditions, condition.Type)
+	status.Conditions = append(newConditions, condition)
```
furykerry, line 295: will SetCloneSetCondition change the order of conditions?

pkg/controller/cloneset/utils/cloneset_utils.go
```diff
@@ -251,3 +252,123 @@ func DoItSlowly(count int, initialBatchSize int, fn func() error) (int, error) {
...
+	if condition == nil {
+		return true
+	}
+
+	if IsCloneSetResumed(cs) {
+		return true
+	}
+
+	return newStatus.UpdatedReplicas != cs.Status.UpdatedReplicas || // scaling or partition changed.
+		newStatus.ExpectedUpdatedReplicas != cs.Status.ExpectedUpdatedReplicas ||
```
furykerry: does `newStatus.ExpectedUpdatedReplicas != cs.Status.ExpectedUpdatedReplicas` required ? if the cloneset had already been resumed , and no matter how we change the partition, i don't think the cloneset is progressing if no new replicas is created or updated.

pkg/controller/cloneset/cloneset_status.go
```diff
@@ -113,4 +120,131 @@ func (r *realStatusUpdater) calculateStatus(cs *appsv1alpha1.CloneSet, newStatus
...
+
+	duration := r.calculateProgressingStatus(cs, newStatus)
+	clonesetutils.DurationStore.Push(clonesetutils.GetControllerKey(cs), duration)
+}
+
+func (r *realStatusUpdater) calculateProgressingStatus(cs *appsv1alpha1.CloneSet, newStatus *appsv1alpha1.CloneSetStatus) time.Duration {
+	oldStatus := cs.Status
+	cond := clonesetutils.GetCloneSetCondition(oldStatus, appsv1alpha1.CloneSetConditionTypeProgressing)
+	if !clonesetutils.HasProgressDeadline(cs) {
+		if cond != nil {
```
furykerry: if progressdeadline seconds has a default value,  when the code can goes in here ?

pkg/controller/cloneset/cloneset_status.go
```diff
@@ -113,4 +120,131 @@ func (r *realStatusUpdater) calculateStatus(cs *appsv1alpha1.CloneSet, newStatus
...
+
+		case clonesetutils.CloneSetPartitionAvailable(cs, newStatus):
+			klog.V(5).InfoS("CloneSet is partition available", "cloneSet", klog.KObj(cs),
+				"newStatus", newStatus, "oldStatus", oldStatus, "cond", cond)
+
+			msg := "CloneSet has been paused due to partition ready"
+			reason := appsv1alpha1.CloneSetProgressPartitionAvailable
+
+			if clonesetutils.CloneSetAvailable(cs, newStatus) {
+				reason, msg = appsv1alpha1.CloneSetAvailable, "CloneSet is available"
```
furykerry: when cloneset is "available", it is strange that the progressing is still True

pkg/controller/cloneset/cloneset_status.go
MichaelRren: this helps avoid milliseconds skew in AddAfter: https://github.com/kubernetes/kubernetes/issues/39785#issuecomment-279959133

furykerry: interesting catch， can you leave a comment to explain the problem and a link to the kubernete issue

pkg/controller/cloneset/utils/cloneset_utils.go
MichaelRren: the order of filtered condition is changed since it will be appended to the end of conditions

pkg/controller/cloneset/cloneset_status.go
MichaelRren: sure

pkg/controller/cloneset/cloneset_status.go
MichaelRren: removed

furykerry (commented):
/lgtm

pkg/webhook/cloneset/validating/validation.go
```diff
@@ -92,6 +92,13 @@ func (h *CloneSetCreateUpdateHandler) validateCloneSetSpec(spec, oldSpec *appsv1
 	allErrs = append(allErrs, h.validateScaleStrategy(&spec.ScaleStrategy, oldScaleStrategy, metadata, fldPath.Child("scaleStrategy"))...)
 	allErrs = append(allErrs, h.validateUpdateStrategy(&spec.UpdateStrategy, int(*spec.Replicas), fldPath.Child("updateStrategy"))...)
 
+	if spec.ProgressDeadlineSeconds != nil {
+		allErrs = append(allErrs, apivalidation.ValidateNonnegativeField(int64(*spec.ProgressDeadlineSeconds), fldPath.Child("progressDeadlineSeconds"))...)
+		if *spec.ProgressDeadlineSeconds <= spec.MinReadySeconds {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("progressDeadlineSeconds"), spec.ProgressDeadlineSeconds, "must be greater than minReadySeconds"))
```
ABNER-1, line 99: I believe this validation may be inappropriate, or the default value should be removed. My reasons are as follows:

1. Before this feature was introduced, MinReadySeconds might already have been set to a value greater than 600. Setting a default of 600 would trigger the error in this check.
2. I also don’t think it’s appropriate to use MinReadySeconds as the default for ProgressDeadlineSeconds.

As a result, Kruise Rollout could hit this error and cause the release to fail after an upgrade.

MichaelRren: Sure, I have removed default ProgressDeadlineSeconds value

apis/apps/v1alpha1/cloneset_types.go
```diff
@@ -82,6 +82,12 @@ type CloneSetSpec struct {
 
 	// Lifecycle defines the lifecycle hooks for Pods pre-available(pre-normal), pre-delete, in-place update.
 	Lifecycle *appspub.Lifecycle `json:"lifecycle,omitempty"`
+
+	// ProgressDeadlineSeconds specifies the maximum time for the CloneSet to reach available
+	// or paused state due to partitioning. The controller will set a ProgressDeadlineExceeded
+	// condition when timeout occurs, while excluding paused state duration from the deadline calculation.
+	// It continues processing failed CloneSets and defaults to 600 seconds which can be overridden explicitly.
```
furykerry: plz correct the default description, and the docs in the openkruise.io should be also updated

MichaelRren: sure

MichaelRren:
Hi @ABNER-1, plz help review it again

pkg/controller/cloneset/cloneset_status.go
```diff
@@ -113,4 +120,130 @@ func (r *realStatusUpdater) calculateStatus(cs *appsv1alpha1.CloneSet, newStatus
...
+	duration := r.calculateProgressingStatus(cs, newStatus)
+	clonesetutils.DurationStore.Push(clonesetutils.GetControllerKey(cs), duration)
+}
+
+func (r *realStatusUpdater) calculateProgressingStatus(cs *appsv1alpha1.CloneSet, newStatus *appsv1alpha1.CloneSetStatus) time.Duration {
+	if !clonesetutils.HasProgressDeadline(cs) {
+		clonesetutils.RemoveCloneSetCondition(newStatus, appsv1alpha1.CloneSetConditionTypeProgressing)
+		return time.Duration(-1)
+	}
+
```
zmberg, line 133: I think this function should follow terminal state-oriented logic. My suggestions are as follows (you can further optimize the details):
```
switch {
	case clonesetutils.CloneSetAvailable(cs, newStatus):
		condition := clonesetutils.NewCloneSetCondition(appsv1alpha1.CloneSetConditionTypeProgressing,
			v1.ConditionTrue, appsv1alpha1.CloneSetAvailable, "CloneSet is available", timer.Now())
		clonesetutils.SetCloneSetCondition(newStatus, *condition)
		return time.Duration(-1)

	case clonesetutils.CloneSetPaused(cs):
		condition := clonesetutils.NewCloneSetCondition(appsv1alpha1.CloneSetConditionTypeProgressing,
			v1.ConditionTrue, appsv1alpha1.CloneSetProgressPaused, "CloneSet is paused", timer.Now())
		clonesetutils.SetCloneSetCondition(newStatus, *condition)
		return time.Duration(-1)

	case clonesetutils.CloneSetPartitionAvailable(cs, newStatus):
		condition := clonesetutils.NewCloneSetCondition(appsv1alpha1.CloneSetConditionTypeProgressing,
			v1.ConditionTrue, appsv1alpha1.CloneSetProgressPartitionAvailable, "CloneSet has been paused due to partition ready", timer.Now())
		clonesetutils.SetCloneSetCondition(newStatus, *condition)
		return time.Duration(-1)

	case clonesetutils.CloneSetDeadlineExceeded(cs, timer.Now()):
		condition := clonesetutils.NewCloneSetCondition(appsv1alpha1.CloneSetConditionTypeProgressing,
			v1.ConditionTrue, appsv1alpha1.CloneSetProgressDeadlineExceeded, "CloneSet has been paused due to partition ready", timer.Now())
		clonesetutils.SetCloneSetCondition(newStatus, *condition)
		return time.Duration(-1)

	default:
		condition := clonesetutils.NewCloneSetCondition(appsv1alpha1.CloneSetConditionTypeProgressing,
			v1.ConditionTrue, appsv1alpha1.CloneSetProgressUpdated, "CloneSet is progressing", timer.Now())
		clonesetutils.SetCloneSetCondition(newStatus, *condition)
		return time.Duration(-1)
	}

	return time.Duration(-1)
```

pkg/controller/cloneset/utils/cloneset_utils.go
```diff
@@ -251,3 +253,124 @@ func DoItSlowly(count int, initialBatchSize int, fn func() error) (int, error) {
...
+	return !cs.Spec.UpdateStrategy.Paused &&
+		newStatus.ExpectedUpdatedReplicas <= newStatus.UpdatedAvailableReplicas
+}
+
+func CloneSetPaused(cs *appsv1alpha1.CloneSet) bool {
+	cond := GetCloneSetCondition(cs.Status, appsv1alpha1.CloneSetConditionTypeProgressing)
+	return cs.Spec.UpdateStrategy.Paused && (cond == nil || cond.Reason != string(appsv1alpha1.CloneSetProgressPaused))
+}
+
+func CloneSetProgressing(cs *appsv1alpha1.CloneSet, newStatus *appsv1alpha1.CloneSetStatus) bool {
```
zmberg, line 339: remove the function

pkg/controller/cloneset/utils/cloneset_utils.go
```diff
@@ -251,3 +253,124 @@ func DoItSlowly(count int, initialBatchSize int, fn func() error) (int, error) {
...
+}
+
+func CloneSetPartitionAvailable(cs *appsv1alpha1.CloneSet, newStatus *appsv1alpha1.CloneSetStatus) bool {
+	return !cs.Spec.UpdateStrategy.Paused &&
+		newStatus.ExpectedUpdatedReplicas <= newStatus.UpdatedAvailableReplicas
+}
+
+func CloneSetPaused(cs *appsv1alpha1.CloneSet) bool {
+	cond := GetCloneSetCondition(cs.Status, appsv1alpha1.CloneSetConditionTypeProgressing)
+	return cs.Spec.UpdateStrategy.Paused && (cond == nil || cond.Reason != string(appsv1alpha1.CloneSetProgressPaused))
```
zmberg: return cs.Spec.UpdateStrategy.Paused

pkg/controller/cloneset/utils/cloneset_utils.go
```diff
@@ -251,3 +253,124 @@ func DoItSlowly(count int, initialBatchSize int, fn func() error) (int, error) {
...
+	if IsCloneSetResumed(cs) {
+		return true
+	}
+	// a change in ExpectedUpdatedReplicas alone does not trigger a Progress update.
+	return newStatus.UpdatedReplicas != cs.Status.UpdatedReplicas || // scaling or partition changed.
+		newStatus.ReadyReplicas > cs.Status.ReadyReplicas ||
+		newStatus.AvailableReplicas > cs.Status.AvailableReplicas
+}
+
+func IsCloneSetResumed(cs *appsv1alpha1.CloneSet) bool {
```
zmberg: remove the function

pkg/controller/cloneset/utils/cloneset_utils.go
```diff
@@ -251,3 +253,124 @@ func DoItSlowly(count int, initialBatchSize int, fn func() error) (int, error) {
...
+	cond := GetCloneSetCondition(cs.Status, appsv1alpha1.CloneSetConditionTypeProgressing)
+	return !cs.Spec.UpdateStrategy.Paused && (cond != nil && cond.Reason == string(appsv1alpha1.CloneSetProgressPaused))
+}
+
+func CloneSetDeadlineExceeded(cs *appsv1alpha1.CloneSet, now time.Time) bool {
+	condition := GetCloneSetCondition(cs.Status, appsv1alpha1.CloneSetConditionTypeProgressing)
```
zmberg: condition := GetCloneSetCondition(cs.Status, appsv1alpha1.CloneSetConditionTypeProgressing)
from := condition.LastUpdateTime
delta := time.Duration(*cs.Spec.ProgressDeadlineSeconds) * time.Second
return from.Add(delta).Before(now)

MichaelRren: condition might be nil since progressDeadlineSeconds is nil or math.MaxInt32. And if the condition is nil we should not calculate timeout anymore.

And I have modified return statement to: 
return condition.LastUpdateTime.Add(time.Duration(*cs.Spec.ProgressDeadlineSeconds) * time.Second).Before(now)

pkg/controller/cloneset/cloneset_status.go
MichaelRren: I think the judgement clonesetutils.CloneSetProgressing(cs, newStatus) should be retained. Since clonesetutils.CloneSetDeadlineExceeded has a higher priority than default, it might lead to timeout directly without calculating any status?

pkg/controller/cloneset/utils/cloneset_utils.go
MichaelRren: There is a corner case that if the cloneset is already paused and reconciled again, this condition is satisfied and update the condition since LastTransitionTime changed. But actually the condition should not be updated.

pkg/controller/cloneset/cloneset_status.go
MichaelRren: And should available has the higher priority than paused?

pkg/controller/cloneset/utils/cloneset_utils.go
MichaelRren: function SetCloneSetCondition always set a new condition to the `newStatus`, and `newStatus` always doesn't no such condition.

Therefore, the paused condition will always be updated with reconciliation. I suggest to reserve the logic of this function.

PR #2109, opened 2025-07-10
Fix statefulset actions stop when pod is stuck in terminating

derlopes-amzn:

Our setup uses StatefulSets with S3 buckets mounted as volumes. We have a sidecar container that mounts and unmounts these volumes. In a separate issue, this sidecar can fail to unmount the volume, which leaves a pod stuck in terminating (success) state and the pod needs to be force deleted.

We recently migrated to OpenKruise Advanced StatefulSets in order to utilize `maxUnavailable` during deployments. However, when we experienced the aforementioned bug during an update, any pod with higher ordinal would have no further actions taken on it until the pod was force deleted. This is extremely problematic because pods that are deleted from terminating nodes or other rebalancing efforts would never be rescheduled which breaches `maxUnavailable`. As well, this does not mirror how vanilla StatefulSets in k8s behave, clearly indicating a bug in the asts controller. Finally, if we set `maxUnavailable` to 100%, we would see every single pod after the problematic one get deleted during the update and never added back until we force deleted the pod.

After debugging the code, the problem comes from how `processReplica` behaves when a pod has been deleted already but has not completed termination. In OpenKruise, the controller returns `true, false, nil` which signifies that the process `shouldExit` in `runForAllWithBreak`. However, in vanilla k8s, `runForAll` is used for the update process which ignores the `shouldExit` return value if it's not a monotonic update ([ref](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/statefulset/stateful_set_control.go#L495-L508)). This is the main fix for this PR. I have updated `runForAllWithBreak` to ignore `shouldExit` if the update is monotonic.

This bug seems to have been introduced with [this refactor](https://github.com/derlopes-amzn/kruise/blob/bbb2d2695ae33ddf57bf144a634901ad6591c753/pkg/controller/statefulset/stateful_set_control.go#L576-L590). Before the change, if a pod being processed was terminating, it would only exit if the update was monotonic, and it would only break if it breached `maxUnavailable`. To mimic this behavior, I have also added a check for `maxUnavailable` along with my prior mentioned change to only exit when an update is monotonic.

NONE

I can create one for tracking purposes if desired.

To reproduce this issue, you will need to have a pod in an asts get stuck in terminating state and send an update. We accomplish this by mounting a volume using the [AWS CSI S3 Driver](https://raw.githubusercontent.com/kubernetes-sigs/aws-csi-driver-s3/master/deploy/kubernetes/driver.yaml) and not unmounting the volume before terminating. There are potentially other ways you can get a pod stuck in terminating like with custom finalizers, webhooks, or other volume mounts.

Below is the state of the asts that I've used to reproduce, before update:
```
NAME                                       READY   STATUS    RESTARTS   AGE
pdx1-with-s3-stg-0                         5/5     Running   0          19h
pdx1-with-s3-stg-1                         5/5     Running   0          19h
pdx1-with-s3-stg-2                         5/5     Running   0    	19h
pdx1-with-s3-stg-3                         5/5     Running   0          19h
pdx1-with-s3-stg-4                         5/5     Running   0          19h
pdx1-with-s3-stg-5                         5/5     Running   0          19h
pdx1-with-s3-stg-6                         5/5     Running   0          19h
pdx1-with-s3-stg-7                         5/5     Running   0          19h
```

**Without this fix**, forcing `pdx1-with-s3-stg-2` to get stuck, and with 100% deployment velocity, this is the state of the asts ends up in during an update:
```
NAME                                       READY   STATUS        RESTARTS   AGE
pdx1-with-s3-stg-0                         5/5     Running       0          23s
pdx1-with-s3-stg-1                         5/5     Running       0          23s
pdx1-with-s3-stg-2                         0/5     Terminating   0          4m42s
```
Note the lack of pods 3-7. After force deleting pod 2, the rest of the pods return.

**With this fix**, this is the state it ends up in:
```
NAME                                       READY   STATUS        RESTARTS   AGE
pdx1-with-s3-stg-0                         5/5     Running       0          29s
pdx1-with-s3-stg-1                         5/5     Running       0          29s
pdx1-with-s3-stg-2                         0/5     Terminating   0          5m2s
pdx1-with-s3-stg-3                         5/5     Running       0          29s
pdx1-with-s3-stg-4                         5/5     Running       0          29s
pdx1-with-s3-stg-5                         5/5     Running       0          29s
pdx1-with-s3-stg-6                         5/5     Running       0          30s
pdx1-with-s3-stg-7                         5/5     Running 	 0          29s
```
Note that pod 2 is still in terminating but the rest of the pods are updated just fine.

Additionally, when deployment velocity is set to 25% with the same setup, the update gets stuck in this state:
```
NAME                                       READY   STATUS        RESTARTS   AGE
pdx1-with-s3-stg-0                         5/5     Running       0          72s
pdx1-with-s3-stg-1                         5/5     Running       0          2m33s
pdx1-with-s3-stg-2                         0/5     Terminating   0          9m37s
pdx1-with-s3-stg-3                         5/5     Running       0          3m54s
pdx1-with-s3-stg-4                         5/5     Running       0          5m14s
pdx1-with-s3-stg-5                         5/5     Running       0          5m14s
pdx1-with-s3-stg-6                         5/5     Running       0          6m34s
pdx1-with-s3-stg-7                         5/5     Running	 0          6m34s
```
However, **without the fix**, deleting any pod 3 through 7 would cause the pod to not come back online even though it was updated successfully which breaches maxUnavailable. For example, deleting pod 4, 5, and 6 **without the fix** results in:
```
NAME                                       READY   STATUS        RESTARTS   AGE
pdx1-with-s3-stg-0                         5/5     Running       0          72s
pdx1-with-s3-stg-1                         5/5     Running       0          2m33s
pdx1-with-s3-stg-2                         0/5     Terminating   0          9m37s
pdx1-with-s3-stg-3                         5/5     Running       0          3m54s
pdx1-with-s3-stg-7                         5/5     Running	 0          6m34s
```
This is breaching 25% maxUnavailable we set. **With the fix**, the pods come back online normally.

Other methods of fixing this bug would be welcome. This is the best fix I could find.

### DIFF
--- a/pkg/controller/statefulset/stateful_set_control.go
+++ b/pkg/controller/statefulset/stateful_set_control.go
@@ -475,7 +475,7 @@ func (ssc *defaultStatefulSetControl) updateStatefulSet(
 	processReplicaFn := func(i int) (bool, bool, error) {
 		return ssc.processReplica(ctx, set, updateSet, monotonic, replicas, i, &status, scaleMaxUnavailable)
 	}
-	if shouldExit, err := runForAllWithBreak(replicas, processReplicaFn); shouldExit || err != nil {
+	if shouldExit, err := runForAllWithBreak(replicas, processReplicaFn, monotonic); shouldExit || err != nil {
 		ssc.updatePVCStatus(&status, set, replicas)
 		updateStatus(&status, minReadySeconds, currentRevision, updateRevision, replicas, condemned)
 		return &status, err
@@ -1097,12 +1097,16 @@ func (ssc *defaultStatefulSetControl) processReplica(
 	// container, not by the reason for pod termination. We should restart the pod
 	// regardless of the exit code.
 	if isFailed(replicas[i]) || isSucceeded(replicas[i]) {
+		if decreaseAndCheckMaxUnavailable(scaleMaxUnavailable) {
+			// We must wait for pods to finish terminating or we violate max scale
+			return false, true, nil
+		}
 		if replicas[i].DeletionTimestamp == nil {
 			if _, _, err := ssc.deletePod(set, replicas[i]); err != nil {
 				return true, false, err
 			}
 		}
-		// New pod should be generated on the next sync after the current pod is removed from etcd.
+		// Continue with next pods (unless monotonic)
 		return true, false, nil
 	}
 	// If we find a Pod that has not been created we create the Pod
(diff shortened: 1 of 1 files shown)


furykerry:
@derlopes-amzn thanks for your in-depth analysis of the problem and the fix, can you sign your commits?

derlopes-amzn:
@furykerry Thanks for the quick turnaround. Could you estimate the approximate timeline for this change to hit the latest helm chart?

derlopes-amzn:
@furykerry @zmberg Could you give me an idea of how long this change will take to release to helm charts? Trying to see if we need to deploy a custom image in the mean time.

furykerry:
@derlopes-amzn we plan to release an alpha version of release 1.9 in the end of Aug with k8s 1.32 support

PR #2111, opened 2025-07-15
Fix: non-percentage string value for maxSurge / maxUnavailable of CloneSet UpdateStrategy will be validated properly

AiRanthem:
numerical maxUnavailable / maxSurge string values like "5" will be rejected by validation webhook


furykerry (approved):
/lgtm
/approve

PR #2127, opened 2025-07-27
Fix webhook server binding to all zero address security issue

### LINKED ISSUE #1916: [BUG] About all zero monitoring and insecure protocols

Parship999:
- Webhook server binds to `0.0.0.0` on port `9876` (and potentially other ports like `8080`, `8090`, `8000`)
- so binding to all zero addresses expose the service to all network interfaces, which can be a security concern in multi network environments
- so as a solution I replace the hardcoded `"0.0.0.0"` webhook host binding with `ResolveBindAddress(nil)`

- fixes #1916

- Build anbd run:
```
go build -o kruise-manager main.go
./kruise-manager
```
- check logs:
`INFO resolved webhook bind address {"address": "172.25.113.206"}`

- Verify webhook no longer binds to 0.0.0.0:
```
netstat -tlnp | grep :9876
# Should show specific IP instead of 0.0.0.0:9876
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -28,6 +28,7 @@ import (
 	"github.com/spf13/pflag"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	utilnet "k8s.io/apimachinery/pkg/util/net"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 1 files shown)


Parship999:
@furykerry Can you please review this?

PR #2154, opened 2025-08-07
 Fix the panic issue when the `--namespace` flag is set

### LINKED ISSUE #1764: [BUG] namespaced kruise-controller panics
**What happened**:
If `--namespace` flag is set (e.g., --namespace=kruise-system), the controller of v1.7.x panics on startup with the following stacktrace:
```
I0927 15:42:21.569651       1 reflector.go:325] Listing and watching *v1.ControllerRevision from pkg/mod/k8s.io/client-go@v0.28.9/tools/cache/reflector.go:229
I0927 15:42:21.670294       1 shared_informer.go:341] caches populated
panic: interface conversion: *cache.multiNamespaceInformer is not cache.SharedIndexInformer: missing method GetController

goroutine 176 [running]:
github.com/openkruise/kruise/pkg/controller/daemonset.newReconciler({0x2c5d3c8, 0xc000129520})
	/workspace/pkg/controller/daemonset/daemonset_controller.go:175 +0x46a
github.com/openkruise/kruise/pkg/controller/daemonset.Add({0x2c5d3c8, 0xc000129520})
	/workspace/pkg/controller/daemonset/daemonset_controller.go:141 +0x13a
github.com/openkruise/kruise/pkg/controller.SetupWithManager({0x2c5d3c8, 0xc000129520})
	/workspace/pkg/controller/controllers.go:74 +0x82
main.main.func2()
	/workspace/main.go:252 +0x12d
created by main.main
	/workspace/main.go:244 +0x142a
```

**How to reproduce it (as minimally and precisely as possible)**: see above

**Environment**:
- Kruise version: v1.7.x
- Kubernetes version (use `kubectl version`): 1.27
- Others:

Parship999:
The panic occurs because when a namespace is specified in the manager configuration (via `--namespace` flag), the controller-runtime cache returns a `multiNamespaceInformer` instead of a `SharedIndexInformer`. The `multiNamespaceInformer` does not implement the `GetController` method that's part of the `SharedIndexInformer` interface, it causses the type assertion to fail with a panic.

The issue was in two controllers:
```
pkg/controller/daemonset/daemonset_controller.go
pkg/controller/statefulset/statefulset_controller.go
```

I implemented proper type checking before the type assertion, following the pattern used in `pkg/util/client/no_deepcopy_lister.go`

- fixes #1764

- both controllers compile successfully
- DaemonSet controller tests pass
- main binary builds successfully

### DIFF
--- a/pkg/controller/daemonset/daemonset_controller.go
+++ b/pkg/controller/daemonset/daemonset_controller.go
@@ -172,10 +172,27 @@ func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
 		return nil, err
 	}
 
-	dsLister := kruiseappslisters.NewDaemonSetLister(dsInformer.(cache.SharedIndexInformer).GetIndexer())
-	historyLister := appslisters.NewControllerRevisionLister(revInformer.(cache.SharedIndexInformer).GetIndexer())
-	podLister := corelisters.NewPodLister(podInformer.(cache.SharedIndexInformer).GetIndexer())
-	nodeLister := corelisters.NewNodeLister(nodeInformer.(cache.SharedIndexInformer).GetIndexer())
+	dsSharedInformer, ok := dsInformer.(cache.SharedIndexInformer)
+	if !ok {
+		return nil, fmt.Errorf("dsInformer %T from cache is not a SharedIndexInformer", dsInformer)
+	}
+	revSharedInformer, ok := revInformer.(cache.SharedIndexInformer)
+	if !ok {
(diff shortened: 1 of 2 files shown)


furykerry:
sorry for the late reply, i am not sure whether the daemonset or statefulset controller can still works in using --namespace after this patch. The root cause of why previous code try to do the type assertion is that the  controller is not using the controller-runtime informer. Maybe a better solution is to switch from the lister to controller-runtime client just as other controller.

ghost:
@furykerry for daemonset I switched reads from listers to controller runtime client. Can you please review it?

PR #2156, opened 2025-08-11
Fix bug that caused the upgrade to block during the sidecarSet hot upgrade due to corner case.

zmberg:


furykerry (approved):
/lgtm
/approve

PR #2219, opened 2025-10-27
expose defaultTTlSecondsForAlways for nodeimage

### LINKED ISSUE #2213: [BUG] setting ttlSecondsAfterFinished in AdvancedCronJob will cause the job run again and again
**What happened**:
when specify the completionPolicy.ttlSecondsAfterFinished in AdvancedCronJob, the broadcastjob will auto be deleted after timeout, but then it will be scheduled again and again

job should not be started again

1. create AdvancedCronJob which should run at a fix time
```
kind: AdvancedCronJob
metadata:
    name: acj-job
    namespace: default
spec:
    concurrencyPolicy: Replace
    failedJobsHistoryLimit: 5
    paused: false
    timeZone: Asia/Shanghai
    schedule: "46 21 * * *"
    successfulJobsHistoryLimit: 3
    template:
      broadcastJobTemplate:
        spec:
         template:
          spec:
           containers:
           - name: test
             image: alpine
             command: ["sleep", "10"]
           restartPolicy: Never
         completionPolicy:
          type: Always
          ttlSecondsAfterFinished: 30

```
2. watch kubectl get broadcastjob
3. the job will deleted after 40 seconds, then it will ber started again

1. actually when I manually delete the broadcastjob, it also restart
2. the root cause is watchBroadcastJob() method is watching all broadcasejob resource change event, include delete, and will recreate it in the reconcileBroadcastJob() as a missing job
3. this issue also impact the newly added feature #2207

**Environment**:
- Kruise version:
- Others:

### LINKED ISSUE #2214: [BUG] expose the default ttlSecondsAfterFinished as a parameter for kruise-controller-manager
**What happened**:
Currently when set completionPolicy.ttlSecondsAfterFinished in ImagePullListJob, both job and tags in nodeimage are deleted when timeout.
However with feature #2207 introduced, we can schedule job with AdvancedCronJob now, the ImagePullListJob history should be maintained by successfulJobsHistoryLimit and failedJobsHistoryLimit.
So if user also set the ttlSecondsAfterFinished, the job would also be deleted.

User should not set completionPolicy.ttlSecondsAfterFinished, but still have ability to control the ttl for nodeimage data, as it would be too large if not clean up.

1. create AdvancedCronJob with imageListPullJobTemplate
```
kind: AdvancedCronJob
metadata:
    name: acj-test
    namespace: default
spec:
    concurrencyPolicy: Replace
    paused: false
    schedule: "*/2 * * * *"
    successfulJobsHistoryLimit: 3
    failedJobsHistoryLimit: 5
    template:
      imageListPullJobTemplate:
        spec:
          completionPolicy:
            ttlSecondsAfterFinished: 30
            type: Always
          imagePullPolicy: IfNotPresent
          images:
          - busybox
          parallelism: 2
          pullPolicy:
            backoffLimit: 1
            timeoutSeconds: 10
    timeZone: Asia/Shanghai
```
2. only data in nodeimage should be deleted after 30 seconds

**Environment**:
- Kruise version: master branch
- Others:

bitsf:
expose defaultTTlSecondsForAlways as a environment for nodeimage, that can allow user to controller the default ttl for nodeimage status, in case user don't want to use the same ttl for imagepulljob

fix #2214
fix #2213

1. kubectl edit deploy kruise-controller-manager, add parameter `--default-ttlseconds-for-always-nodeimage` for container manager
2. create imagepulljob
3. check the nodeimage ttlSecondsAfterFinished value
4. for detailed calculation methods, please refer to the test case

### DIFF
--- a/main.go
+++ b/main.go
@@ -62,11 +62,12 @@ import (
 )
 
 const (
-	defaultLeaseDuration              = 15 * time.Second
-	defaultRenewDeadline              = 10 * time.Second
-	defaultRetryPeriod                = 2 * time.Second
-	defaultControllerCacheSyncTimeout = 2 * time.Minute
-	defaultWebhookInitializeTimeout   = 60 * time.Second
+	defaultLeaseDuration                     = 15 * time.Second
+	defaultRenewDeadline                     = 10 * time.Second
+	defaultRetryPeriod                       = 2 * time.Second
+	defaultControllerCacheSyncTimeout        = 2 * time.Minute
+	defaultWebhookInitializeTimeout          = 60 * time.Second
+	defaultTtlsecondsForAlwaysNodeimageConst = 300
 )
 
 var (
@@ -103,6 +104,7 @@ func main() {
 	var retryPeriod time.Duration
 	var controllerCacheSyncTimeout time.Duration
 	var webhookInitializeTimeout time.Duration
+	var defaultTtlsecondsForAlwaysNodeimage int
 
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&healthProbeAddr, "health-probe-addr", ":8000", "The address the healthz/readyz endpoint binds to.")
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -128,6 +130,7 @@ func main() {
 		"leader-election-retry-period is the duration the LeaderElector clients should wait between tries of actions. Default is 2 seconds.")
 	flag.DurationVar(&controllerCacheSyncTimeout, "controller-cache-sync-timeout", defaultControllerCacheSyncTimeout, "CacheSyncTimeout refers to the time limit set to wait for syncing caches. Defaults to 2 minutes if not set.")
 	flag.DurationVar(&webhookInitializeTimeout, "webhook-initialize-timeout", defaultWebhookInitializeTimeout, "WebhookInitializeTimeout refers to the time limit set to wait for webhook initialization. Defaults to 60 seconds if not set.")
+	flag.IntVar(&defaultTtlForAlwaysNodeimage, "default-ttl-for-always-nodeimage", defaultTtlForAlwaysNodeimageConst, "DefaultTtlForAlwaysNodeimage refers to the calculation of the time limit the lifetime of a pulling task that has finished execution. Defaults to 300 seconds if not set.")
```
furykerry: if value is not time.Duration, then we should be more explicit that what unit default-ttl-for-always-nodeimage use. Consider rename it to default-ttlseconds-for-always-nodeimage

PR #2223, opened 2025-11-01
Fix: webhook TLS and remove deprecated TLS fields

### LINKED ISSUE #1916: [BUG] About all zero monitoring and insecure protocols

sonalgaud12:

Hardens webhook TLS and secures internal endpoints.

Enforces TLS 1.2+ and modern cipher suites

Removes deprecated TLS options

Keeps webhook port 9876

Binds metrics/health/pprof to 127.0.0.1

Fixes #1916

Check that only TLS 1.2+ is accepted on port 9876 and no endpoints bind to 0.0.0.0.

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,11 +19,13 @@ package main
 import (
 	"crypto/tls"
 	"flag"
+	"math/rand"
+	"net"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
 	"time"
-	_ "time/tzdata" // for AdvancedCronJob TZ support
+	_ "time/tzdata" // for AdvancedCronJob Time Zone support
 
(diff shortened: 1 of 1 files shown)


sonalgaud12:
@furykerry I have tried to resolve the issue as per my understanding. Can you review it once and let me know the needful?

PR #2270, opened 2025-12-19
Webhook controller merges namespace selectors added by cloud providers

### LINKED ISSUE #2220: [BUG] OpenKruise webhook controller should NOT overwrite namespaceSelectors added by other actors
**What happened**:
We are deploying OpenKruise to Azure Kubernetes Service (AKS) clusters with Helm chart ([v1.7.2](https://github.com/openkruise/charts/blob/master/versions/kruise/1.7.2/templates/webhookconfiguration.yaml)). After the kruise-controller-manager pod starts up, we observed frequent conflict errors in updating the Kruise ValidatingWebhookConfiguration and MutatingWebhookConfiguration. This unexpected behavior leads to large volume of requests from  kruise-controller-manager to API Server (even after the built-in client-go throttling).

OpenKruise’s webhook controller stores a copy of the webhook configuration as template in annotations on its first syn. After that, it reconciles the webhook configuration [using the cached template](https://github.com/openkruise/kruise/blob/release-1.7/pkg/webhook/util/configuration/configuration.go).

```go
if templateStr := mutatingConfig.Annotations["template"]; len(templateStr) > 0 {
var mutatingWHs []admissionregistrationv1.MutatingWebhook
if err := json.Unmarshal([]byte(templateStr), &mutatingWHs); err != nil {
return nil, err
}
return mutatingWHs, nil
}
```
This template (based on [webhookconfiguration.yaml](https://github.com/openkruise/charts/blob/master/versions/kruise/next/templates/webhookconfiguration.yaml)) sets empty namespaceSelectors on some webhooks and the minimal namespace selector (exclude kube-system) on others.

```yaml
    namespaceSelector:
      matchExpressions:
        - key: kubernetes.io/metadata.name
          operator: NotIn
          values:
            - kube-system
```

AKS has an admissions enforcer, which automatically excludes kube-system and AKS internal namespaces by patching the namespaceSelector of all webhooks to prevent webhooks from running in critical namespaces. Its reconciliation loop patches each webhook's namespaceSelector with two matchExpressions:

```yaml
        - key: control-plane
          operator: NotIn
          values:
            - "true"
        - key: kubernetes.azure.com/managedby
          operator: NotIn
          values:
            - aks
```

OpenKruise webhook controller should NOT overwrite namespaceSelectors added by other actors (such as AKS in this case).

1. Deploy kruise-manger to an AKS cluster
2. Run `kubectl get mutatingwebhookconfiguration kruise-mutating-webhook-configuration  -n kruise-system -o yaml` multiple times (5-10) to observer the flipping of namespaceSelector values in the mutating webhook configuration
3. Run ` kubectl get validatingwebhookconfiguration kruise-validating-webhook-configuration -n kruise-system -o yaml` multiple times (5-10) to observer the flipping of namespaceSelector values in the validating webhook configuration

I filed [a separate issue](https://github.com/openkruise/charts/issues/155) in the openkruise/charts repo for the same problem. However, a more durable fix should be changing the behavior of webhook controller to not overwrite namespaceSelector changes made by others. 

**Environment**:
- Kruise version: v1.7.2
- Kubernetes version (use `kubectl version`): v1.34.1
   ```
   helm install kruise-manager openkruise/kruise \
	--wait \
	--timeout 5m \
	--set image.repository=mcr.microsoft.com/oss/v2/openkruise/kruise-manager,image.tag=v1.7.2,image.pullPolicy=IfNotPresent,manager.replicas=1
   ```
- Others: None

peng2x:

Webhook controller merges base namespace selectors and the ones added by cloud providers

Fix #2220

### DIFF
--- a/pkg/webhook/util/configuration/configuration.go
+++ b/pkg/webhook/util/configuration/configuration.go
@@ -22,9 +22,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
-	"reflect"
+	"sort"
 
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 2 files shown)


peng2x:
I merged a simpler fix in the OpenKruise charts repository that resolves the same issue:
https://github.com/openkruise/charts/pull/157

While this draft PR technically works and passes all tests, it may be more difficult to maintain in the long term—especially if the Mutating/Validating WebhookConfiguration schema changes in the future.

Therefore, I’m closing this PR in favor of the fix implemented in the charts repository.

PR #2274, opened 2025-12-24
Fix critical race in InPlaceUpdateReady condition update that can permanently stall AdvancedStatefulSet in-place env updates

### LINKED ISSUE #2269: [BUG] AdvancedStatefulSet stuck in PreparingUpdate when using lifecycle.inPlaceUpdate.labelsHandler with InPlaceUpdateEnvFromMetadata (v1.8.2)
## Description

We are facing an issue in **Kruise v1.8.2** where an **AdvancedStatefulSet (ASTS)** gets permanently stuck in:

```
lifecycle.apps.kruise.io/state: PreparingUpdate
```

The update never proceeds to `Updating`, even though:

* the Pod is `Ready`
* `InPlaceUpdateReady=True`
* container has already restarted
* `env` from metadata has been refreshed

This happens when **all of the following are used together**:

* `lifecycle.inPlaceUpdate.labelsHandler`
* `podUpdatePolicy: InPlaceIfPossible`
* env updated via `fieldRef` from Pod annotations (`InPlaceUpdateEnvFromMetadata`)

---

## Environment

* **Kruise version**: v1.8.2
* **Kubernetes version**: 1.24.x
* **Workload**: AdvancedStatefulSet
* **Namespace**: `test`

---

## StatefulSet manifest (simplified & reproducible)

```yaml
kind: StatefulSet
metadata:
  name: asts-inplace-env-demo
  namespace: test
spec:
  replicas: 1
  serviceName: asts-inplace-env-demo
  selector:
    matchLabels:
      app: asts-inplace-env-demo
  lifecycle:
    inPlaceUpdate:
      labelsHandler:
        example.com/connect-drain-block: "true"
  updateStrategy:
    type: RollingUpdate
    rollingUpdate:
      podUpdatePolicy: InPlaceIfPossible
      maxUnavailable: 1
  template:
    metadata:
      labels:
        app: asts-inplace-env-demo
        example.com/connect-drain-enabled: "true"
      annotations:
        example.com/instance-config-version: v3
    spec:
      readinessGates:
        - conditionType: InPlaceUpdateReady
      containers:
        - name: main
          image: busybox
          command: [ "sh", "-c", "sleep infinity" ]
          env:
            - name: INSTANCE_CONFIG_VERSION
              valueFrom:
                fieldRef:
                  apiVersion: v1
                  fieldPath: metadata.annotations['example.com/instance-config-version']
...
```

---

## Pod status when stuck

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: asts-inplace-env-demo-0
  namespace: test
  labels:
    app: asts-inplace-env-demo
    example.com/connect-drain-block: "false"
    example.com/connect-drain-enabled: "false"
    lifecycle.apps.kruise.io/state: PreparingUpdate
  annotations:
    apps.kruise.io/inplace-update-state: |
      {
        "revision":"asts-inplace-env-demo-564cdf7ddb",
        "updateEnvFromMetadata":true
      }
    apps.kruise.io/runtime-containers-meta: |
      {
        "containers":[
          {
            "name":"main",
            "hashes":{
              "extractedEnvFromMetadataHash":2499068171
            }
          }
        ]
      }
spec:
  containers:
    - name: main
      env:
        - name: INSTANCE_CONFIG_VERSION
          valueFrom:
            fieldRef:
              fieldPath: metadata.annotations['example.com/instance-config-version']
status:
  conditions:
    - type: Ready
      status: "True"
    - type: InPlaceUpdateReady
      status: "True"
```

---

## StatefulSet status

```yaml
status:
  replicas: 1
  readyReplicas: 1
  updatedReplicas: 0
  currentRevision: asts-inplace-env-demo-564cdf7ddb
  updateRevision: asts-inplace-env-demo-7c7c67b649
```

---

## Controller logs

```text
StatefulSet has unhealthy Pods
StatefulSet is waiting for Pod to be Available
updatedReplicas=0
```

---

## Expected behavior

After the external controller removes the `labelsHandler` label (`example.com/connect-drain-block=true`), the Pod should:

1. move from `PreparingUpdate` → `Updating`
2. perform in-place update
3. finally reach `Updated` / `Normal`
4. increase `updatedReplicas`

---

## Actual behavior

The Pod remains **forever in `PreparingUpdate`**, and:

* never enters `Updating`
* `updatedReplicas` stays `0`
* ASTS rollout is permanently blocked

---

## Analysis
This looks like a **state machine deadlock** when combining:

* `lifecycle.inPlaceUpdate.labelsHandler`
* `InPlaceUpdateEnvFromMetadata`
* AdvancedStatefulSet

ktalg:

**Fix a potential deadlock where AdvancedStatefulSet in-place update can get stuck forever in `lifecycle.apps.kruise.io/state=Updating` (and Pod `InPlaceUpdateReady=True`) when `InPlaceUpdateEnvFromMetadata` is used together with `lifecycle.inPlaceUpdate.labelsHandler`.**

In Kruise v1.8.2, the in-place update flow updates the Pod readiness-gate condition `InPlaceUpdateReady` and the in-place update state via different code paths:

* `Update()` sets `InPlaceUpdateReady=False` first (via `UpdatePodStatus`), then updates the Pod spec (image/env-from-metadata related fields).
* `Refresh()` periodically checks whether the in-place update has completed (`CheckContainersUpdateCompleted`). If it *thinks* the update is completed, it sets `InPlaceUpdateReady=True`.

The issue happens under conflict/race scenarios (commonly triggered when lifecycle hooks and external controllers are involved):

1. `Update()` starts in-place update and intends to set `InPlaceUpdateReady=False`, but there is a short window where the Pod **spec update has not been applied yet**.
2. `Refresh()` can run in that window and evaluate completion against an **older Pod object**. In some cases `CheckContainersUpdateCompleted()` returns success (false positive), and `Refresh()` tries to set `InPlaceUpdateReady=True`.
3. The current `updateCondition()` implementation ignores conflicts by **fetching the latest Pod and retrying**. When a `409 Conflict` occurs, it may end up applying `InPlaceUpdateReady=True` onto the **newer Pod version**, even though the newer Pod is actually still updating (env hash mismatch, etc.).
```
I1224 09:01:38.587401      24 inplace_update.go:173] >>> update pod cond, name: asts-inplace-env-demo-0, old ver: 608347, new ver: 608347<<<
I1224 09:01:38.587943      24 no_deepcopy_lister.go:110] Listed apps.kruise.io/v1alpha1 ImagePullJob objects 0 without DeepCopy, cost 17.879µs
I1224 09:01:38.596579      24 round_trippers.go:454] PUT https://10.96.0.1:443/api/v1/namespaces/aa/pods/asts-inplace-env-demo-0/status 409 Conflict in 8 milliseconds
I1224 09:01:38.604667      24 round_trippers.go:454] POST https://10.96.0.1:443/api/v1/namespaces/aa/events 201 Created in 15 milliseconds
I1224 09:01:38.607833      24 inplace_update.go:173] >>> update pod cond, name: asts-inplace-env-demo-0, old ver: 608347, new ver: 608349<<<
I1224 09:01:38.625273      24 round_trippers.go:454] PUT https://10.96.0.1:443/api/v1/namespaces/aa/pods/asts-inplace-env-demo-0/status 409 Conflict in 17 milliseconds
I1224 09:01:38.679523      24 inplace_update.go:173] >>> update pod cond, name: asts-inplace-env-demo-0, old ver: 608347, new ver: 608351<<<
I1224 09:01:38.687926      24 round_trippers.go:454] PUT https://10.96.0.1:443/api/v1/namespaces/aa/pods/asts-inplace-env-demo-0/status 200 OK in 8 milliseconds
```

4. Once `InPlaceUpdateReady` is incorrectly flipped to `True`, Kruise-daemon will not trigger the required container restart for `env-from-metadata` hash reconciliation (daemon only restarts when `InPlaceUpdateReady=False`). As a result, the manager keeps observing hash mismatch and the rollout stays stuck at `Updating` indefinitely.

This PR introduces a version-sensitive status update for the “mark completed” path in `Refresh()`:

* when `Refresh()` wants to set `InPlaceUpdateReady=True` (i.e., when it believes the update is completed), so we **only mark completion for the exact Pod version we validated**.
* Keep the retry-on-conflict behavior for the `InPlaceUpdateReady=False` path (start of update), which is safe and desirable.

Net effect: avoid applying a “completed” condition to a newer Pod revision that has not actually completed, preventing the stuck `Updating` state.

---

fixes #2269

---

**Reproduction (before this PR):**

1. Install Kruise v1.8.2 and enable `InPlaceUpdateEnvFromMetadata`.
2. Create an AdvancedStatefulSet with:

   * `lifecycle.inPlaceUpdate.labelsHandler` (e.g., `example.com/block: "true"`)
   * Pod env referencing metadata annotation via `fieldRef`, e.g.:

     ```yaml
     - name: INSTANCE_CONFIG_VERSION
       valueFrom:
         fieldRef:
           apiVersion: v1
           fieldPath: metadata.annotations['example.com/version']
     ```
   * Add readinessGate `InPlaceUpdateReady`.
3. Update the referenced annotation in the workload template to trigger in-place env update.
4. Let an external controller remove the lifecycle label to unblock the hook.

**Observed behavior:**

* Rollout can get stuck with:

  * Pod annotation `lifecycle.apps.kruise.io/state: Updating`
  * Pod condition `InPlaceUpdateReady: True`
  * Kruise-controller keeps logging env hash mismatch and “unhealthy/unavailable” pod blocks further updates.
* Manager logs show `PUT .../status 409 Conflict` around the time `InPlaceUpdateReady` flips to True.

**Verification (after this PR):**

Repeat the same steps and confirm:

* If a conflict happens when `Refresh()` tries to set `InPlaceUpdateReady=True`, the update is rejected for that version and retried later with a re-check (instead of incorrectly marking the newer Pod as completed).
* The Pod condition stays `InPlaceUpdateReady=False` until env hash is truly consistent and the in-place update is actually finished.
* The rollout proceeds and eventually reaches `Updated` / stable state without being stuck at `Updating`.

---

* This change is intentionally **narrow**: it only makes the “mark update completed” (`InPlaceUpdateReady=True`) status update **version-safe**. The “start update” path keeps the original retry-on-conflict behavior.
* The core intent is to preserve correctness under concurrent reconciles / external controllers: **only mark completion for the exact Pod object that passed the completion check**.
* Behavior change: `Refresh()` may return conflict and retry later, which is expected and preferable to incorrectly setting `InPlaceUpdateReady=True` and deadlocking the rollout.

### DIFF
--- a/pkg/util/inplaceupdate/inplace_update.go
+++ b/pkg/util/inplaceupdate/inplace_update.go
@@ -169,28 +169,26 @@ func (c *realControl) Refresh(pod *v1.Pod, opts *UpdateOptions) RefreshResult {
 		Status:             v1.ConditionTrue,
 		LastTransitionTime: metav1.NewTime(Clock.Now()),
 	}
+	// Do not retry on conflict; only update the condition for the checked Pod version.
+	// see https://github.com/openkruise/kruise/pull/2274
 	err := c.updateCondition(pod, newCondition)
 	return RefreshResult{RefreshErr: err}
 }
 
+// updateCondition update the given Pod's condition by updating status on a copy with the same ResourceVersion;
+// no conflict retry or refetch is performed, so it only applies to this specific Pod version.
 func (c *realControl) updateCondition(pod *v1.Pod, condition v1.PodCondition) error {
-	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		clone, err := c.podAdapter.GetPod(pod.Namespace, pod.Name)
-		if err != nil {
-			return err
-		}
+	clone := pod.DeepCopy()
+	if hasEqualCondition(clone, &condition) {
+		return nil
+	}
 
-		if hasEqualCondition(clone, &condition) {
-			return nil
-		}
-
-		util.SetPodCondition(clone, condition)
-		// We only update the ready condition to False, and let Kubelet update it to True
-		if condition.Status == v1.ConditionFalse {
-			util.SetPodReadyCondition(clone)
-		}
-		return c.podAdapter.UpdatePodStatus(clone)
-	})
+	util.SetPodCondition(clone, condition)
+	// We only update the ready condition to False, and let Kubelet update it to True
+	if condition.Status == v1.ConditionFalse {
+		util.SetPodReadyCondition(clone)
+	}
+	return c.podAdapter.UpdatePodStatus(clone)
 }
 
 func (c *realControl) finishGracePeriod(pod *v1.Pod, opts *UpdateOptions) (time.Duration, error) {
@@ -332,7 +330,15 @@ func (c *realControl) Update(pod *v1.Pod, oldRevision, newRevision *apps.Control
 			Status:             v1.ConditionFalse,
 			Reason:             "StartInPlaceUpdate",
 		}
-		if err := c.updateCondition(pod, newCondition); err != nil {
+		// ensure the condition is updated with conflict retry
+		err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+			clone, err := c.podAdapter.GetPod(pod.Namespace, pod.Name)
+			if err != nil {
+				return err
+			}
+			return c.updateCondition(clone, newCondition)
+		})
+		if err != nil {
 			return UpdateResult{InPlaceUpdate: true, UpdateErr: err}
 		}
 	}


ktalg:
There are two key points:

1. In the scenario of **AdvancedStatefulSet + in-place env update + lifecycle hook**, the controller may perform additional reconciliations during the transition from *PreparingUpdate* to *Updating*. Regardless of the external trigger or reconciliation timing, `Refresh()` must behave deterministically: it should evaluate completion and update `InPlaceUpdateReady` **consistently against the same, specific Pod version** that was validated. In particular, it must not mark `InPlaceUpdateReady=True` based on a stale view while applying that result to a newer Pod revision. Ensuring this version-consistent “check-then-update” behavior is necessary for correctness.

2. To preserve compatibility, this PR follows a minimal-change approach. The adjustment is the smallest effective change that resolves the issue without altering the overall semantics of the in-place update flow.

gaopeiliang:
I have find in version 1.7.2 ;  should release 1.8.3 ?

PR #2277, opened 2025-12-30
Fixes a timing/cache inconsistency issue in the inplace update

### LINKED ISSUE #2276: [BUG] Unexpected traffic during Pod in-place upgrade
# Bug: InPlaceUpdate timing + cache inconsistency causes `InPlaceUpdateReady=False` drain to be skipped, leading to unexpected traffic during Pod in-place upgrade

## Background

In the current CloneSet InPlaceUpdate implementation, the controller performs the following actions in each reconcile loop (conceptually repeating in order `1 -> 2 -> 3 -> ...`):

1. **Update Pod Status**: set `InPlaceUpdateReady=true` for Pods considered reaching the desired terminal version state
2. **Update Pod Status**: set `InPlaceUpdateReady=false` for Pods that are going to be in-place upgraded (to drain traffic)
3. **Update Pod Spec**: update image/fields to trigger the in-place upgrade (via the **Update** API)

## Problem timeline

The following sequence was observed (key point is step 4):

1. **Update Pod Status succeeds**: `InPlaceUpdateReady=false`, Pod is drained (still serving the old version normally)
2. **Update Pod Spec fails**: because updating spec uses **Update**, it is subject to resourceVersion checks (ABA/rv conflict). At this moment, the controller cache has not yet reflected the status change from step 1, so the controller updates with a stale Pod object. Even with backoff retries, the update keeps failing due to slow cache sync.
3. **Update Pod Status succeeds**: since spec update failed, the Pod is still on the old version and is treated as “old version not updated”, so the controller sets `InPlaceUpdateReady=true` again (traffic resumes)
4. **Patch Pod Status is skipped (BUG)**: the controller cache still shows the old value from step 2 (`InPlaceUpdateReady=false`), so the drain patch logic is skipped, while the actual Pod is already `InPlaceUpdateReady=true`
5. Later **Update Pod Spec succeeds**, the in-place upgrade proceeds; around the old container exit / new container not-ready window, there can be an **unexpected traffic-serving window** (should be drained but is not)

## Expected behavior

Once the Pod enters (or is about to enter) an in-place upgrade flow, the drain action (`InPlaceUpdateReady=false`) should not be skipped due to cache inconsistency. The readiness/traffic gating should remain consistent with the upgrade state, even when `Update Pod Spec` temporarily fails and retries.

veophi:
Fixes a timing/cache inconsistency issue in the InPlaceUpdate reconcile flow that could cause the traffic-drain step to be skipped.

In the original flow, `InPlaceUpdateReady=false` (drain) is updated before `Update Pod Spec`. When `Update Pod Spec` fails due to resourceVersion conflicts (controller cache not catching up), the controller may later set `InPlaceUpdateReady=true` again, and then incorrectly **skip** the next drain patch because it relies on a stale cached `InPlaceUpdateReady` value. This may lead to unexpected traffic being served during the in-place upgrade window.

This PR makes the drain/ready-state update logic robust against informer cache staleness so that once a Pod is going to be in-place upgraded, the controller will not miss (or skip) setting `InPlaceUpdateReady=false`.

fixes #2276

### DIFF
--- a/pkg/util/inplaceupdate/inplace_update.go
+++ b/pkg/util/inplaceupdate/inplace_update.go
@@ -169,6 +169,10 @@ func (c *realControl) Refresh(pod *v1.Pod, opts *UpdateOptions) RefreshResult {
 		Status:             v1.ConditionTrue,
 		LastTransitionTime: metav1.NewTime(Clock.Now()),
 	}
+	if hasEqualCondition(pod, &newCondition) {
+		return RefreshResult{}
+	}
+
 	// Do not retry on conflict; only update the condition for the checked Pod version.
 	// see https://github.com/openkruise/kruise/pull/2274
 	err := c.updateCondition(pod, newCondition)
@@ -179,10 +183,6 @@ func (c *realControl) Refresh(pod *v1.Pod, opts *UpdateOptions) RefreshResult {
 // no conflict retry or refetch is performed, so it only applies to this specific Pod version.
 func (c *realControl) updateCondition(pod *v1.Pod, condition v1.PodCondition) error {
 	clone := pod.DeepCopy()
-	if hasEqualCondition(clone, &condition) {
-		return nil
-	}
-
 	util.SetPodCondition(clone, condition)
 	// We only update the ready condition to False, and let Kubelet update it to True
 	if condition.Status == v1.ConditionFalse {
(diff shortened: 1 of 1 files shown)


pkg/util/inplaceupdate/inplace_update.go
```diff
@@ -336,6 +336,9 @@ func (c *realControl) Update(pod *v1.Pod, oldRevision, newRevision *apps.Control
 			if err != nil {
 				return err
 			}
+			// The Pod from informer cache may be not synced as the latest, so we
```
furykerry, line 339: shall we compare hasEqualCondition before retryOnConflict

veophi: > shall we compare hasEqualCondition before retryOnConflict

@furykerry  No. If we still use `hasEqualCondition` here, we need to make sure we’re working with **the latest Pod object**; otherwise we may still read **stale/dirty** cache data and incorrectly **skip** the following`updatePodCondition` step.

PR #2291, opened 2026-01-10
Fix the issue where Pods remain NotReady when CloneSet UpdateStrategy…

zmberg:
… is set to OnDelete.


furykerry (approved):
/lgtm
/approve

Issue #2286, opened 2026-01-07
[BUG] ImagePullJob指定selector为空{}时，status会卡在Active导致无法启动下一批


stan-DP:
**What happened**:
配置ImagePullJob资源如下：
``` yaml
apiVersion: apps.kruise.io/v1beta1
kind: ImagePullJob
metadata:
  creationTimestamp: "2026-01-05T11:30:00Z"
  finalizers:
  - apps.kruise.io/deletion-protection
  generateName: xcj-test-0104-1767612600-
  generation: 1
  labels:
    controller-revision-hash: c46cf8bb9
  name: xcj-test-0104-1767612600-2xglm
  namespace: kube-system
  ownerReferences:
  - apiVersion: apps.kruise.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: ImageListPullJob
    name: xcj-test-0104-1767612600
    uid: 0cd4fca8-f224-41d5-a3e9-0eb0db1c8293
  resourceVersion: "1454898300"
  uid: 56b8ccd1-af63-4693-a487-08bc3c12c04d
spec:
  completionPolicy:
    type: Always
  image: docker.io/library/nginx:latest
  imagePullPolicy: IfNotPresent
  parallelism: "5"
  pullPolicy:
    backoffLimit: 3
    timeoutSeconds: 600
  pullSecrets:
  - default-secret
  selector: {}
status:
  active: 5
  desired: 85
  failed: 0
  message: job is running, progress 0.0%
  startTime: "2026-01-05T11:30:00Z"
  succeeded: 0
```
镜像预热任务会匹配到所有节点，按照parallelism每批5个。首批五个NodeImage CR的状态刷成Succeeded之后，imagepulljob的状态并没有更新，仍然是active: 5, succeeded: 0

**What you expected to happen**:
两种情况都能接受：
1、最合理：按照参考LabelSelector的默认行为，空结构体意味着不匹配任何标签。空的selector（not nil）应该不会匹配到任何节点
2、将selector: {}等价于不指定selector，匹配到所有节点，能够正常按parallelism并发度进行镜像下载，状态正常更新，不会出现卡住的情况

**How to reproduce it (as minimally and precisely as possible)**:
ImagePullJob中不指定PodSelector，指定selector为空{}

**Anything else we need to know?**:
导致状态不刷新的原因
``` go
[...]

PR #2300, opened 2026-01-16
fix: handle empty selector as match-all for ImagePullJob

kajal-jotwani:
Fixes an issue where ImagePullJob with an empty node selector (selector: {}) was treated as matching no nodes.
The job now correctly treats an empty selector as matching all nodes, consistent with Kubernetes selector semantics.

fixes:#2286

Added a new test case using an empty selector.
Updated existing tests to ensure:
- selector: {} behaves the same as no selector.
- Jobs with empty selectors are correctly associated with all nodes.
All related unit tests pass in CI.


kajal-jotwani:
@furykerry could you please take a look when you get a chance? Thanks!

PR #2328, opened 2026-01-23
fix: prevent 409 conflicts after refresh

### LINKED ISSUE #2167: [discuss] The first image-change patch on a pod always conflict(409) in cloneSet controller
The scenario  triggered by the following set of conditions:

1.  The UpdateStrategy of the CloneSet is configured as 'InPlaceOnly' or 'InPlaceIfPossible'.
2.  A pod contains 'InPlaceUpdateReady' condition. In fact, the condition must be injected when inPlace-gate was enabled
3. A pod was ready to inPlace upgrade, after it paased various validations.

**Issue**:

The container image was updated immediately after the InPlaceUpdateReady condition was updated, almost at the same time. 

The code is located at `/pkg/util/inplaceupdate/inplace_update.go#(c *realControl) Update`

Image

This is an always reproducible issue.

Even though the code has retry logic(retry.RetryOnConflict) to maintain functionality, 
but is it possible to prevent these unnecessary requests?

**Discuss**:
I came up with a few possible solutions:

1. Maybe requeue key when update the status subresource?
2. Maybe get the latest revison after updated the status subresource? (it will be return the latest by default)
3. Maybe get the latest pod when update pod spec?

dibyanshu-pal-kushwaha:
This PR fixes a race condition in the CloneSet controller that causes 409 Conflict errors during the first in-place image update of a Pod.

Fixes [#2167](https://github.com/openkruise/kruise/issues/2167)
- Added RefreshUpdated boolean field to RefreshResult in pkg/util/inplaceupdate.
- Updated Refresh and finishGracePeriod to correctly report true whenever a write operation occurs.
- Updated cloneset_update.go to yield execution if Refresh modifies the Pod, ensuring the cache syncs before spec updates are attempted.

1. Local Verification: Verified locally using a temporary reproduction test case that mocked Refresh to force an update. Confirmed that the controller yields execution immediately instead of proceeding to call Update with stale data.
2. Regression Testing:
Ran go test ./pkg/controller/cloneset/sync/... -> PASSED
Ran go test ./pkg/util/inplaceupdate/... ->PASSED

None

### DIFF
--- a/pkg/controller/cloneset/sync/cloneset_update.go
+++ b/pkg/controller/cloneset/sync/cloneset_update.go
@@ -171,6 +171,9 @@ func (c *realControl) refreshPodState(cs *appsv1beta1.CloneSet, coreControl clon
 			"cloneSet", klog.KObj(cs), "pod", klog.KObj(pod))
 		return false, 0, res.RefreshErr
 	}
+	if res.RefreshUpdated {
+		return true, res.DelayDuration, nil
+	}
 
 	var state appspub.LifecycleStateType
 	switch lifecycle.GetPodLifecycleState(pod) {
--- a/pkg/util/inplaceupdate/inplace_update.go
+++ b/pkg/util/inplaceupdate/inplace_update.go
@@ -50,8 +50,9 @@ var (
 )
 
 type RefreshResult struct {
-	RefreshErr    error
-	DelayDuration time.Duration
+	RefreshErr     error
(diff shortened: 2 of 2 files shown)


dibyanshu-pal-kushwaha:
@furykerry Please test this and review this,Sir

dibyanshu-pal-kushwaha:
@furykerry ,This pr is ready to review. Please review it sir.

PR #2342, opened 2026-01-25
fix: prevent lifecycle patch from deleting existing pod finalizers

Aman-Cool:
## Fix: Preserve Pod Finalizers During Lifecycle Updates

Fixes a bug where lifecycle updates could unintentionally clear all existing pod finalizers when `finalizersHandler` was not set.

The patch now avoids touching `metadata.finalizers` unless explicitly configured, preventing silent data loss and preserving system and third-party finalizers.


Aman-Cool:
@zmberg @furykerry @FillZpp , Prevents lifecycle updates from accidentally clearing existing pod finalizers when no `finalizersHandler` is configured.

PR #2354, opened 2026-01-30
docs/dev: add dev overlay with IfNotPresent imagePullPolicy for local KIND development

### LINKED ISSUE #2325: [BUG] Local KIND development fails because imagePullPolicy=Always
# Local KIND development fails because imagePullPolicy=Always

**What happened**:
When running OpenKruise locally with kind using locally-built images (for example `openkruise/kruise-manager:test`), pods go into `ImagePullBackOff` because the manifests set `imagePullPolicy: Always`. Even after loading the image into the kind nodes (`kind load docker-image ...`), Kubernetes still attempts to pull from the remote registry and fails.

When a developer loads a locally-built image into kind, the cluster should use the local image (or at least not always try remote pull). Local dev manifests should allow `IfNotPresent` so local development works without pushing images.

1. Build image locally.
2. Create a kind cluster.
3. Install CRDs (if not already):
4. Deploy the current default manifests:
5. Load the locally-built image into kind:

   ```bash
   kind load docker-image openkruise/kruise-manager:test --name kind
   ```
6. Observe pods in `kruise-system`:

   ```bash
   kubectl get pods -n kruise-system
   ```

   You’ll likely see `ImagePullBackOff` / `ErrImagePull` events complaining about failing to pull from registry.

* The error surfaces for local dev workflows (kind). It’s not a production runtime bug; production may intentionally use `Always`.

**Environment**:
* Kubernetes version: v1.22.15
* Others: macOS, Docker Desktop, kind

d1vyanshu-kumar:

This PR improves the local development workflow for contributors using KIND by adding a dedicated dev overlay (`config/dev`) that sets `imagePullPolicy: IfNotPresent` for Kruise manager and daemon images.

Currently, the default manifests use `imagePullPolicy: Always`. When developers build images locally (e.g. `openkruise/kruise-manager:test`) and load them into a KIND cluster using `kind load docker-image`, Kubernetes still tries to pull images from the remote registry, which leads to `ImagePullBackOff` errors.

To avoid changing production behavior, this PR does NOT modify `config/manager`. Instead, it adds a separate `config/dev` manifest set intended only for local development. This keeps production/CI behavior unchanged while making local contributor setup smoother.

Changes included:
- Added `config/dev` overlay based on `config/manager`
- Updated imagePullPolicy from `Always` → `IfNotPresent` in dev manifests only
- Added documentation for local KIND-based development workflow

This is focused purely on developer experience and does not change runtime logic.

fixes #2325

Verify locally with KIND:

```bash
make docker-build
kind create cluster --name kind
kind load docker-image openkruise/kruise-manager:test --name kind
kubectl apply -f config/dev
kubectl get pods -n kruise-system

Expected result:
* Pods start successfully
* No ImagePullBackOff errors
* Manager and daemon pods reach Running state
```

* This PR intentionally does NOT modify `config/manager` to avoid impacting production or CI behavior.
* The change is isolated to a new dev-only manifest set.
* The goal is to improve contributor onboarding and local testing with KIND.
* If maintainers prefer a kustomize patch overlay instead of a copied dev config, I can refactor this PR accordingly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -105,6 +105,12 @@ make build
 
 # Run the unit tests
 make test
+
+### Local development with KIND
+
+For local development using KIND and locally built images,
+use manifests under `config/dev`, which set imagePullPolicy
+to `IfNotPresent`.
 ```
 
 **There are some guide documents for contributors in [./docs/contributing/](./docs/contributing), such as a debug guide to help you test your own branch in a Kubernetes cluster.**
--- a/config/dev/kustomization.yaml
+++ b/config/dev/kustomization.yaml
@@ -1,0 +1,11 @@
+# Adds namespace to all resources.
+namespace: kruise-system
+
+resources:
+- manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: openkruise/kruise-manager
+  newTag: test
--- a/config/dev/manager.yaml
+++ b/config/dev/manager.yaml
@@ -1,0 +1,154 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: system
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: manager
+  namespace: system
+---
+apiVersion: apps/v1
+kind: Deployment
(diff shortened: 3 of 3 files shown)


d1vyanshu-kumar:
@furykerry Could you please review my PR and let me know if it looks good or if any changes are needed?

PR #2413, opened 2026-04-18
fix(uniteddeployment): clamp negative maxUnavailable to 0 during adaptive rescheduling scale-down race

### LINKED ISSUE #2412: [BUG] UnitedDeployment: negative maxUnavailable written to child workload during adaptive rescheduling scale-down race
**What happened**:

I was working with a `UnitedDeployment` that had `reserveUnschedulablePods: true` enabled under the adaptive schedule strategy. While a subset was in an unschedulable state and I was scaling down at the same time, I noticed that all the pods in that subset started restarting all at once — no rolling update behaviour, just everything going down together. That's when I dug into the controller logic and found something off.

The controller computes a `maxUnavailable` value to overwrite the child workload's (CloneSet/StatefulSet) rolling update limit when a subset is unschedulable. The formula is:

```
maxUnavailable = Spec.Replicas - Status.ReadyReplicas + UpdateTimeoutPods
```

The problem is that during a scale-down, there's a natural race window where `ReadyReplicas` can temporarily be higher than `Spec.Replicas` — new pods from the receiving subset are already up and ready before the excess pods on the current subset have finished terminating. When that happens, the subtraction goes negative, and that negative value gets directly written as `MaxUnavailable` into the child workload. As far as the CloneSet or StatefulSet is concerned, there's no limit anymore, and it lets everything roll at once.

The rolling update limit should stay sane even during transient conditions. If the computed value ends up negative for whatever reason, the controller should treat it as `0` (no additional pods may go unavailable) rather than silently disabling all limits. It's a small race window, but the consequences — an uncontrolled full restart during an already degraded state — feel pretty severe for production workloads.

1. Set up a `UnitedDeployment` with at least two subsets and adaptive strategy:

```yaml
spec:
  topology:
    scheduleStrategy:
      type: Adaptive
      adaptive:
        reserveUnschedulablePods: true
    subsets:
      - name: subset-a
        nodeSelectorTerm: ...
      - name: subset-b
        nodeSelectorTerm: ...
```

2. Trigger a situation where `subset-a` becomes unschedulable — either by tainting the nodes or causing pods to stay pending past the rescheduling threshold.

3. While rescheduling is happening and pods from `subset-b` are coming up, reduce `spec.replicas` at the same time.

4. Watch the pods in `subset-a`. During the window where `subset-b`'s pods are ready but `subset-a`'s excess pods haven't terminated yet, the controller reconciles and sets `MaxUnavailable` to a negative number on the child workload.

5. The child workload (CloneSet or StatefulSet) treats it as essentially unlimited, and all pods in `subset-a` go down simultaneously.

The bug only affects setups where `reserveUnschedulablePods: true` is configured — so it's behind a flag, which might be why it hasn't surfaced widely. But the race condition itself is not unusual at all; it happens during any normal scale-down while rescheduling is active. The tricky part is that there's no error, no warning in the logs — the controller just quietly sets a bad value and the workload behaves unexpectedly.

**Environment**:
- Kruise version: tested on latest main
- Kubernetes version (use `kubectl version`): v1.28+
- Install details (e.g. helm install args): standard helm install, no custom flags
- Others: only reproducible when `reserveUnschedulablePods: true` is set on the UnitedDeployment

rakshaak29:

While looking into the adaptive rescheduling logic for `UnitedDeployment`, I found that the controller can silently write a **negative** value as `MaxUnavailable` to the child workload during a scale-down race.

When a subset is unschedulable and `reserveUnschedulablePods: true` is set, the controller computes a `maxUnavailable` override inside `UpdateSubset`:

```go
maxUnavailable := subset.Spec.Replicas - subset.Status.ReadyReplicas + subset.Status.UnschedulableStatus.UpdateTimeoutPods
```

During scale-down there is a natural window where pods from the receiving subset are already `Ready` while the excess pods on the current subset are still terminating. In that window `ReadyReplicas > Spec.Replicas`, so the subtraction goes negative:

```
Spec.Replicas=3, ReadyReplicas=5, UpdateTimeoutPods=1
→ maxUnavailable = 3 − 5 + 1 = -1
```

That `-1` gets passed straight to `adapter.SetMaxUnavailable(workload, -1)`, writing it into the child workload's `MaxUnavailable` field. The child workload (CloneSet / StatefulSet / Deployment) treats this as unlimited and lets all pods restart at once — no rolling update, no limit, everything goes down together. This is the exact opposite of what the safety mechanism is supposed to do, and it happens silently with no warning in the logs.

This PR fixes it with a simple clamp: if the computed value is negative, it is set to `0` before being passed to `SetMaxUnavailable`. `MaxUnavailable=0` is the safest conservative fallback — it prevents any additional pods from going unavailable while the controller waits for the transient overcount to resolve on the next reconciliation. A structured `V(3)` log line is also emitted when the clamp fires, giving operators clear visibility into when and why the guard triggered.

---

fixes #2412 (I opened an issue for this bug report)
---

Run the new unit test directly:

```bash
go test ./pkg/controller/uniteddeployment/ -run TestUpdateSubset_MaxUnavailableClamp -v
```

The test (`TestUpdateSubset_MaxUnavailableClamp` in `subset_control_test.go`) uses a spy adapter that intercepts `SetMaxUnavailable` and asserts the value is never negative. It covers four cases:

- **positive** — `ReadyReplicas < Spec.Replicas`, formula gives a valid positive value, passed through as-is
- **zero** — `ReadyReplicas == Spec.Replicas`, formula gives exactly `0`, passed through as-is
- **negative (the bug case)** — `Spec=3, Ready=5, TimeoutPods=1 → raw=-1 → clamped to 0`
- **large excess** — `Spec=2, Ready=10, TimeoutPods=3 → raw=-5 → clamped to 0`

To reproduce the original bug manually: create a `UnitedDeployment` with `reserveUnschedulablePods: true`, drive one subset into an unschedulable state, then scale down while rescheduling is in progress. Before this fix, all pods in the affected subset restart simultaneously with no rolling-update limit enforced.

---

The clamp value is `0` not `1`. This is intentional — `MaxUnavailable=0` is the most conservative correct value for a subset that is already in a degraded state. It does not block the update entirely; pods that are already unavailable still count toward progress. The controller will naturally recompute the correct positive `maxUnavailable` on the very next reconciliation once `ReadyReplicas` drops back below `Spec.Replicas` as the terminating pods finish.


pkg/controller/uniteddeployment/subset_control.go
```diff
@@ -103,7 +103,21 @@ func (m *SubsetControl) UpdateSubset(subset *Subset, ud *alpha1.UnitedDeployment
...
 		if subset.Status.UnschedulableStatus.Unschedulable && ud.Spec.Topology.ScheduleStrategy.ShouldReserveUnschedulablePods() {
-			maxUnavailable := subset.Spec.Replicas - subset.Status.ReadyReplicas + subset.Status.UnschedulableStatus.UpdateTimeoutPods
+			// maxUnavailable is computed as:
+			//   (target replicas) - (currently ready) + (pods timed out waiting for scheduling)
+			// This gives the number of pods we can afford to take down during an update
+			// while keeping the subset functional at its target capacity.
+			maxUnavailable := replicas - subset.Status.ReadyReplicas + subset.Status.UnschedulableStatus.UpdateTimeoutPods
+			// Guard: ReadyReplicas can transiently exceed the target replicas during a scale-down race,
+			// making the result negative. Clamp to 0 to avoid disabling all rolling-update limits.
+			if maxUnavailable < 0 {
```
AiRanthem, line 113: Would it be better to clamp `maxUnavailable` to **1** instead of **0**?

The reason is that Kruise workloads have webhook validations that reject `maxUnavailable < 1`:
- AdvancedStatefulSet: unconditionally requires `>= 1`
- CloneSet: requires `maxUnavailable + maxSurge >= 1`

With `maxUnavailable = 0`, the `Client.Update` call would hit the same webhook error as a negative value, causing the reconciliation to fail on every retry. Even if it somehow gets written through, a value of 0 blocks all rolling updates (`len(unavailablePods) >= 0` is always true).

Clamping to 1 ensures the update is always accepted by the webhook while still being conservative — it just allows one pod to update at a time during the scale-down transient period.

Suggested change:

```go
if maxUnavailable < 1 {
    maxUnavailable = 1
}
```

AiRanthem (commented):
Shall we generate the open api in a stand-alone PR? @furykerry

PR #2445, opened 2026-05-07
feat(resourcedistribution): promote to v1beta1 storage version

Kunal1522:
namespaceLabelSelector renamed to namespaceSelector
Aligns with Kubernetes conventions and other Kruise APIs (UnitedDeployment, CloneSet). The conversion webhook transparently remaps old objects so existing v1alpha1 resources continue to work.

kube-system and kube-public are always excluded from distribution
System namespaces are now filtered out at the top of both matchViaTargets and listNamespacesForDistributor, so they are never distributed into regardless of allNamespaces, includedNamespaces, or namespaceSelector. Previously only the allNamespaces path filtered them 

spec.resource.metadata.namespace is silently ignored instead of rejected
The controller always overwrites the namespace when distributing anyway.

status.failedNamespaces JSON tag corrected to camelCase
Was serialized as FailedNamespaces (Go default). Fixed to failedNamespaces per JSON conventions.


Kunal1522:
the failure of e2e 1.26 is probably flake

pkg/controller/resourcedistribution/utils.go
```diff
@@ -52,6 +52,10 @@ const (
 	OperationSucceeded     = "Succeeded"
 )
 
+// defaultSystemNamespaces are excluded from allNamespaces distribution.
+// Use spec.targets.includedNamespaces to target them explicitly.
+var defaultSystemNamespaces = sets.NewString("kube-system", "kube-public")
```
furykerry: this variable can be modified, consider change it to a simple function

pkg/controller/resourcedistribution/utils.go
```diff
@@ -90,13 +94,13 @@ func matchViaLabelSelector(namespace *corev1.Namespace, distributor *appsv1alpha
...
+func matchViaTargets(namespace *corev1.Namespace, distributor *appsv1beta1.ResourceDistribution) (bool, error) {
 	targets := &distributor.Spec.Targets
 	if isInList(namespace.Name, targets.ExcludedNamespaces.List) {
 		return false, nil
 	}
 	if targets.AllNamespaces {
-		return true, nil
+		return !defaultSystemNamespaces.Has(namespace.Name), nil
 	}
 	if isInList(namespace.Name, targets.IncludedNamespaces.List) {
```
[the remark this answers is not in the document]
furykerry: @Kunal1522 allNamespaces system namespace exclusion is not overridable via IncludedNamespaces despite doc claims

Kunal1522:
hi @furykerry

I am sorry but can you please explain these doubts and guide which option should be go with 

The docs say:

>"AllNamespaces, includedNamespaces and excludedNamespaces are "OR" relationship, and excludedNamespaces will always effect if users set it. By the way, targets will always ignore the kube-system and kube-public namespaces"

So if we follow the docs, I think the code change we need is to ignore kube-system and kube-public always — even if the user puts them explicitly in includedNamespaces, regardless of whether allNamespaces is true or false. The code would look like this:

```
func matchViaTargets(namespace *corev1.Namespace, distributor *appsv1beta1.ResourceDistribution) (bool, error) {
    targets := &distributor.Spec.Targets
    // system namespaces are always excluded, no exceptions
    if defaultSystemNamespaces.Has(namespace.Name) {
        return false, nil
    }
    if isInList(namespace.Name, targets.ExcludedNamespaces.List) {
        return false, nil
    }
    if targets.AllNamespaces {
        return true, nil
    }
    if isInList(namespace.Name, targets.IncludedNamespaces.List) {
        return true, nil
    }
    return matchViaLabelSelector(namespace, distributor)
}
 ```


But if we decide to make the system namespace exclusion configurable — i.e. the user CAN override it by explicitly listing kube-system in includedNamespaces — then the code would look like this:
```
func matchViaTargets(namespace *corev1.Namespace, distributor *appsv1beta1.ResourceDistribution) (bool, error) {
    targets := &distributor.Spec.Targets
    if isInList(namespace.Name, targets.ExcludedNamespaces.List) {
        return false, nil
    }
    // explicit inclusion always wins, even for system namespaces
    if isInList(namespace.Name, targets.IncludedNamespaces.List) {
        return true, nil
    }
    if targets.AllNamespaces {
        return !defaultSystemNamespaces.Has(namespace.Name), nil
    }
    return matchViaLabelSelector(namespace, distributor)
}
```

pkg/controller/resourcedistribution/utils.go
Kunal1522: fixed https://github.com/openkruise/kruise/pull/2445/changes/6c5e3f709972ae52e1b402953fc3183f42d2049f

PR #2483, opened 2026-06-17
fix: resolve ImagePullJob high CPU caused by cascading reconcile amplification

zmberg:
## Summary

- When multiple ImagePullJobs target different tags of the same image, a single tag TTL expiry triggers reconciliation of ALL sibling jobs (~887 reconcile/s, 94.6% CPU in a cluster with 198 jobs × 989 nodes)
- **syncJobPullSecrets**: early return when job has no pullSecrets, eliminating unnecessary namespace/secret cache queries
- **syncNodeImages**: fix a bug where pullSecrets were never persisted to NodeImage when tag+ownerRef already existed
- **handleUpdate**: refine NodeImage change detection from image-name level to tag level, reducing reconcile rate by ~99.5%

## Test plan

- [x] Unit tests pass (`go test ./pkg/controller/imagepulljob/...`)
- [x] Added tag-level event handler tests covering: tag change precision, pullSecrets change enqueue-all, tag deletion, status change precision, image deletion enqueue-all, pure status single-tag change
- [ ] Run existing E2E `[ImagePullJob]` suite to verify no regression

### DIFF
--- a/pkg/controller/imagepulljob/imagepulljob_controller.go
+++ b/pkg/controller/imagepulljob/imagepulljob_controller.go
@@ -296,6 +296,10 @@ func (r *ReconcileImagePullJob) Reconcile(_ context.Context, request reconcile.R
 // so it is necessary to synchronize the pullsecrets of imagepulljob to the kruise-daemon-config namespace
 // to facilitate kruise-daemon to fetch them.
 func (r *ReconcileImagePullJob) syncJobPullSecrets(job *appsv1beta1.ImagePullJob) ([]appsv1beta1.ReferenceObject, error) {
+	if len(job.Spec.PullSecrets) == 0 && job.DeletionTimestamp.IsZero() && job.Status.CompletionTime == nil {
+		return nil, nil
+	}
+
 	var secretRefs []appsv1beta1.ReferenceObject
 	// If it's in kruise-daemon-config namespace, no need to go through the sync logic, return directly
 	if job.Namespace == util.GetKruiseDaemonConfigNamespace() {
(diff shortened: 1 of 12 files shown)


pkg/controller/imagepulljob/imagepulljob_controller.go
```diff
@@ -296,6 +296,10 @@ func (r *ReconcileImagePullJob) Reconcile(_ context.Context, request reconcile.R
 // so it is necessary to synchronize the pullsecrets of imagepulljob to the kruise-daemon-config namespace
 // to facilitate kruise-daemon to fetch them.
 func (r *ReconcileImagePullJob) syncJobPullSecrets(job *appsv1beta1.ImagePullJob) ([]appsv1beta1.ReferenceObject, error) {
+	if len(job.Spec.PullSecrets) == 0 && job.DeletionTimestamp.IsZero() && job.Status.CompletionTime == nil {
```
furykerry, line 299: The new early return if len(job.Spec.PullSecrets) == 0 && job.DeletionTimestamp.IsZero() && job.Status.CompletionTime == nil { return nil, nil } skips the secret cleanup logic when a running job has its pullSecrets removed (set to empty) by the user. Previously, syncJobPullSecrets → classifyPullSecretsForJob → releaseImagePullJobSecrets would detect that previously-synced secrets in the kruise-daemon-config namespace no longer match the job's desired secrets and release/delete them. With the early return, these orphaned secrets are not cleaned up until the job eventually completes or is deleted (at which point finalize calls syncJobPullSecrets without hitting the early return).

pkg/controller/imagepulljob/imagepulljob_event_handler.go
```diff
@@ -125,6 +161,42 @@ func (e *nodeImageEventHandler) handleUpdate(nodeImage, oldNodeImage *appsv1beta
 	}
 }
 
+func getChangedSpecTags(newTags, oldTags []appsv1beta1.ImageTagSpec) sets.String {
```
furykerry, line 164: consider add direct unit test for getChangedSpecTags and getChangedStatusTags, so that  edge cases like tag reordering (same tags, different order → should return empty set) and duplicate tags are covered.

furykerry:
plz add unit test for syncJobPullSecrets, the changes made in this patch is not covered in unit tests

PR #2485, opened 2026-06-20
 fix CRR queue progression after deletion

### LINKED ISSUE #2474: [BUG] ContainerRecreateRequest stuck in Pending when recreating different containers in the same Pod continuously
**What happened**:
When using `ContainerRecreateRequest` (CRR) to restart a specific container in a Pod, the first CRR succeeds. After deleting the successful CRR, if I immediately create a new CRR to restart *another* container in the exact same Pod, the second CRR gets stuck in the `Pending` state indefinitely. 

This process is executed programmatically via code. Interestingly, if I add a delay of a few seconds between deleting the first CRR and creating the second one, the issue does not occur, and the second CRR is processed normally.

The Kruise controller should be able to process the second `ContainerRecreateRequest` normally and recreate the second container without getting stuck in the `Pending` state, even if the second CRR is created immediately after the deletion of the first one.

1. Deploy a Pod with at least two containers (e.g., `container-a` and `container-b`).
2. Create a `ContainerRecreateRequest` (CRR-1) to recreate `container-a`.
3. Wait for CRR-1 to complete successfully.
4. Delete CRR-1.
5. Immediately (via script/code, without sleep) create a new `ContainerRecreateRequest` (CRR-2) targeting `container-b` in the same Pod.
6. Observe that CRR-2 remains in the `Pending` state and does not progress.

This seems to be related to the informer cache sync delay or state conflict. Since adding a few seconds of sleep between the operations resolves the issue, it is highly likely that when the second CRR is created, the Kruise controller or daemon is reading a stale Pod state from the cache (possibly thinking the Pod is still undergoing the previous recreation process), which blocks the processing of the new CRR.

**Environment**:
- Kruise version: master branch (built on May 25, 2026)
- Kubernetes version (use `kubectl version`): v1.30.14
  `kruise-manager` startup args:
  ```yaml
  args: 
    - '--enable-leader-election' 
    - '--metrics-addr=:8080' 
    - '--health-probe-addr=:8000' 
    - '--logtostderr=true' 
    - '--leader-election-namespace=kruise-system' 
    - '--v=4' 
    - >- 
      --feature-gates=ResourcesDeletionProtection=true,CloneSetShortHash=true,InPlaceUpdateEnvFromMetadata=true,PodWebhook=true,PreDownloadImageForInPlaceUpdate=false,PodUnavailableBudgetDeleteGate=false,WorkloadSpread=true,StatefulSetAutoDeletePVC=false,PodProbeMarkerGate=false 
    - '--sync-period=1h' 
    - '--cloneset-workers=10' 
    - '--statefulset-workers=10'

Jayant-kernel:

This PR prevents a `ContainerRecreateRequest` (CRR) from remaining `Pending` when it is created immediately after a completed CRR for the same Pod is deleted.

The kruise-daemon intentionally serializes CRRs by using `<namespace>/<podName>` as its workqueue key. A race was possible when a second CRR was created while the daemon informer still contained the first CRR:

1. The second CRR's add event queued the Pod.
2. The daemon's stale informer view still selected the first CRR, leaving the second CRR `Pending`.
3. When the first CRR was deleted or lost its `active` label after completion, the filtered informer's delete handler only cleared its resource-version expectation.
4. Because the informer has no periodic resync, no event queued the Pod again and the second CRR remained `Pending` indefinitely.

The delete handler now:

- clears the deleted CRR's resource-version expectation;
- requeues its Pod so the next pending CRR can progress; and
- handles `DeletedFinalStateUnknown` tombstones when the original delete notification was missed.

Unit coverage verifies that both normal delete events and tombstone delete events enqueue the expected Pod key. The existing one-at-a-time execution semantics for CRRs targeting the same Pod are preserved.

Fixes #2474

The following checks pass locally with Go 1.23.8:

```shell
go test -count=1 ./pkg/daemon/containerrecreate
go test -run TestEnqueueAfterDelete -count=100 ./pkg/daemon/containerrecreate
go vet ./pkg/daemon/containerrecreate
git diff --check
```

To verify the original scenario in a cluster:

1. Create a Pod containing two containers.
2. Create a CRR for the first container and wait for it to complete.
3. Delete that CRR and immediately create another CRR for the second container without sleeping.
4. Confirm that the second CRR advances from `Pending` to `Recreating` and then `Completed`.

This change deliberately keeps the Pod-scoped queue key instead of allowing CRRs for one Pod to execute concurrently. It only restores the missing wake-up when the previous active CRR leaves the filtered informer.

`go test ./pkg/daemon/...` was also attempted. The modified `containerrecreate` package and the other daemon packages passed, except the unrelated existing `pkg/daemon/criruntime/imageruntime` test, which requires a credential-provider plugin binary that is not present in the local environment. A live cluster e2e run was not available locally.

### DIFF
--- a/pkg/daemon/containerrecreate/crr_daemon_controller.go
+++ b/pkg/daemon/containerrecreate/crr_daemon_controller.go
@@ -101,10 +101,7 @@ func NewController(opts daemonoptions.Options) (*Controller, error) {
 			}
 		},
 		DeleteFunc: func(obj interface{}) {
-			crr, ok := obj.(*appsv1beta1.ContainerRecreateRequest)
-			if ok {
-				resourceVersionExpectation.Delete(crr)
-			}
+			enqueueAfterDelete(queue, obj)
 		},
 	})
 
@@ -156,6 +153,29 @@ func enqueue(queue workqueue.Interface, obj *appsv1beta1.ContainerRecreateReques
 		return
 	}
 	queue.Add(objectKey(obj))
+}
+
+// enqueueAfterDelete requeues the Pod so another CRR targeting it can be processed.
+// The filtered informer sends a delete event both when a CRR is deleted and when
+// its active label is removed after completion. It can also send a tombstone when
+// the original delete notification was missed.
+func enqueueAfterDelete(queue workqueue.Interface, obj interface{}) {
+	crr, ok := obj.(*appsv1beta1.ContainerRecreateRequest)
+	if !ok {
+		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("couldn't get object from tombstone %+v", obj))
+			return
+		}
+		crr, ok = tombstone.Obj.(*appsv1beta1.ContainerRecreateRequest)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("tombstone contained object that is not a ContainerRecreateRequest %+v", obj))
+			return
+		}
+	}
+
+	resourceVersionExpectation.Delete(crr)
+	queue.Add(objectKey(crr))
 }
 
 func objectKey(obj *appsv1beta1.ContainerRecreateRequest) string {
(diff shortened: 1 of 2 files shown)


Jayant-kernel:
@furykerry review my pr

PR #2513, opened 2026-07-19
fix: add validation for BurstReplicas in DaemonSet validating webhook

### LINKED ISSUE #2512: [BUG] Missing Validation for BurstReplicas in DaemonSet Webhook Causes Controller Panic and DoS
**What happened**:
The `DaemonSet` validating webhooks (for both `v1alpha1` and `v1beta1`) do not validate the `BurstReplicas` field, which limits the rate of scaling pods. This allows users to create or update a DaemonSet with a negative integer (e.g., `-5`) or an invalid string for `BurstReplicas`.

When the `DaemonSet` controller reconciles a DaemonSet with `burstReplicas: -5`, the `getBurstReplicas` helper fetches this negative value and ignores any parsing errors (due to an explicit assumption that `// Error caught by validation`). The controller then uses this negative limit in `syncNodes` and `inPlaceUpdatePods` to cap the variables `createDiff`, `deleteDiff`, and `updateDiff` at `-5`. Finally, the controller allocates an error channel using `make(chan error, updateDiff)` or `make(chan error, createDiff+deleteDiff)` with a negative capacity. This immediately triggers a runtime panic (`panic: makechan: size out of range`), placing the `kruise-manager` in a crash loop and causing a Denial of Service (DoS) for all controllers.

Additionally, specifying `burstReplicas: 0` or an invalid string like `"invalid%"` silently freezes the DaemonSet, as the controller thinks no pods can be created or deleted, with no useful error surfaced.

The validating webhooks for `DaemonSet` should enforce that `BurstReplicas` is a strictly positive integer or valid percentage. Invalid or negative configurations should be rejected at the API layer, rather than crashing the controller or freezing reconciliations.

1. Ensure OpenKruise is running.
2. Apply an Advanced DaemonSet with a negative `burstReplicas`:
```yaml
kind: DaemonSet
metadata:
  name: malicious-ds
spec:
  burstReplicas: -5
  selector:
    matchLabels:
      app: malicious
  template:
    metadata:
      labels:
        app: malicious
    spec:
      containers:
      - name: nginx
        image: nginx
```
3. Check the `kruise-manager` logs. The controller will panic during reconciliation due to an invalid negative channel size allocation (e.g. `make(chan error, -10)`).

**Anything else we need to know?**:.

**Proposed Fix**
Add validation for `BurstReplicas` in `pkg/webhook/daemonset/validating/daemonset_validation.go` (for both `v1alpha1` and `v1beta1`):
```go
import (
	appsvalidation "k8s.io/kubernetes/pkg/apis/apps/validation"
)

func validateDaemonSetSpecV1beta1(spec *appsv1beta1.DaemonSetSpec, fldPath *field.Path) field.ErrorList {
	allErrs := field.ErrorList{}
	// ... existing validations ...

	if spec.BurstReplicas != nil {
		allErrs = append(allErrs, appsvalidation.ValidatePositiveIntOrPercent(*spec.BurstReplicas, fldPath.Child("burstReplicas"))...)
	}
	// ...
}
```

**Environment**:
- Kruise version: master
- Kubernetes version (use `kubectl version`): N/A
- Install details (e.g. helm install args): N/A
- Others: N/A

vishalmore90:
This PR adds missing field validation for `spec.BurstReplicas` in both `v1alpha1` and `v1beta1` `DaemonSet` validating webhooks.

Previously, `BurstReplicas` was not validated at admission time, allowing users to pass negative integers or invalid values. When the DaemonSet controller reconciled an object with a negative `BurstReplicas` value, it used this value directly to calculate buffer capacities when allocating error channels (e.g. `make(chan error, updateDiff)`), causing a runtime panic (`makechan: size out of range`) and bringing down the controller manager in a crash loop (DoS).

With this fix:
- `appsvalidation.ValidatePositiveIntOrPercent` is used to validate `BurstReplicas` upon creation or update.
- Invalid/negative `BurstReplicas` values are properly rejected at the admission phase.

Fixes #2512

Run unit tests for validating webhook:
```bash
go test -v ./pkg/webhook/daemonset/validating/...
```

None.


vishalmore90:
@Fei-Guo @zmberg 
Could someone please take a look at this PR?

PR #2538, opened 2026-08-22
fix(broadcastjob): prevent active counter underflow panic in deleteJobPods

### LINKED ISSUE #2537: [BUG] BroadcastJob controller panic: `deleteJobPods` double-decrements `active` counter causing int32 overflow
**What happened:**

The kruise-controller-manager enters CrashLoopBackOff with the following panic
during BroadcastJob reconciliation:

panic: makechan: size out of range [recovered]
        panic: makechan: size out of range

goroutine 941 [running]:
github.com/openkruise/kruise/pkg/controller/broadcastjob.(*ReconcileBroadcastJob).reconcilePods(...)
        /workspace/pkg/controller/broadcastjob/broadcastjob_controller.go:434
github.com/openkruise/kruise/pkg/controller/broadcastjob.(*ReconcileBroadcastJob).Reconcile(...)
        /workspace/pkg/controller/broadcastjob/broadcastjob_controller.go:326

**Root cause:**

There is an accounting bug in `deleteJobPods` that double-decrements the
`active` pod counter, which eventually causes an int32 overflow and panic in
`reconcilePods`.

The bug chain:

1. `filterPods()` (broadcastjob_utils.go:46-64) correctly excludes pods with
   `DeletionTimestamp != nil` (Terminating) from `activePods`. So `active`
   starts at 0 when all remaining pods are Terminating.

2. `getNodesToRunPod()` places pods that no longer fit their node into
   `podsToDelete`, which can include the same Terminating pods already
   excluded by `filterPods`.

3. `deleteJobPods()` (broadcastjob_controller.go:660-692) is called at
   line 318 with `podsToDelete`. On each successful delete, it decrements
   `active--` (line 679). But these pods were **never counted** in `active`
   by `filterPods`, so `active` underflows to negative values (e.g. 0 → -8).

4. The corrupted `active` is then passed to `reconcilePods()` at line 326:

   ```go
   // line 407: parallelism defaults to 1<<31-1 (2,147,483,647) when unset
   parallelism := int32(parallelismInt)

   // line 421-423: active is negative, so this branch is taken
   } else if active < parallelism {
       diff := integer.Int32Min(parallelism-active, rest)
       // parallelism - (-8) = 2,147,483,655 which overflows int32

   // line 434: diff is now a large negative number
   errCh = make(chan error, diff)  // PANIC: makechan: size out of range

Trigger condition:

Nodes become NotReady/Unknown while a BroadcastJob has pods running on them.
The pods get stuck in Terminating state because no kubelet is available to
execute the deletion. Each reconcile loop:
- filterPods excludes them (DeletionTimestamp is set) → active = 0
- deleteJobPods re-attempts deletion, succeeds (API-side), decrements
  active → active = -N
- reconcilePods overflows → panic
- controller-runtime recovers → requeue → same panic → CrashLoopBackOff



// line 677-681, broadcastjob_controller.go
} else {
    failedLock.Lock()
    failed++
    if active > 0 {
        active--
    }
    r.recorder.Eventf(job, corev1.EventTypeNormal, kubecontroller.SuccessfulDeletePodReason, "Delete pod: %v", pods[ix].Name)
    failedLock.Unlock()
}

Alternatively, reconcilePods could guard against negative active:

if active < 0 {
    active = 0
}

Environment:
- Kruise version: v1.7.2 (also confirmed unpatched in v1.9.1 master)
- Kubernetes version: (fill in)
- Cluster: large Slurm-backed cluster with ~150 nodes, frequent node churn

How to reproduce:
1. Create a BroadcastJob targeting all nodes
2. Make several nodes go NotReady (e.g. cordon + stop kubelet)
3. Pods on those nodes enter Terminating but cannot be finalized
4. Observe kruise-controller-manager panic on the next reconcile loop

photoszzt:
## Summary

`deleteJobPods` unconditionally decrements the `active` counter for every successfully deleted pod (line 679). However, some pods were never counted as active by `filterPods` — for example, pods classified as "failed" due to exceeding `restartLimit` (RestartPolicy=OnFailure with high RestartCount). When these pods also appear in `podsToDelete` (because their node no longer passes `checkNodeFitness`, e.g. a NoExecute taint was added), the `active` counter underflows to a negative value.

The negative `active` is then passed to `reconcilePods`, where `parallelism - active` overflows int32 and produces a negative `diff`. `make(chan error, diff)` panics with `makechan: size out of range`, and controller-runtime's recovery re-enqueues the same object, causing a permanent CrashLoopBackOff.

### Bug chain

1. **`filterPods()`** classifies a Running pod as "failed" (RestartCount > restartLimit) → not counted in `active`
2. **`getNodesToRunPod()`** puts the same pod in `podsToDelete` (node has NoExecute taint, pod has no toleration, DeletionTimestamp is nil)
3. **`deleteJobPods()`** successfully deletes it and does `active--` → `active` goes from 0 to -1 (or lower)
4. **`reconcilePods()`** computes `parallelism(2^31-1) - active(-N)` → int32 overflow → `make(chan error, <negative>)` → **panic**

### Changes

- **`deleteJobPods`** (line 679): guard `active--` with `if active > 0` to prevent underflow
- **`reconcilePods`** (line 413): defensive `if active < 0 { active = 0 }` clamp as a second safety net
- **Test**: `TestDeleteJobPodsDoesNotUnderflowActive` — 3 nodes, pods with high RestartCount (classified as "failed"), 2 nodes tainted NoExecute. Verifies no panic and `Status.Active >= 0`.

### Observed in production

Kruise v1.7.2 on a large cluster. Nodes went NotReady, pods stuck Terminating / classified as failed, kruise-controller-manager entered CrashLoopBackOff with restarts.

Fixes #2537


photoszzt:
I don't understand why this step fails:
```
Install Trivy
```
in the golangci-lint.

FillZpp:
/lgtm

@zmberg looks like there is an issue with the golangci-lint CI

PR #2556, opened 2026-09-07
fix(imagepulljob): join image name and digest with "@" instead of ":"

qnduncan:

An `ImagePullJob` whose `spec.image` carries a digest fails on every node, without a single request ever reaching the registry. With `spec.image: docker.io/library/nginx:1.31.5-trixie@sha256:05b8cb60c354a44ab824ea6e7dc69b46d50762cdbe728a347a5b656e6fb3d7c4`:

```
couldn't parse image name "docker.io/library/nginx:sha256:05b8cb60c354a44ab824ea6e7dc69b46d50762cdbe728a347a5b656e6fb3d7c4": invalid reference format
```

Note that the `1.31.5-trixie` tag is gone and the digest has taken its place, joined with a colon.

`NormalizeImageRefToNameTag` returns the **digest** as the `tag` for a digested reference — this is deliberate: `NormalizeImageRef` normalizes `name:tag@digest` down to `name@digest`, and `TestNormalizeImageRef` asserts it explicitly ("name with both tag and digest (digest should be preferred)").

Every place that joins the name and tag back together, however, uses `":"`. A tag matches `[\w][\w.-]{0,127}` and can never contain a colon, so `name` + `":"` + `sha256:...` is not a valid reference and `parsers.ParseImageName` rejects it in `pullImageV1` / `pullImageV1alpha2`. kruise-daemon then retries until `pullPolicy.backoffLimit` is exhausted and the job is marked failed:

```
I "Worker is starting to pull image" name="docker.io/library/nginx" tag="sha256:05b8cb60c354a44ab824ea6e7dc69b46d50762cdbe728a347a5b656e6fb3d7c4"
E "Pulling image backoff" err="couldn't parse image name \"docker.io/library/nginx:sha256:05b8cb60c354a44ab824ea6e7dc69b46d50762cdbe728a347a5b656e6fb3d7c4\": invalid reference format" backoff=1
...
E "Worker failed to pull image" err="couldn't parse image name ..." cost="15.0s"
```

Both the `name@digest` and `name:tag@digest` forms are affected, on both the CRI v1 and v1alpha2 pull paths.

This PR adds `daemonutil.JoinImageNameTag` as the counterpart to `NormalizeImageRefToNameTag` — it joins with `"@"` when the tag is a digest — and uses it wherever a name and tag are recombined:

- **`pkg/daemon/criruntime/imageruntime/cri.go`** — both CRI pull paths. This is the actual cause of the pull failure.
- **`pkg/daemon/criruntime/imageruntime/helpers.go`** — `ContainsImage` now searches `RepoDigests` as well, because a digest tag never appears in `RepoTags`. Without this, `PullIfNotPresent` could never short-circuit for a digested image, and `ImageTagStatus.ImageID` was left empty even after a successful pull.
- **`pkg/util/secret/parse.go`** — `AuthInfos` built an invalid reference, so `NormalizeImageRef` errored and it returned no auth info at all. It currently has no callers, but it is exported and has the same defect.
- **`pkg/daemon/imagepuller/*`, `pkg/controller/nodeimage/*`** — log messages, events and `ImagePullJob` status messages reported references such as `name:sha256:...`, which is what makes this failure mode confusing to diagnose (it reads like a malformed image name in the job spec). Map and set keys built the same way are changed consistently.

NONE

Automated:

- `TestJoinImageNameTag` covers tag, digest, and registry-with-port combinations.
- `TestNormalizeImageRefToNameTagRoundTrip` asserts that whatever `NormalizeImageRefToNameTag` returns can be joined back into a parseable reference. On master it fails for all three digest forms with exactly the error seen in production:

  ```
  joined reference "ubuntu:sha256:f2b6de56..." is not parseable
  ```

- `TestContainsImage` gains digest cases: present in `RepoDigests`, absent from it, and a tag match while `RepoDigests` is populated.

Manual:

```yaml
kind: ImagePullJob
metadata:
  name: prepull-by-digest
spec:
  image: docker.io/library/nginx:1.31.5-trixie@sha256:05b8cb60c354a44ab824ea6e7dc69b46d50762cdbe728a347a5b656e6fb3d7c4
  # the digest-only form fails identically:
  # image: docker.io/library/nginx@sha256:05b8cb60c354a44ab824ea6e7dc69b46d50762cdbe728a347a5b656e6fb3d7c4
  completionPolicy:
    type: Always
  parallelism: 1
```

Before this change the job ends with `failed: 1` and `couldn't parse image name ... invalid reference format` in the corresponding `NodeImage` status. After it, the image is pulled by digest and the job succeeds.

`go build ./pkg/...`, `go vet ./pkg/...` and the tests for `pkg/daemon/...`, `pkg/util/secret`, `pkg/controller/nodeimage` and `pkg/controller/imagepulljob` pass. `TestMatchRegistryAuths/test_credential_plugin_if_matched` fails in my environment on master as well (it needs a credential-provider plugin binary present) and is unrelated to this change.

- `IsDigest` deliberately tests for a colon rather than parsing the digest: the tag grammar forbids colons, so a colon unambiguously marks the value as a digest, and an unparseable digest still reaches `parsers.ParseImageName`, which reports a precise error instead of being silently rejoined with `":"`.
- Nothing in the webhook rejects a digest in `spec.image`, so the API already accepts these jobs today — they just always fail.
- The changes to log/event/status strings are not cosmetic-only: those strings are the primary signal an operator sees, and today they print an image reference that does not exist in the spec.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


qnduncan:
Hello @furykerry @zmberg,

Sorry for the pining.

This PR is to fix the issue that image with sha digest cannot be used in the ImagePullJob, which is an important feature needed by the communities.

Could you please help review the PR?

PR #2559, opened 2026-09-10
fix(e2e): check errors from PollUntilContextTimeout and GetSelectorPods

### LINKED ISSUE #2558: [BUG] e2e tests silently ignore errors from PollUntilContextTimeout and GetSelectorPods
## What happened

Several e2e test cases assign the `err` returned by `wait.PollUntilContextTimeout(...)` or `tester.GetSelectorPods(...)` but never check it with `gomega.Expect(err).NotTo(gomega.HaveOccurred())` before moving on. Found by running `ineffassign ./...`, which flags each of these as an ineffectual assignment.

This means that if the underlying wait times out, or the pod-listing API call fails, the test does not fail at that point with a clear error message. It either silently continues, or (in the best case) fails later on an unrelated assertion with a confusing message that doesn't point at the real cause.

Affected locations:
- test/e2e/apps/v1alpha1/daemonset.go:331  "Wait for daemonset ready" poll
- test/e2e/apps/v1alpha1/daemonset.go:498  "Check all surging Pods created" poll
- test/e2e/apps/v1alpha1/daemonset.go:520  "Check all old Pods deleted" poll
- test/e2e/apps/v1beta1/daemonset.go  same three, mirrored (lines 331/498/520)
- test/e2e/apps/v1alpha1/workloadspread.go:1538  `tester.GetSelectorPods(...)` result used without checking `err`

## What you expected to happen

Every `err` returned from a wait/poll or API-listing call in these test files should be checked immediately with a `gomega.Expect(err).NotTo(gomega.HaveOccurred(), ...)` (as is already done consistently elsewhere in these same files), so genuine failures surface with an accurate message instead of being masked.

## How to reproduce it

1. Run `ineffassign ./...` from the repo root on a fresh checkout (after `go install github.com/gordonklaus/ineffassign@latest`).
2. Observe the "ineffectual assignment to err" hits at the lines listed above.
3. Optionally, trace through the code: in each case, `err` is either never referenced again, or immediately overwritten by an unrelated `err :=` in a following statement.

## Anything else we need to know?

I have a fix ready (adds the missing `gomega.Expect(err)` checks at each location) and will open a PR referencing this issue. No behavioral change to the tests otherwise.

Faisal77666:
## 1. Describe what this PR does

Fixes an issue where several e2e test cases assign the `err` returned by `wait.PollUntilContextTimeout(...)` or `tester.GetSelectorPods(...)` but never check it before proceeding. This means a genuine timeout or API failure during a wait/list step would not fail the test with an accurate error it would either be silently ignored or surface later as a confusing, unrelated assertion failure. Adds the missing `gomega.Expect(err).NotTo(gomega.HaveOccurred())` checks at each location, matching the pattern already used elsewhere in the same files. Found via static analysis (`ineffassign ./...`), not from a live test failure.

## 2. Does this pull request fix one issue?

fixes #2558

## 3. Describe how to verify it

Run `ineffassign ./...` before this change and confirm the daemonset.go (v1alpha1 and v1beta1) and workloadspread.go (v1alpha1) lines listed in #2558 are flagged. Apply this PR's changes, re-run `ineffassign ./...`, and confirm those specific lines no longer appear. Also confirm `go build ./...` and `go vet ./...` remain clean.

## 4. Special notes for reviews

This is purely an e2e test hardening change  no production code (pkg/, apis/) is touched, and no test behavior changes when the underlying operations succeed (the happy path). The only change in behavior is that a genuine timeout or API failure will now correctly fail the test with a clear gomega message instead of being silently dropped.


Faisal77666:
@furykerry please take a look

PR #2572, opened 2026-09-20
fix(pub): do not mark digest-pinned pods inconsistent on bare imageID

### LINKED ISSUE #2570: [BUG] PUB counts digest-pinned pods as unavailable on containerd (imageID is the config digest), and the webhook then skips them entirely
**What happened:**

I put a PodUnavailableBudget (`targetRef` → Deployment, `maxUnavailable: 1`) on an ingress-nginx controller with 2 replicas. Both pods are Running/Ready for days, `kubectl get deploy` says 2/2, but the PUB status stays at:

```
{"currentAvailable":0,"desiredAvailable":1,"observedGeneration":1,"totalReplicas":2,"unavailableAllowed":0}
```

kruise-manager logs show the controller did reconcile it (`PodUnavailableBudget update status ... expectedCount 2, currentAvailable 0, unavailableAllowed 0`), so it's not a stale status — it really counts both pods as unavailable.

The part that bit me: because `PodUnavailableBudgetValidatePod` returns `allowed` early for pods where `IsPodStateConsistent` is false ("doesn't count towards healthy and we should not decrement", pub_control_utils.go around L71), the webhook never checks the quota for these pods. So the PUB shows `unavailableAllowed: 0` which *looks* super strict, while in reality every pod of that workload can be deleted/evicted freely. A `kubectl drain` can take both replicas at once. That's the opposite of what I expected from the status.

**Why (as far as I can tell from the source, v1.9.1):**

The upstream ingress-nginx manifest pins the image by digest:

```
image: registry.k8s.io/ingress-nginx/controller:v1.13.0@sha256:dc75a7baec7a3b827a5d7ab0acd10ab507904c7dad692365b3e3b596eca1afd2
```

On containerd the pod status reports:

```
status.containerStatuses[0].image:   sha256:d9b2df1fa40e21139c739be6f30bc530b64a9500e4ccc3d86814d28835f1f4b3
status.containerStatuses[0].imageID: sha256:d9b2df1fa40e21139c739be6f30bc530b64a9500e4ccc3d86814d28835f1f4b3
```

i.e. the imageID is the bare image config digest, no repo, and it's a different digest than the manifest digest in the spec.

`IsPodStateConsistent` (pkg/control/pubcontrol/pub_control.go L286-306) takes the "all images are digest format" branch and calls `util.IsPodContainerDigestEqual` → `IsContainerImageEqual(spec.image, imageID)` (pkg/util/tools.go L139-157). With both sides parsed as digests it does `repo1 == repo2 && digest1 == digest2`; repo is empty on the imageID side and the digests are two different things (manifest digest vs config digest), so it's always false → pod is "inconsistent" → not available → and also bypasses the webhook.

For comparison, pods in the same cluster whose image is a plain tag get an imageID like `registry.example.com/team/app@sha256:<manifest digest>`, which would compare fine — but those never enter this branch anyway because the spec image isn't digest form. So the branch only ever runs for digest-pinned images, which is exactly where containerd (at least 2.x here) gives back something it can't compare against.

I also looked at #2563 — it stops the early `return true` in that branch, but still does the same imageID comparison, so I don't think it changes this case.

**What you expected to happen:**

A Ready pod that isn't in the middle of any in-place update should count as available, regardless of how the image is written in the spec. And a pod that the PUB can't reason about probably shouldn't silently lose protection.

Two ideas, not sure which you'd prefer:

1. Small: in `IsPodContainerDigestEqual`, if the imageID has no repository part (bare `sha256:...`), treat the comparison as "unknown" and fall through to the annotation/in-place checks instead of returning false.
2. Bigger: only require image consistency for PUB when the pod actually has an in-place update in progress (`apps.kruise.io/inplace-update-state` present); otherwise Ready is enough, like a plain PDB.

Happy to send a PR for (1) if that direction is okay.

**How to reproduce it (as minimally and precisely as possible):**

1. containerd-based cluster, install kruise 1.9.1.
2. `kubectl create deploy demo --image=registry.k8s.io/ingress-nginx/controller:v1.13.0@sha256:dc75a7baec7a3b827a5d7ab0acd10ab507904c7dad692365b3e3b596eca1afd2 --replicas=2` (any digest-pinned image works, the ingress-nginx one is just what I had; add `--` with a `sleep` command if you don't want it to actually start nginx).
3. Create a PUB with `targetRef` pointing at it, `maxUnavailable: 1`.
4. Wait for pods to be Ready, then `kubectl get pub demo -o jsonpath='{.status}'` → `currentAvailable 0`, `unavailableAllowed 0`.
5. `kubectl delete pod <one of them>` → goes through, no `PubPreventPodDeletion` event. Delete the second one → also goes through.

**Anything else we need to know?:**

Pods have the `kruise.io/related-pub` annotation set, and there's no SidecarSet or in-place update involved. kubelet is 1.36, containerd 2.3.4. Digest-pinned images are pretty common in upstream install manifests (ingress-nginx, cert-manager, etc.), so I guess anyone putting a PUB on those hits the same thing.

**Environment**:
- Kruise version: v1.9.1
- Kubernetes version (use `kubectl version`): v1.36.2
- Install details (e.g. helm install args): helm chart, default values
- Others: containerd 2.3.4

pujitha24:

Fixes `PodUnavailableBudget` (PUB) incorrectly treating Ready pods with digest-pinned
images (`image: repo@sha256:<digest>`) as unavailable on containerd, which both
under-reports `currentAvailable`/`unavailableAllowed` in PUB status and silently
bypasses the PUB admission webhook for those pods (see report for the exact mechanism).

`IsPodStateConsistent` compares `spec.containers[i].image` against
`status.containerStatuses[i].imageID`. On containerd, `imageID` can come back as a bare
digest with no repository, e.g. `sha256:<hex>` — this is the image *config* digest, a
different value than the manifest digest in `spec.image`, and it has no repository
component to compare at all. The existing comparison always evaluated this as "not
equal" and permanently marked the pod as inconsistent, even though the pod is Ready and
not undergoing any in-place update.

This PR adds a check for that specific bare-digest shape and treats it as inconclusive
(skip the comparison for that container) instead of treating it as a hard mismatch.

NONE

New unit tests in `pkg/util/pods_test.go` (`TestIsPodContainerDigestEqual`):
- case 1: a digest-pinned spec image compared against a bare `sha256:<hex>` imageID
  (the containerd shape from the report) now returns `true` (previously `false`). This
  case fails on the pre-fix code and passes on the post-fix code — verified locally by
  stashing the `pkg/util/pods.go` change and re-running the test.
- case 2: a full-reference imageID (`repo@sha256:<hex>`) that genuinely mismatches the
  spec digest still correctly returns `false`, confirming the fix does not disable
  mismatch detection for the normal (non-bare-digest) case.

Commands run and their results:
- `go build ./...` succeeds, except for a pre-existing, unrelated failure in
  `hack/gen-openapi-spec` (`undefined: appspub.GetOpenAPIDefinitions` etc.) that is
  present on `master` without this change too (missing generated code in this
  checkout, not something this PR touches).
- `go test -race ./pkg/util/... ./pkg/control/pubcontrol/... ./pkg/control/sidecarcontrol/...`
  passes.
- `go vet` and `gofmt -l` on the changed files are clean.
- `golangci-lint run ./pkg/util/... ./pkg/control/pubcontrol/...` (v2.13.2 locally,
  close to the `v2.1` this repo's CI pins) reports only two pre-existing issues in
  files this PR does not touch (`pkg/util/client/no_deepcopy_lister.go`,
  `pkg/util/refmanager/ref_manager.go`); nothing new from this diff. Note: I could not
  build this repo's exact pinned `golangci-lint@v1.61.0` locally (its vendored
  `golang.org/x/tools` fails to compile against the Go toolchain available here), so
  this is a close-but-not-exact mirror of CI's lint step; disclosing that limitation
  here rather than skipping the check.
- Did not run `make test` end-to-end (it also runs the full envtest suite, which
  requires downloading `etcd`/`kube-apiserver` binaries not available in this
  environment); ran the equivalent targeted `go test -race` on the changed and calling
  packages instead, which is sufficient to prove this specific fix.
- Base branch `master` CI is green as of the tip this branch is built from.

- Scope is intentionally narrow: `IsPodContainerDigestEqual` (this fix) is only called
  from PUB's `IsPodStateConsistent`. The general-purpose `IsContainerImageEqual` used
  elsewhere (e.g. `sidecarset_control.go`) is untouched, so this does not change
  SidecarSet hot-upgrade behavior, which may have a similar underlying exposure to the
  same containerd quirk but is out of scope for this report.
- `github.com/opencontainers/go-digest` was already an indirect dependency (present in
  go.sum); this PR moves it to a direct dependency in `go.mod` since `pkg/util/pods.go`
  now imports it directly. No other `go.mod`/`go.sum` churn.
- Report: https://github.com/openkruise/kruise/issues/2570

Fixes #2570


wuwudu:
Reporter of #2570 — this matches what I found on containerd: `imageID` comes back as a bare `sha256:<config digest>`, there's no repo part and it's a different digest from the one in `spec.image`, so the comparison can never succeed. Treating it as "can't tell" rather than "inconsistent" is what I'd expect; the webhook then stops silently skipping these pods and the status counts them like everyone else.

Only thought I had: if some runtime ever reports a bare digest that *is* the manifest digest, this skips a check that would have passed anyway, so it's safe in that direction too. Happy to confirm on our containerd nodes once there's an image with this in it.

PR #2573, opened 2026-09-20
fix(pub): resolve targetRef through a replicaSet's non-Deployment owner

### LINKED ISSUE #2571: [BUG] PUB targetRef to a Rollout-style workload (pods owned by a non-Deployment ReplicaSet) never associates its pods: status frozen at 0, webhook lets every delete through
**What happened:**

Follow-up to #2570, different root cause. A PUB with `targetRef` pointing at an Argo Rollout (blue-green, 2 replicas) never gets a usable status and doesn't protect anything:

```
{"currentAvailable":0,"desiredAvailable":1,"observedGeneration":1,"totalReplicas":2,"unavailableAllowed":0}
```

The pods are Running/Ready, image is a plain tag (so not the digest thing from #2570), and `kubectl evict` / `kubectl delete pod` on every pod goes straight through — no `PubPreventPodDeletion` event at all. I could evict both replicas back to back.

I watched it second by second after creating the Rollout and the PUB together:

- the PUB got reconciled once right away: `PodUnavailableBudget controller pods expectedCount podCount=0 expectedCount=2` + event `NoPods`. At that moment the Rollout's scale `status.selector` was still just `app=drill` and the ReplicaSet/pods didn't exist yet.
- pods came up a few seconds later, but they never got the `kruise.io/related-pub` annotation, and the PUB was never reconciled again (4+ minutes, nothing in kruise-manager logs for it).

**Why (from the v1.9.1 source, as far as I can tell):**

1. `pub_pod_event_handler.go` `GetPubForPod` (used by the mutating webhook to stamp `related-pub` on new pods) takes the pod's controllerRef → ReplicaSet → `controllerfinder.getPodReplicaSet`. That only climbs to the parent when the RS owner is a Deployment; for an RS owned by anything else it returns the RS itself (kind `ReplicaSet`). So the comparison against `targetRef.kind: Rollout` never matches and no annotation is written.
2. The pod event handler (`updatePod` → `PubControl.GetPubForPod`) and the validating webhook both only look at that annotation, so un-annotated pods neither re-trigger the PUB reconcile nor get checked on delete/evict. Status stays frozen at whatever the first reconcile saw.
3. The controller side (`GetPodsForRef` → `getScaleController` for a non-builtin kind) does find the pods *if* they already exist when the PUB is reconciled — it falls back to the scale subresource selector. That's why an earlier test where the pods happened to be created before the first PUB reconcile looked fine (pods got patched via the "old pods related-pub annotation" path). It only works by ordering luck; with a controller creating the PUB right after the Rollout it always loses the race. And even in the lucky case, pods from the *next* ReplicaSet (next rollout) go through path 1 again.

Same Rollout with a `selector` PUB instead of `targetRef` works fine: status correct within ~20s, new-revision pods get the annotation immediately, evictions are denied as expected. So this is specific to `targetRef` + a workload whose pods are owned by a ReplicaSet that isn't owned by a Deployment.

**What you expected to happen:**

Either `targetRef` to such workloads works (climb RS → its controller and compare that, or fall back to the scale selector in the webhook path the same way the controller does), or the PUB reports that it can't resolve the target (a condition / event / a validation error), instead of a healthy-looking object with `unavailableAllowed: 0` that lets everything through. The docs list targetRef as supporting "custom workloads that implement the scale subresource", which is what I assumed a Rollout was.

**How to reproduce it (as minimally and precisely as possible):**

1. Cluster with Argo Rollouts and kruise 1.9.1 (no custom workload whitelist configured).
2. Apply a blue-green Rollout `demo` with 2 replicas and, in the same apply, a PUB with `targetRef: {apiVersion: argoproj.io/v1alpha1, kind: Rollout, name: demo}`, `maxUnavailable: 1`. (Or create the PUB first, then the Rollout — same result.)
3. Wait for pods Ready. `kubectl get pub demo -o jsonpath='{.status}'` → `currentAvailable: 0`. Pods have no `kruise.io/related-pub` annotation.
4. `kubectl delete pod <any>` → allowed, no event. Delete the other one → also allowed.
5. Replace `targetRef` with `selector: {matchLabels: {app: demo}}` → status becomes 2/1 and deletes get denied.

**Anything else we need to know?:**

No `kruise-configuration` ConfigMap in the cluster (so no PPS custom workload whitelist). I'd guess CloneSet-style workloads and Deployment are fine because they're handled explicitly in the controller finder; it's the RS-owned-by-something-else shape that falls through.

**Environment**:
- Kruise version: v1.9.1
- Kubernetes version (use `kubectl version`): v1.36.2
- Install details (e.g. helm install args): helm chart, default values
- Others: containerd 2.3.4, Argo Rollouts (blue-green strategy)

pujitha24:

Fixes `PodUnavailableBudget.spec.targetRef` never matching pods whose ReplicaSet is owned by
something other than a builtin `Deployment` (e.g. an Argo Rollout). `getPodReplicaSet` in
`pkg/util/controllerfinder/controller_finder.go` used to climb past the ReplicaSet to its
owner only when that owner's `Kind` was literally `"Deployment"`; for any other owner kind it
returned the ReplicaSet itself as the workload. Since a PUB's `targetRef.Kind` is the real
owner (e.g. `Rollout`), the returned `Kind` (`ReplicaSet`) never matched, so the pod never got
the `related-pub` annotation, the PUB status stayed frozen, and the validating webhook let
every delete/evict through.

The fix generalizes the climb: it now always resolves the ReplicaSet's owner through the full
finder chain (`GetScaleAndSelectorForRef`), which already falls back to querying the
Kubernetes scale subresource for any kind it doesn't otherwise recognize. If the owner can't
be resolved by any finder, it falls back to the ReplicaSet itself, preserving prior behavior
for that case.

While generalizing this path, `getScaleController` was also given a `meta.IsNoMatchError`
carve-out (mirroring the pattern already used in `pkg/controller/controllers.go` and
`pkg/util/fieldindex/register.go`): if the RESTMapper can't resolve the owner's kind at all
(e.g. its CRD isn't installed), that's now treated as "owner not found" instead of a hard
error that would abort the whole caller (this matters because the newly-generalized climb can
now reach `getScaleController` from more call sites, including `GetExpectedScaleForPods`,
which aggregates over many pods and previously would bail entirely on the first such error).

NONE

```
go test ./pkg/webhook/pod/mutating/... -run TestPubMutatingPod -v
go test ./pkg/util/controllerfinder/... -v
```

Both new cases in `TestPubMutatingPod` reproduce the bug: with the fix reverted, the
"replicaSet's non-deployment owner" case fails with a clean assertion mismatch (pod not
annotated), and `Test_getScaleController_unresolvableKind` fails with a propagated
`*meta.NoKindMatchError`. Both pass with the fix applied.

I could not reproduce this against a live cluster with Argo Rollouts installed (no such
environment available here); validation is the targeted unit tests above, which fail before
the fix and pass after it, plus `go build ./...`, `go vet`, and `golangci-lint run` on the
changed packages (all clean), plus a full `go test ./pkg/...` run showing no regressions in
this package or in the other consumers of the shared `controllerfinder` utility
(`workloadspread`, `persistentpodstate`, `podprobemarker`, `nodepodprobe`,
`cloneset/sync`) and no change in the pre-existing, unrelated failures (missing `etcd`
binary for envtest suites, missing credential-provider-plugin binary).

Note: `master`'s own CI (`ci.yaml`) is currently green as of the last few commits, so any red
check on this PR should be from this change.

- This touches a shared low-level utility (`ControllerFinder`) used by PUB, WorkloadSpread,
  PersistentPodState, PodProbeMarker, NodePodProbe, and CloneSet sync, not just PUB. The
  generalization is additive/permissive only: it can now resolve owners it previously
  couldn't, and falls back to the old behavior whenever it can't resolve them, so it should
  not change behavior for any caller that was already working.

Fixes #2571


wuwudu:
Reporter of #2571 here — thanks for picking this up so fast.

Read through the diff and this is the exact spot I'd traced it to: `getPodReplicaSet` stopping at the ReplicaSet whenever the owner isn't a builtin Deployment, so the webhook ends up comparing `kind=ReplicaSet` against `targetRef.kind=Rollout`, never writes the `related-pub` annotation, and from then on nothing wakes the PUB controller up for those pods. Climbing through `GetScaleAndSelectorForRef` looks right to me, it's the same path the controller already uses to resolve the targetRef itself, so the two sides finally agree on what the workload is.

The `IsNoMatchError -> nil, nil` fallback is a nice touch too; without it a PUB pointing at a kind whose CRD isn't installed would start erroring instead of just not matching.

One thing worth keeping an eye on: in my case the status also stayed frozen at `currentAvailable: 0` because the controller only listed pods once at PUB creation (before the Rollout had created any). As far as I can tell the annotation landing on new pods is what makes the pod event handler enqueue the PUB again, so this change should fix that half as well — just calling it out so it's covered by whoever tests it. If it helps, I still have the Rollout repro from the issue and can run it against a build of this branch.

